PR #10, opened 2016-12-16
Update registry to point it at the right namespace

jfchevrette:
This udpates the registry in the cico_build_deploy.sh script so that we push to the right namespace/image:tag to ensure we don't overwrite the almighty stuff


kbsingh:
ok to [test]

PR #17, opened 2016-12-19
Adip membership

pmuir:
Deck completed to date


src/app/app-routing.module.ts
```diff
@@ -13,6 +13,14 @@ export const routes: Routes = [
     pathMatch: 'full'
   },
   {
+    path: 'home',
```
joshuawilson: @pmuir why did you add a Home route? The first route takes you to Home. If you want a direct path then you should use the Login example. The style you used here is for lazy loading.

src/app/home/home-routing.module.ts
```diff
@@ -6,11 +6,6 @@ import { HomeComponent } from './home.component';
 const routes: Routes = [
   {
     path: '',
-    redirectTo: 'home',
```
joshuawilson: This is how you do it for the default page. Is Home not the default anymore?

src/app/profile/profile-routing.module.ts
```diff
@@ -15,8 +15,10 @@ const routes: Routes = [
     component: ProfileComponent,
     children: [
       { path: '',          component: OverviewComponent },
-      { path: 'spaces',    loadChildren: './collaboration/collaboration.module#CollaborationModule' },
+      { path: 'spaces',    loadChildren: './spaces/spaces.module#SpacesModule' },
       { path: 'resources', loadChildren: './resources/resources.module#ResourcesModule' },
+      { path: '', loadChildren: './resources/resources.module#ResourcesModule' }
```
joshuawilson: What are you trying to do here without the path? The first line with no path is the default going to the Overview page.

src/app/app-routing.module.ts
pmuir: Hmm, I seem to have not actually comitted all the changes to app-routing - as routes for public, signin and others are missing. I will have to check in the morning, as I've not got the laptop I was using on me right now.

To answer your question - home is no longer the "first route" as you are calling it, public is, as it's the landing page for the app right now - this allows us to present a logged in and not logged in experience - but as those changes are missing...

src/app/home/home-routing.module.ts
pmuir: No, it's not, public should be, but I must have missed some changes.

src/app/profile/profile-routing.module.ts
pmuir: Will fix.

PR #26, opened 2016-12-19
Dummy spaces

pmuir:

* **Other information**:


pmuir:
@joshuawilson this is my approach to start to remove the variables from being totally hardcoded - thoughts?

pmuir:
Added dummy resources as well. Next will do activities unless I get told otherwise (that will be tmw).

PR #34, opened 2016-12-21
Tidy css

pmuir:

* **Other information**:

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -151,7 +151,7 @@ module.exports = function (options) {
 
         {
           test: /\.scss$/,
-          loaders: ["css-to-string", "css", "sass"]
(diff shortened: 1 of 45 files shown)


config/webpack.common.js
```diff
@@ -151,7 +151,7 @@ module.exports = function (options) {
 
         {
           test: /\.scss$/,
-          loaders: ["css-to-string", "css", "sass"]
+          loaders: ["css-to-string", "css-loader", "sass-loader"]
```
joshuawilson: I didn't think you needed to call out the loader in the name.

pmuir: Recently they started advising it and will require it soon AIUI.

joshuawilson: interesting, ok

PR #40, opened 2017-01-05
Teams Membership dialog

pmuir:
Layers on previous commit to add team membership in model


src/app/models/space.ts
```diff
@@ -1,5 +1,12 @@
-export class Space {
+import { Team } from './team';
+import { ProcessTemplate } from './process-template';
+
+export interface Space {
     name: String;
     path: String;
     description: String;
+    process?: ProcessTemplate;
+    private?: boolean;
```
joshuawilson: You can't use 'private' as a name it is a reserved word in TS.

src/app/shared/context.service.ts
```diff
@@ -0,0 +1,77 @@
...
+  get current(): Context {
+    return this._current;
+  }
+
+  private computeContext() {
+    // Find the most specific context menu path and display it
+    let c;
+    for (let m of this.dummy.contexts) {
+      if (this.router.url.startsWith(m.path)) {
+        if (c == null || m.path.length > c.path.length) {
```
joshuawilson: It does not appear that you initialize or set 'c' and therefor it would always be null.

src/app/shared/context.service.ts
```diff
@@ -0,0 +1,77 @@
...
+export class ContextService {
+
+  private _current: Context;
+  constructor(
+    private dummy: DummyService,
+    private router: Router,
+    private broadcaster: Broadcaster
+  ) {
+    // Listen for any context refreshes requested by the app
+    this.broadcaster.on<string>('refreshContext').subscribe(message => {
```
joshuawilson: Is 'message' used?

src/app/models/space.ts
pmuir: Fixed

src/app/shared/context.service.ts
pmuir: Look on the line below!

src/app/shared/context.service.ts
pmuir: No, I just copied the way this was used in other files.

pmuir:
@joshuawilson fixed menu

src/app/shared/context.service.ts
joshuawilson: I was not considering the for loop.

PR #43, opened 2017-01-05
Properly integrate current user handling

pmuir:

* **Other information**:


src/app/profile/overview/overview.component.html
```diff
@@ -1,40 +1,40 @@
...
+      </div>
+      <div class="row">
+        <h2>{{context.current.entity.attributes.fullName}}</h2>
+      </div>
+      <div class="row">
+        <p>
+          {{context.current.entity.attributes.username}}
+        </p>
+        <p>
+          {{context.current.entity.attributes.bio}}
```
joshuawilson: I'm unsure but it appears that attributes is unresolved here.

src/app/dummy/dummy.service.ts
```diff
@@ -344,6 +350,17 @@ export class DummyService {
...
+      message => {
+        this.addUser(message);
+        this._currentUser = this.lookupUser(message.attributes.username);
+      }
+    );
+    this.broadcaster.on<string>('logout').subscribe(
+      message => {
+        this._currentUser = null;
+      }
+    )
```
joshuawilson: This statement is missing the closing semi-colon.

src/app/profile/overview/overview.component.html
pmuir: Yeah, I wasn't sure how to best do a cast here. I kinda need something like:

   {{ (context.current.entity as User).attributes.bio }}

Any ideas?

src/app/dummy/dummy.service.ts
pmuir: See #44

PR #46, opened 2017-01-06
Fix the way emails work

pmuir:
After discussion with Todd, make emails a bit simpler

### DIFF
--- a/src/app/models/email.ts
+++ b/src/app/models/email.ts
@@ -1,4 +1,0 @@
-export class Email {
-    address: string;
-    keepPrivate: boolean;
-}
--- a/src/app/models/profile.ts
+++ b/src/app/models/profile.ts
@@ -6,8 +6,11 @@ export class Profile {
     bio?: string;
     username?: string;
     url?: string;
-    emails?: Email[];
-    primaryEmail?: Email;
+    emails?: string[];
+    primaryEmail?: string;
+    notificationEmail?: string;
+    publicEmail?: string;
+    primaryEmailPrivate?: boolean;
     emailPreference?: string;
     notificationMethods?: string[];
 }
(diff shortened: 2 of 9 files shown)


src/app/models/profile.ts
```diff
@@ -6,8 +6,11 @@ export class Profile {
     bio?: string;
     username?: string;
     url?: string;
-    emails?: Email[];
-    primaryEmail?: Email;
+    emails?: string[];
```
joshuawilson: Does it really make sense to enumerate the emails like this, instead of having email and email type? This doesn't scale as nicely.

src/app/models/email.ts
```diff
@@ -1,4 +0,0 @@
-export class Email {
```
joshuawilson: I might be nice to keep this as a type instead of storing them as a string. This may help with validation. Though it is not necessary.

src/app/models/profile.ts
pmuir: You mean have the type attached to the email? like

```
interface Email {
  address: string;
  type: string;
}
```

joshuawilson: yes, then you can have as many new types as you want and you fetch them from the db. We can take another level and create email types so that we are sure they will match.

joshuawilson: it may also depend on what keycloak does for us and what we do on the backend to store extra user info

pmuir: I don't think that's the right data model tbh. What is your actual concern?

pmuir: I disagree, the client side data model should cleanly represent what the UX is. You then use the service to adapt that on to your server side endpoints.

pmuir: The way we could model this is with some sort of map and labels, but tbh I want to understand what you are acutally worried about ;-)

joshuawilson: I was thinking that if you want to change the type or add a type then you need to change the code. But if you store the type then you can add it to the server side and just fetch it. The data model should still reflect the UX, just not in a hard-coded fashion.

PR #48, opened 2017-01-09
Add a control screen

pmuir:
This is a meta-screen to allow us to control the app for now. It will need removing once we get full settings mgmt integrated


src/app/app-routing.module.ts
```diff
@@ -22,6 +22,10 @@ export const routes: Routes = [
     path: 'signup',
     loadChildren: './signup/signup.module#SignupModule'
   },
+    {
+    path: '_control',
```
joshuawilson: why did you start this path with an underscore?

joshuawilson:
Maybe I'm missing it but I don't see where you link to this new component. I guess you just need to know that it exists since it is outside of the scope of the design. Just bringing it up in case you actually wanted to add a link to it.

pmuir:
@joshuawilson I don't link to the component.

As you say, this page is outside the design, and exists only to allow us to "control" the webapp until the functionality is either included in the design or no longer needed.

For example, the clear local storage function will no longer be needed once we have the backend implemented for all functionality.

I was also going to put the controls for feature toggle enablement on this page, so that we have a way to control them that is outside of the webapp.

The page starts with an _ to indicate it's not a regular page.

And yes, you need to know the page is there :-)

PR #60, opened 2017-01-14
simpler app layout and easier component sizing

mitchharpur:
Pull request introduces a layout structure for components that offers simpler more html5 centric angular component content layout using flex. Content scrolling is more automatic and requires much less tweaking from styles to get content to 'hug' containers. Also included is a sample home component with example mofications.   

-fixed positioning for the menu
-poor scrolling behavior and layout for content in components that require a lot of tweaking 
-jittery header menu behavior

-flex based layout with patternfly components
-menu improvements
-more robust layout

* **Other information**:


joshuawilson:
[test]

src/app/header/header.component.html
```diff
@@ -27,13 +29,13 @@
                     </li>
                 </ul>
             </li>
-            <li *ngFor="let m of context.current.type.menus" [class.active]="m.active">
+            <li *ngFor="let m of context.current.type.menus"  [class.active]="m.active" [class.with-no-children]="(m.menus||[]).length===0" >
```
pmuir: Seems to be some extraneous whitespace in here.

src/app/home/home.component.html
```diff
@@ -1,93 +1,67 @@
...
-        <div class="col-md-2">
-            <a class="btn btn-default btn-block" type="button" href="https://fabric8.io/guide/getStarted/index.html">Learn</a>
-        </div>
-        <div class="col-md-2">
-            <button class="btn btn-default btn-block" type="button" (click)="createSpace.open()">
-                Start Collaborating
-            </button>
-        </div>
-        <div class="col-md-4"></div>
+<div class="container-fluid container-cards-pf">
```
pmuir: Did you mean these to no longer be centered?

src/app/home/home.component.html
```diff
@@ -1,93 +1,67 @@
...
-                    </div>
-                </div>
-            </div>
+  </div>
+  <div class="row row-cards-pf">
+    <div class="col-sm-4">
+      <div class="card-pf card-pf-accented">
+        <div class="card-pf-title">
+          <h4>Recent Collaboration Spaces</h4>
+          <a (click)="createSpace.open()">New</a> | <a [routerLink]="['/pmuir/spaces']">Browse</a>
```
pmuir: You deleted the box content

pmuir: Actually no, the text just appears as white on my screen.

src/app/home/home.component.html
mitchharpur: yes .. just took a stab at it to make for more space for content in desktop usage. If we set a max width we can go back to being centered.

src/app/home/home.component.html
mitchharpur: yep .. not a layout problem... content shows up when refresh browser ...pointing to another rendering issue ...

mitchharpur: I want to get to the bottom of why its not showing up ... just need the layout in first

src/app/home/home.component.html
mitchharpur: I you want max width ... i can add that in ... make no difference to me ... just thought it made a better experience in desktop usage mode

src/app/header/header.component.html
mitchharpur: the extra white space above scrollbar is there a a temp work around until all components have been tweaked to have a top padding of xpx. I just inserted a bottom padding in the header to make component content not lie right up agains header.

src/app/app.component.html
```diff
@@ -1,7 +1,11 @@
-<alm-app-header></alm-app-header>
-
-<main class="app">
-  <router-outlet></router-outlet>
+<header class="padding-bottom-10" >
```
joshuawilson: I think it makes more sense to have the header component contain the header tag. Same for 'main' and 'footer'. Though it may be worth looking at the Angular style guide to confirm.

mitchharpur: The problem with the header tag being in the alm-app-header is that it effectively becomes a child of the alm-app-header (effectively a div from a layout perspective). The goal is to always to reduce the amount markup required and separate 'layout' concerns (i.e semantic header) with 'content' concerns (contents of the alm-app-header element). This results in more style rules needed which slow things down from a rendering perspective.

joshuawilson: I don't think there is anything that says alm-app-header needs a div at the top level. It could be a header tag. The idea is that the header is self-contained.

joshuawilson:
The middle box on home http://localhost:3000/home is shorter then the outer ones.

src/app/app.component.html
mitchharpur: In semantic html5 a header is a tag with specific meaning from a layout perspective... a separate concern to navigation. The angular alm-app-header is not really a header by function ... it's really a navigation strategy ... The idea is to separate layout (header) from function (navigation) from content  as much as possible. In general this results in more maintainable css and cleaner html5. Navigation could be placed anywhere in the layout depending on different responsive layouts. Also from a styling perspective, separating layout from content always results in fewer css and less nested rules.

src/app/home/home.component.html
pmuir: agreed.

src/app/home/home.component.html
pmuir: Ok

src/app/app.component.ts
```diff
@@ -10,6 +10,9 @@ import { AppState } from './app.service';
  * Top Level Component
  */
 @Component({
+  host:{
+    class:'app app-component flex-container in-column-direction flex-grow-1'
```
joshuawilson: I'm not sure why you want to do it this way, but it you do, the word "class" needs to be in quotes.

src/app/header/header.component.html
```diff
@@ -27,13 +28,13 @@
                     </li>
                 </ul>
             </li>
-            <li *ngFor="let m of context.current.type.menus" [class.active]="m.active">
+            <li *ngFor="let m of context.current.type.menus" [class.active]="m.active" [class.with-no-children]="(m.menus||[]).length===0" >
```
joshuawilson: Please add spaces to improve readability.

src/app/header/header.component.scss
```diff
@@ -36,32 +29,41 @@
...
-}
+ }
 .navbar-pf {
+    .navbar-primary
+    {
+      position:relative;
+      height:auto !important;
+      //min-height:$navbar-height
+    }
     .navbar-primary.persistent-secondary
```
joshuawilson: I think this line should end in a comma.

src/app/header/header.component.scss
```diff
@@ -36,32 +29,41 @@
...
+    >li
+    {
+      a{
+        // hack:ensures elements do not cause secondsry menu to  'jiggle'
+        max-height:$navbar-height;
+      }
+      &.active {
+        &.with-no-children
+        {
+          margin-bottom:0px
```
joshuawilson: the "px" is redundant, I don't think it should be on a "0"

src/app/home/home.component.html
```diff
@@ -1,93 +1,67 @@
...
+          <p style="font-size:xx-small;" ><span> * for the currently logged in user.</span></p>
+</div>
+<div class="card-pf-body">
+  <div class="card-pf-items">
+    <ul class="list-unstyled">
+      <li><small>TODO: Add infinite scroll</small></li>
+      <li><small>TODO: What does the activity look like?</small></li>
+    </ul>
+  </div>
+</div>
```
joshuawilson: for readability, please clean up the indention

src/app/home/home.component.scss
```diff
@@ -1,9 +1,50 @@
...
-  float: none;
+div[class*='col-']
+{
+  position:relative;
+  border:solid $debug-pf-border $debug-pf-column-color;
+  padding-left:$content-offset;
+  padding-right:$content-offset;
+  .card-pf
+  {
+    padding:0px $content-offset;
```
joshuawilson: more use "px" on a "0", there are 2 more below

src/app/home/home.component.scss
```diff
@@ -1,9 +1,50 @@
...
+    min-height:200px;
+    .card-pf-title
+    {
+      background-color: #EEE;
+      margin:-$content-offset;
+      margin-top:-0px;
+      margin-bottom:0px;
+      padding:$content-offset;
+      padding-left:$content-offset;
```
joshuawilson: why call out "padding-left" and "padding-right" when they are covered by "padding"?

src/app/home/home.component.ts
```diff
@@ -3,6 +3,9 @@ import { Component, OnInit } from '@angular/core';
 import { Router }            from '@angular/router';
 
 @Component({
+  host:{
+      class:"app-component flex-container in-column-direction flex-grow-1"
```
joshuawilson: "class" needs to be in quotes

src/assets/stylesheets/_custom.scss
```diff
@@ -73,17 +73,17 @@ $textCenter: center;
...
+	-o-transform:      $transform;
 	transform:         $transform;
 }
 @mixin transition($transition) {
-	-webkit-transition: $transition;  
+	-webkit-transition: $transition;
 	-moz-transition:    $transition;
-	-ms-transition:     $transition; 
-	-o-transition:      $transition; 
+	-ms-transition:     $transition;
```
joshuawilson: I'm not sure, but I think this is not a valid CSS property.

src/assets/stylesheets/_layout.scss
```diff
@@ -0,0 +1,228 @@
...
+}
+
+.overflow-auto{
+  overflow: auto;
+}
+
+.fill-container
+{
+  position:absolute;
+  top:0px;
```
joshuawilson: more use of "px" with "0", there are at least 8

src/app/app.component.ts
mitchharpur: This the correct way to style the host element (http://stackoverflow.com/questions/32853924/angular-2-how-to-style-host-element-of-the-component) . host is just simple json ... not a stringified jason fragment ... so "class" or 'class' is not required . What is the reason you want "class" in quotes ?

joshuawilson: Class is a reserved word in TypeScript. However I can't find anything in the Angular docs providing an example. I only found http://stackoverflow.com/questions/35996212/is-it-possible-to-add-a-dynamic-class-to-host-in-angular-2

src/app/header/header.component.scss
mitchharpur: 0px shows intent making it more clear what the author had in mind. it could be em rem of % , but pixels give fine grained control

src/app/home/home.component.scss
mitchharpur: 0px shows intent and it is the general convention to be specific.

src/assets/stylesheets/_layout.scss
mitchharpur: use of unit of measure shows intent.

src/assets/stylesheets/_custom.scss
mitchharpur: I did not add these styles ... may have inadventently auto-formatted ..

src/assets/stylesheets/_layout.scss
mitchharpur: px demonstrates intent. https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties
https://developer.mozilla.org/en-US/docs/Web/CSS/length ... or am i missing your point ...

src/app/app.component.html
pmuir: I tend to agree with @mitchharpur this makes the header (i.e. the navbar) a reusable component.

pmuir: https://github.com/fabric8io/fabric8-ui/issues/72

pmuir: I've created an issue for this, so we can come back to it. I suggest we just accept the change for now.

src/app/home/home.component.scss
joshuawilson: The W3 spec on css only says that the units are optional. I'm from the school of thought that they should not be used since 0 equals the same value for all units. 
We are not using 'px' except to define the base for 'em', so I'm not sure they should be 'px' but should be 'em'. I caught it because I saw the use of 'px'. 
All that said, if this is the only thing left I won't hold up the merge for it.

mitchharpur: this makes sense to me... going forward , for consistency sake i will not use px for 0

src/assets/stylesheets/_custom.scss
joshuawilson: I think earlier versions of opera didn't support transition and that is why I thought it wouldn't work. Or it might be something else, but it appears to be valid now. That is why a caveated my original statement. :)

joshuawilson:
[test]

PR #61, opened 2017-01-16
Initial version of SpaceService.

### LINKED ISSUE #65: Create an Angular service for querying the Spaces REST API
**Description**

We need an Angular service to interact with the spaces REST API. For now, it would handle querying/listing of spaces.

**Implementation**

This service is imported in the module. It should be available as an injectable service.

**Reference**

User story - #53

VineetReynolds:

Feature: 

Currently capable of listing all spaces under /api/spaces from the Backend API.
Fixes #65 

No Angular service exists for querying spaces in the backend.

Makes available an injectable service to query spaces.

* **Other information**:

TODO: Additional unit tests for pagination. The `nextLink` property is lost across spaceService invocations. Needs additional work to understand this behavior only in unit-tests.
TODO: Cleaner separation of Space model interface, so prototyping and REST API concerns can be separated. The structure of the existing Space type does not map to the backend API cleanly.


src/app/profile/spaces/space.service.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  constructor(private http: Http, private logger: Logger, private auth: AuthenticationService) {
+    if (this.auth.getToken() != null) {
+      this.headers.set('Authorization', 'Bearer ' + this.auth.getToken());
+    }
+    logger.log('SpaceService using url:' + this.spacesUrl)
+  }
+
+  getSpaces(pageSize: number = 20): Promise<Space[]> {
+    console.log("Is logger there in getspaces?" + (this.logger == null));
```
pmuir: This looks like debugging.

src/app/profile/spaces/space.service.ts
```diff
@@ -0,0 +1,100 @@
...
+  private spacesUrl = process.env.API_URL + "spaces";
+  private nextLink: string = null;
+
+  private spaces: Space[] = [];
+  private spaceIdIndexMap = {};
+
+  constructor(private http: Http, private logger: Logger, private auth: AuthenticationService) {
+    if (this.auth.getToken() != null) {
+      this.headers.set('Authorization', 'Bearer ' + this.auth.getToken());
+    }
```
pmuir: Should this be a reusable function on the AuthenticationService.

src/app/profile/spaces/space.service.ts
```diff
@@ -0,0 +1,100 @@
...
+    console.log("Is logger there in getspaces?" + (this.logger == null));
+    let url = this.spacesUrl + '?page[limit]=' + pageSize;
+
+    return this.http
+      .get(url, { headers: this.headers })
+      .toPromise()
+      .then(response => {
+        let spaces: Space[];
+        let links = response.json().links;
+        console.log('links:' + links);
```
pmuir: More debugging

src/app/profile/spaces/space.service.ts
```diff
@@ -0,0 +1,100 @@
...
+        if (links.hasOwnProperty('next')) {
+          this.nextLink = links.next;
+          console.log('Setting next link to:' + this.nextLink);
```
pmuir: More debugging

src/app/profile/spaces/space.service.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    return this.http
+      .get(url, { headers: this.headers })
+      .toPromise()
+      .then(response => {
+        let spaces: Space[];
+        let links = response.json().links;
+        console.log('links:' + links);
+        if (links.hasOwnProperty('next')) {
+          this.nextLink = links.next;
```
pmuir: This maybe warrants a code comment?

src/app/profile/spaces/space.service.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  getSpaces(pageSize: number = 20): Promise<Space[]> {
+    console.log("Is logger there in getspaces?" + (this.logger == null));
+    let url = this.spacesUrl + '?page[limit]=' + pageSize;
+
+    return this.http
+      .get(url, { headers: this.headers })
+      .toPromise()
+      .then(response => {
+        let spaces: Space[];
```
pmuir: You probably don't need to declare this up here, as you don't use it until you assign lower down.

src/app/profile/spaces/space.service.ts
```diff
@@ -0,0 +1,100 @@
...
+        let links = response.json().links;
+        console.log('links:' + links);
+        if (links.hasOwnProperty('next')) {
+          this.nextLink = links.next;
+          console.log('Setting next link to:' + this.nextLink);
+        }
+        spaces = response.json().data as Space[];
+        spaces.forEach((item) => {
+          this.logger.log(item);
```
pmuir: Probably makes sense to add some context to the log here (what are you logging).

src/app/profile/spaces/space.service.ts
```diff
@@ -0,0 +1,100 @@
...
+        });
+        // Update the existing workItem big list with new data
```
pmuir: Looks like the comment is wrong (should refer to spaces not work items)

src/app/profile/spaces/space.service.ts
```diff
@@ -0,0 +1,100 @@
...
+        this.updateSpacesList(spaces);
+        return this.spaces;
+      })
+      .catch(this.handleError);
+  }
+
+  getMoreSpaces(): Promise<any> {
```
pmuir: Some commonality with the previous method, worth refactoring out?

src/app/profile/spaces/space.service.ts
```diff
@@ -0,0 +1,100 @@
...
+          return newSpaces;
+        })
+        .catch(this.handleError);
+    } else {
+      return Promise.reject('No more item found');
+    }
+  }
+
+  private updateSpacesList(spaces: Space[]): void {
+    spaces.forEach((space) => {
```
pmuir: A comment here to explain the whole method probably wouldn't go amiss.

src/app/profile/spaces/space.service.ts
```diff
@@ -0,0 +1,100 @@
...
+          .splice(this.spaces.length, this.spaces.length, space);
+      }
+    });
+    // Re-build the map once done updating the list
+    this.buildSpaceIndexMapMap();
+  }
+
+  private buildSpaceIndexMapMap(): void {
+    this.spaces.forEach((space, index) => {
+      this.spaceIdIndexMap[space.id] = index;
```
pmuir: This might be neater if you use rxjs.

VineetReynolds: I'm not quite sure how rxjs would fit in here. Could you please clarify?

pmuir:
I was wondering if it would be neater to do with this with a map operator
or something like that.

package.json
```diff
@@ -102,6 +102,7 @@
     "@angularclass/conventions-loader": "1.0.2",
     "@angularclass/hmr": "1.2.0",
     "@angularclass/hmr-loader": "3.0.2",
+    "@types/lodash": "^4.14.50",
```
joshuawilson: types should go in the devdependencies section

src/app/profile/spaces/space.service.spec.ts
```diff
@@ -0,0 +1,89 @@
+import { cloneDeep } from 'lodash';
```
joshuawilson: We are following the same style pattern that the Angular team recommends. Please list the @ angular imports first, then a space followed by the external imports, and another space and internal imports.

src/app/profile/spaces/space.service.ts
joshuawilson: I'm looking into the Auth service this week. I'll see if it should be added. I can always refactor it out.

joshuawilson:
[test]

src/app/profile/spaces/space.service.ts
VineetReynolds: I looked at this comment, but put this on the shelf until further discussion. IMHO, the cleanest practice suitable for us, is to use a separate http service that handles bearer token auth, and delegates to the Angular http client service for other actions, like so: http://stackoverflow.com/a/34465070/3916

package.json
VineetReynolds: I'm using lodash, for deep cloning, in the SpaceService class, not just in the unit test.

src/app/profile/spaces/space.service.spec.ts
VineetReynolds: Done. Fixed in both SpaceService and in the unit test.

src/app/profile/spaces/space.service.ts
pmuir: +1

package.json
joshuawilson: Types are for converting TypeScript to JavaScript and are only needed when we compile it, not run it.

joshuawilson:
[test]

PR #76, opened 2017-01-20
improved menu responsive behavior and neaten up

mitchharpur:
Neatens up and improved responsive behavior of the the menu bar. The menu bar is smaller in height to make more space for content as originally requested. The size of the menu text is driven by patternfly/bootstrap $font-size-large , automatically and correctly scaling around this font size.
Also the collapsed menu (mobile) correctly fills the view width and moves to the left. The html structure is also more in line with patternfly (hav-header now included), and the drop down menu icons are correctly aligned.

-The menu bar is too large for the type of application being built ... taking up to much real estate
-The responsive behavior is jittery and does not scale well when zoomed in
-The menu bar content does not align up well with a clearly defined box model making it very difficult to style and tweak without nasty cascading visual side effects 

-Fixes the above items with more neatly and well defined html layout

* **Other information**:

### DIFF
--- a/src/app/header/header.component.html
+++ b/src/app/header/header.component.html
@@ -1,76 +1,157 @@
-<nav class="navbar navbar-default navbar-pf" role="navigation" *ngIf="loggedIn">
-    <div class="navbar-header"></div>
-    <div class="collapse navbar-collapse navbar-collapse-1">
-        <ul class="nav navbar-nav navbar-primary persistent-secondary">
-            <!-- This part of the menu is dynamic based on context -->
-            <li class="context dropdown" dropdown>
-                <a dropdown-open>
-                    <span>
-                        <span class="{{context.current.type.icon}}"></span> {{context.current.name}}
-                    </span>
-                    <b class="caret"></b>
-                </a>
-                <ul class="dropdown-menu">
-                    <li *ngFor="let m of dummy.contexts">
-                        <a [routerLink]="[urlFeatureToggle + m.path]" *ngIf="m.path !== null && m != context.current"><span class="{{m.type.icon}}"></span> {{m.name}}</a>
-                        <span *ngIf="m.path === null"><span  class="{{m.type.icon}}"></span> {{m.name}}</span>
-                    </li>
-                    <li>
-                        <a [routerLink]="[urlFeatureToggle + '/pmuir/spaces']">Browse all</a>
-                    </li>
-                    <li class="divider"></li>
-                    <li>
-                        <a [routerLink]="[urlFeatureToggle + '/home']">Account home</a>
-                    </li>
-                    <li>
-                        <!-- TODO: This should pop up the new space dialog -->
-                        <a>New space</a>
-                    </li>
-                </ul>
-            </li>
-            <li *ngFor="let m of context.current.type.menus" [class.active]="m.active" [class.with-no-children]="(m.menus||[]).length===0" >
-                <a [routerLink]="[urlFeatureToggle + m.fullPath]">
-                    <span *ngIf="m.icon"  class="{{m.icon}}"></span>
-                    <span *ngIf="m.name">{{m.name}}</span>
-                </a>
-                <ul class="nav navbar-nav navbar-persistent" *ngIf="m.menus">
-                    <li *ngFor="let n of m.menus"  [class.active]="n.active">
-                        <a [routerLink]="[urlFeatureToggle + n.fullPath]">{{n.name}}</a>
-                    </li>
-                </ul>
-            </li>
+<nav class="navbar navbar-inverse navbar-pf" role="navigation" *ngIf="loggedIn">
+  <div class="navbar-header">
+    <ul class="nav navbar-nav navbar-right hidden-sm hidden-md hidden-lg ">
+      <li *ngIf="!loggedIn">
+        <a (click)='login();'>Sign In</a>
+      </li>
+      <li *ngIf="loggedIn" dropdown class="pull-right dropdown">
+        <a dropdown-open class="pull-right">
+          <div *ngIf="loggedInUser" id="header_dropdownToggle_header">
+            <span class="nav-item-icon">
+              <span *ngIf="!imgLoaded" class="nav-icon pficon-user"></span>
+              <img (load)="onImgLoad()" src="{{ loggedInUser.attributes?.imageURL }}" [ngClass]="{'show' : imgLoaded,'img-icon' : imgLoaded, 'hide' : !imgLoaded}">
+            </span>
+            <span class="nav-item-text">{{loggedInUser.attributes?.fullName}}</span>
+            <span class="nav-icon caret"></span>
+          </div>
+        </a>
+        <ul class="dropdown-menu view-width-100">
+          <li>
+            <a [routerLink]="[urlFeatureToggle + '/pmuir']">
+              <span class="nav-item-text">Profile</span>
+            </a>
+          </li>
+          <li>
+            <a href="https://fabric8.io/guide/getStarted/index.html">
+              <span class="nav-item-text">Help</span>
+            </a>
+          </li>
+          <li class="divider"></li>
+          <li>
+            <a [routerLink]="[urlFeatureToggle + '/pmuir/settings']">
+              <span class="nav-item-text">Settings</span>
+            </a>
+          </li>
+          <li>
+            <a *ngIf="loggedIn" (click)='logout();'>
+              <span class="nav-item-text">Sign out</span>
+            </a>
+          </li>
         </ul>
-    </div>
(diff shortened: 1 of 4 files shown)


joshuawilson:
[test]

src/app/header/header.component.html
```diff
@@ -1,74 +1,157 @@
...
-    <div class="navbar-header"></div>
+<nav class="navbar navbar-inverse navbar-pf" role="navigation" *ngIf="loggedIn">
+  <div class="navbar-header">
+    <ul class="nav navbar-nav navbar-right hidden-sm hidden-md hidden-lg ">
+      <li *ngIf="!loggedIn" >
+        <a (click)='login();'>Sign In</a>
+      </li>
+      <li *ngIf="loggedIn" dropdown class="pull-right dropdown">
+        <a dropdown-open class="pull-right" >
+          <div *ngIf="loggedInUser" id="header_dropdownToggle">
```
joshuawilson: You should not duplicate an 'ID".

src/app/header/header.component.scss
```diff
@@ -1,107 +1,132 @@
...
-    }
-    a {
-        color: white;
-        background-color: transparent !important;
-    }
-}
+$pf-cust-navbar-text-size:$font-size-large;
+
+$pf-custom-navbar-top-border-width:2px;
+$pf-custom-navbar-header-border-width:0px;
```
joshuawilson: Did we agree to no units on zero?
If we did, then there are at least 2 of them with units here.

src/app/header/header.component.html
```diff
@@ -1,74 +1,157 @@
...
-        <ul class="nav navbar-nav navbar-primary persistent-secondary">
+        <ul class="nav navbar-nav navbar-primary persistent-secondary navbar-left">
             <!-- This part of the menu is dynamic based on context -->
             <li class="context dropdown" dropdown>
                 <a dropdown-open>
-                    <span>
-                        <span class="{{context.current.type.icon}}"></span> {{context.current.name}}
-                    </span>
-                    <b class="caret"></b>
+                    <div *ngIf="loggedInUser" id="header_dropdownToggle">
```
pmuir: This chunk is incorrect. The left hand menu shows the context, not the logged in user. Needs reverting.

src/app/header/header.component.html
```diff
@@ -1,74 +1,157 @@
-<nav class="navbar navbar-default navbar-pf" role="navigation" *ngIf="loggedIn">
-    <div class="navbar-header"></div>
+<nav class="navbar navbar-inverse navbar-pf" role="navigation" *ngIf="loggedIn">
```
pmuir: Not sure what you are doing here, we now have this menu twice in the html - here, and at the bottom of the file.

src/app/header/header.component.scss
```diff
@@ -1,107 +1,132 @@
...
+            position:relative;
+            width:100%;
+            height:100%;
+            line-height: $pf-custom-navbar-item-height;
+            font-size:$pf-cust-navbar-text-size;
+            vertical-align:center;
+          }
+       }
+     }
+   }
```
pmuir: Styles look a lot better on the page.

src/app/header/header.component.html
mitchharpur: The responsive behavior of the overall menu makes the one hidden when mobile experience is  active and the other when desktop experience is active.  When you run the app .... collapse the width of the browser to mobile size and this menu kicks in ...

pmuir: Just wondering if there is a way to avoid duplicating the code inside the menu.

mitchharpur: yes ... once could write drop down menu as a sub component ... but that would prob be a premature optimization at this point ... so initially it's simpler and allows for standard usage of patternfly bootstrap menu without forcing a round peg into a square hole and writing duplicate script code. ... patternfly got a certain paradigm with the second level menus that makes it easier and simpler to do it this way

src/app/header/header.component.html
mitchharpur: The 'header' menu only displays when in mobile mode ... so its not on the left in desktop mode

src/app/header/header.component.html
joshuawilson: We do have a dropdown menu component already. It is available in the shared-component folder. Will that work for you?

joshuawilson:
[test]

src/app/header/header.component.html
```diff
@@ -1,74 +1,157 @@
...
             <li class="context dropdown" dropdown>
                 <a dropdown-open>
-                    <span>
-                        <span class="{{context.current.type.icon}}"></span> {{context.current.name}}
-                    </span>
-                    <b class="caret"></b>
+                    <div *ngIf="loggedInUser" id="header_dropdownToggle">
+                        <span class="nav-item-icon">
+                          <span *ngIf="!imgLoaded" class="pficon pficon-user"></span>
+                          <img (load)="onImgLoad()" src="{{ loggedInUser.attributes?.imageURL }}" [ngClass]="{'show' : imgLoaded,'img-icon' : imgLoaded, 'hide' : !imgLoaded}" />
```
joshuawilson: This shows up as a small square instead of the user image.

src/app/header/header.component.html
joshuawilson: How do you access the menu for selecting the context

src/app/header/header.component.html
```diff
@@ -1,74 +1,157 @@
...
+          <li>
+            <a *ngIf="loggedIn" (click)='logout();'>
+              <span class="nav-item-text">Sign out</span>
+            </a>
+          </li>
+        </ul>
+      </li>
+    </ul>
+  </div>
     <div class="collapse navbar-collapse navbar-collapse-1">
```
joshuawilson: The indention is off from here down. We are using 2 space indents for the UI.

src/app/header/header.component.html
```diff
@@ -1,74 +1,157 @@
...
-                    <a *ngIf="loggedInUser" id="header_dropdownToggle">
-                        <span *ngIf="!imgLoaded" class="pficon pficon-user"></span>
-                        <img (load)="onImgLoad()" src="{{ loggedInUser.attributes?.imageURL }}" [ngClass]="{'show' : imgLoaded,'img-icon' : imgLoaded, 'hide' : !imgLoaded}"
-                            class="pull-left marginR10" /> {{loggedInUser.attributes?.fullName}}
-                    </a>
-                    <b class="caret"></b>
+                    <div *ngIf="loggedInUser" id="header_dropdownToggle2">
+                        <span class="nav-item-icon">
+                          <span *ngIf="!imgLoaded" class="pficon pficon-user"></span>
+                          <img (load)="onImgLoad()" src="{{ loggedInUser.attributes?.imageURL }}" [ngClass]="{'show' : imgLoaded,'img-icon' : imgLoaded, 'hide' : !imgLoaded}" />
```
pmuir: It works fine for me.

src/app/header/header.component.html
pmuir: On a mobile?

src/app/header/header.component.html
pmuir: That's a good upgrade but shouldn't block this PR

src/app/header/header.component.html
mitchharpur: mobile context selection not yet working ... this is a problem i am aware of

mitchharpur: the selected context is now showing

src/app/header/header.component.html
```diff
@@ -1,76 +1,157 @@
...
+    <ul id="header_rightDropdown" class="nav navbar-nav navbar-right">
+      <li>
+        <a *ngIf="!loggedIn" (click)='login();'>Sign In</a>
+      </li>
+      <li *ngIf="loggedIn" dropdown class="pull-right dropdown">
+        <a dropdown-open>
+          <div *ngIf="loggedInUser" id="header_dropdownToggle2">
+            <span class="nav-item-icon">
+              <span *ngIf="!imgLoaded" class="nav-icon pficon-user"></span>
+              <img (load)="onImgLoad()" src="{{ loggedInUser.attributes?.imageURL }}" [ngClass]="{'show' : imgLoaded,'img-icon' : imgLoaded, 'hide' : !imgLoaded}">
```
joshuawilson: If you login with developers.redhat.com directly and don't have an image on file with them, there is no image to display and it leaves behind an artifact on the screen.

mitchharpur: i think it's a caching issue and unrelated to menu visual appearance which is the purpose of this pull request ... I suspect this is an issue with the login code

joshuawilson: ok, we can deal with this as a bug if it persists.

joshuawilson:
[test]

PR #77, opened 2017-01-23
Added placeholder images for codebases and workspaces

qodfathr:
Added some placeholder images for Create/Codebases and Create/Workspaces. Placeholders came from UXD.

Note that I wasn't able to use just a relative link to the .png files for some reason ... I'd love someone to show me how to do this correctly. I looked at what we did for placeholder images in the Almighty days, and that did not seem to work.


src/app/create/codebases/codebases.component.html
```diff
@@ -1,3 +1,5 @@
 <div id="codebases" class="container content">
-  <h1>TODO</h1>
+  <!-- can someone please show me how to make this be a relative url for the current component's directory?
+       I really just want to use 'codebases.png', but I've been unable to figure out how to do that -->
+  <img src='/app/create/codebases/codebases.png' style="width:100%;">
```
joshuawilson: @qodfathr as long as the image is in the same directory, you can just use the file name. 

img src='codebases.png' style="width:100%;"

src/app/create/codebases/codebases.module.ts
```diff
@@ -1,3 +1,4 @@
+import { module } from '@angular/upgrade/src/angular_js';
```
joshuawilson: Why did you add this line to the code?

joshuawilson:
[test]

joshuawilson:
hit the wrong button

joshuawilson:
[test]

src/app/create/codebases/codebases.component.html
qodfathr: If only that were true...I tried that over and over again and it did not work.

src/app/create/codebases/codebases.module.ts
qodfathr: I did not intend to; I did not type that. Perhaps some VS Code extension did it.

src/app/create/codebases/codebases.component.html
pmuir: It's not going to work like that, as this is an SPA app, the path to the file is actually /.

The right answer is to put codebases.png in /src/app/assets/img/ and then use a path of /assets/img/codebases.png in the html.

pmuir: Will fix it.

pmuir:
See #79

src/app/create/codebases/codebases.component.html
qodfathr: Thanks, @pmuir, for the explanation and the fix. I think this is worthy of some future consideration, because it makes components feel less 'componentized'. I believe I've seen some people write helper methods which augment the component with the ability to resolve its path. I think you then end up with html such as .

pmuir:
Yeah, that makes sense.

In general, you do want your assets split out from your components when
deployed, so you can apply different caching strategies etc.

Agreed we should add some sort of convention / helper, but that will come
later.

PR #79, opened 2017-01-26
Codebases

pmuir:
Closes #77 

Fixing review issues from Todd


joshuawilson:
@pmuir please squash the whitespace commit, we don't need that in the history. Otherwise it looks good.

pmuir:
I would always leave whitespace commits split out in the history, otherwise it becomes very hard to distinguish real change from cosmetic change in the history.

PR #85, opened 2017-01-27
Recent Contexts

pmuir:
* Implements recent contexts.
* Big tidy up of context service

Recent contexts are static

* **Other information**:


src/app/header/header.component.ts
```diff
@@ -50,8 +54,8 @@ export class HeaderComponent implements OnInit {
 
   getTogglePath(): void {
     this.toggleService
-        .getToggles()
-        .then(togglePaths => this.togglePaths = togglePaths);
+      .getToggles()
```
joshuawilson: It is acceptable to indent to the "dot" for this type of notation. It is common practice. It doesn't break anything but most find it easier to read.

src/app/header/header.component.ts
```diff
@@ -36,7 +38,9 @@ export class HeaderComponent implements OnInit {
     public context: ContextService
   ) {
     router.events.subscribe((val) => {
-      this.onNavigate();
+      if (val instanceof NavigationEnd) {
```
joshuawilson: You can "type" the variable like "val:NavigationEnd".  I haven't tested it will be enough of a check. But is worth know you can do it.

src/app/shared/context.service.ts
```diff
@@ -14,59 +18,183 @@ import { Broadcaster } from '../shared/broadcaster.service';
...
-        if (c == null || m.path.length > c.path.length) {
-          c = m;
-        }
+  lookupContextType(type: string): ContextType {
+    let ct = this.dummy.CONTEXT_TYPES.get(type);
+    // Make a copy of the context type before returning
+    return ct ? JSON.parse(JSON.stringify(ct)) : null;
+  }
+
+  get default(): Context {
```
joshuawilson: "default" is a reserved keyword

src/app/shared/context.service.ts
```diff
@@ -14,59 +18,183 @@ import { Broadcaster } from '../shared/broadcaster.service';
...
 
+  private computeContext(): Context {
+    // First, check if there is a recent context
+    let entityStr: string = this.extractEntity(this.router.url);
+    let spaceStr: string = this.extractSpace(this.router.url);
+    // TODO Implement team URLs
+    let teamStr: string = null;
+    // TODO This is horrible, make a switch?
+    for (let c of this.recent) {
+      if (entityStr && c.entity && !c.team && typeof c.entity === 'User') {
```
joshuawilson: I don't think this is the way "typeof" works. I think typeof is comparing with one of the standard types: "undefined", "object", "boolean", "number", "string", "function", "symbol".

src/app/profile/spaces/space.service.ts
```diff
@@ -108,6 +108,8 @@ export class SpaceService {
   private buildSpaceIndexMap(): void {
     this.spaces.forEach((space, index) => {
       this.spaceIdIndexMap[space.id] = index;
+      // TODO Hack in the path
+      space.path = space.attributes.name;
```
joshuawilson: I think this "hack" is break the unit tests.

pmuir:
It does seem to work, but I will switch to instanceof.

pmuir:
The indent previously was 4 spaces, not our standard two, which breaks the
auto format in tools.

pmuir:
Without this hack the UI is broken, as there is no path being generated -
one of @vineetreynolds PRS did this. I need to get Vineet to fix properly.

src/app/profile/spaces/space.service.ts
VineetReynolds: The attributes of the Space model class will eventually have to match the one of the Spaces resource in the backend. It looks like we'll have to move `path` to a computed property of a new UI model class, or compute path from `name`. Given that we need only the space name to query for one from the backend and display it on changing a route, I'm not sure we need to retain `path`.

src/app/header/header.component.ts
pmuir: The indentation was 4 spaces, it's now 2. This is a whitespace fix.

src/app/profile/spaces/space.service.ts
pmuir: I'll remove it.

@VineetReynolds what I would really like is a way to use the typescript get property syntax in model's but I can't find out how to do that!

src/app/shared/context.service.ts
pmuir: Fixed

src/app/shared/context.service.ts
pmuir: Switched to instanceof

src/app/header/header.component.ts
pmuir: That's a type assertion I think, wheras I need a type guard (see https://www.typescriptlang.org/docs/handbook/advanced-types.html)

pmuir:
Changes addressed. Let's see if this fixes the tests

pmuir:
[test]

PR #95, opened 2017-02-01
Remove beta/alpha routes and add support for parameterized user id

dgutride:
The links are no longer hardcoded to a specific username, also the alpha and beta routes that are unused were removed from the routing file.

Routed had hardcoded user name in the user section

Router links can now be created using the user id - the actual id needs to be added later when the auth code is ready.

* **Other information**:
@joshuawilson

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -36,82 +36,34 @@ export const routes: Routes = [
     path: 'home',
     loadChildren: './home/home.module#HomeModule'
   },
-  {
-    path: 'beta/home',
-    loadChildren: './home/home.module#HomeModule'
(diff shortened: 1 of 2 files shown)


pmuir:
[test]

pmuir:
Actually I need to check if this works or not

joshuawilson:
@dgutride this looks like the approach I was thinking of, but it doesn't appear to be ready to merge yet. 

1) it only effects a few routes
2) we need to switch from the ID to the username

dgutride:
Abandoning - going to restructure the change to remove the alpha/beta routes for now and leave the id parameterization until it's ready from the backend

PR #103, opened 2017-02-07
dockerfile cleanup

sbose78:
Cleaned up docker related files not in use.


joshuawilson:
@sbose78 this doesn't look like cleanup but removal. Why did you remove these files?

joshuawilson:
@sbose78 If you can verify with KB that we can delete this, please open this back up. Until them I'm closing it, as I understand we still want those files.

sbose78:
Sure @joshuawilson

PR #116, opened 2017-02-10
Fixes #112 by obtaining username from keycloak

### LINKED ISSUE #112: Menu bar is not loading if you don't enter a username in to the signup screen
I made an async error when I rewired the context service, and I managed to allow the menu bar to disappear if someone somehow doesn't have a username set up

pmuir:

* **Other information**:

### DIFF
--- a/src/app/shared/user.service.ts
+++ b/src/app/shared/user.service.ts
@@ -1,3 +1,4 @@
+import { ProfileService } from './../profile/profile.service';
 import { Injectable } from '@angular/core';
 import { Headers, Http } from '@angular/http';
 import { Logger } from './logger.service';
@@ -22,7 +23,8 @@ export class UserService {
   constructor(private http: Http,
               private logger: Logger,
               private auth: AuthenticationService,
-              private broadcaster: Broadcaster) {
+              private broadcaster: Broadcaster,
+              private profile: ProfileService) {
 
     this.broadcaster.on<string>('logout')
       .subscribe(message => {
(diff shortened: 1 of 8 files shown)


src/app/shared/user.service.ts
```diff
@@ -52,18 +54,20 @@ export class UserService {
         .get(this.userUrl, {headers: this.headers})
         .toPromise()
         .then(response => {
-          let userData = response.json().data as User;
+          let userData = response.json().data;
           // The reference of this.userData is
           // being used in Header
           // So updating the value like that
           this.userData.attributes = {
             fullName: userData.attributes.fullName,
```
dgutride: Object.assign(this.userData.attributes, userData.attributes) might be a better long term plan for copying over attributes

src/app/shared/user.service.ts
```diff
@@ -52,18 +54,20 @@ export class UserService {
...
-            imageURL: userData.attributes.imageURL
+            imageURL: userData.attributes.imageURL,
+            username: userData.attributes.username,
+            bio: userData.attributes.bio,
+            primaryEmail: userData.attributes.email
           };
           this.userData.id = userData.id;
+          this.profile.initDefaults(this.userData);
```
joshuawilson: I'm in the process of abstracting out the user.service. I don't think I should take the Profile with it. Can you have the profile listen to the broadcast service? Or find another way to set the Profile defaults?

pmuir:
Please file an issue as this is now merged.

PR #126, opened 2017-02-13
Fixes #120

### LINKED ISSUE #120: Settings link is only valid if the user is the currently logged in user

pmuir:

* **Other information**:


src/app/header/header.component.ts
```diff
@@ -9,17 +10,29 @@ import { Context } from './../models/context';
...
   providers: []
 })
-
 export class HeaderComponent implements OnInit {
   title = 'Almighty';
   imgLoaded: Boolean = false;
 
+  menuCallbacks = new Map<String, MenuHiddenCallback>([
+    [
+      'settings', function (context, menuItem) {
```
joshuawilson: menuItem is not used.

src/app/header/header.component.ts
```diff
@@ -9,17 +10,29 @@ import { Context } from './../models/context';
...
   templateUrl: './header.component.html',
   styleUrls: ['./header.component.scss'],
   providers: []
 })
-
 export class HeaderComponent implements OnInit {
   title = 'Almighty';
   imgLoaded: Boolean = false;
 
+  menuCallbacks = new Map<String, MenuHiddenCallback>([
```
joshuawilson: For flow sake, I think functions should be below the constructor. It's not a big deal.

src/app/models/menu-item.ts
```diff
@@ -1,8 +1,12 @@
-export class MenuItem {
+import { Context } from './context';
```
joshuawilson: Why are these imported? You are not using them.

joshuawilson:
This doesn't appear to fix the problem in #120. If I'm logged out, I can't see http://localhost:3000/pmuir/settings. There are errors.

src/app/header/header.component.ts
pmuir: Will remove.

src/app/header/header.component.ts
pmuir: Declaring variables after the constructor is a linting error.

src/app/models/menu-item.ts
pmuir: Good spot.

pmuir:
Ok, commit updated with changes requested.

Good catch on the logged out point - I hadn't thought through that scenario - I added a couple of guards to neatly handle that.

src/app/header/header.component.ts
joshuawilson: yes, you are right. I must have looked at it too fast and thought it was a function instead.

PR #148, opened 2017-02-15
Fixes #138 redirect after login

### LINKED ISSUE #138: Support remembering originating URL when redirecting to login
**Note: for support questions, please use one of these channels:** [Chat: AngularClass.slack](http://angularclass.com/member-join/) or [Twitter: @AngularClass](https://twitter.com/AngularClass)

* **I'm submitting a ...**
[ ] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Do you want to request a *feature* or report a *bug*?**

* **What is the current behavior?**

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via
https://plnkr.co or similar (you can use this template as a starting point: http://plnkr.co/edit/tpl:AvJOMERrnz94ekVua0u5).

* **What is the expected behavior?**

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**

- Angular version: 2.0.0-beta.X
- Browser: [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

pmuir:


src/app/run/pipelines/pipelines-routing.module.ts
```diff
@@ -1,17 +1,19 @@
-import { NgModule }  from '@angular/core';
+import { AuthGuard } from './../../shared/auth-guard.service';
```
joshuawilson: For consistency we have the imports done in the groups. First is Angular, then external modules, then internal. 

Sometimes I try to alphabetize them within their group too.

src/app/run/run-routing.module.ts
```diff
@@ -1,3 +1,4 @@
+import { AuthGuard } from './../shared/auth-guard.service';
```
joshuawilson: you are not using AuthGuard

src/app/shared/login.service.ts
```diff
@@ -22,4 +30,18 @@ export class LoginService {
     window.location.href = this.githubUrl;
   }
 
+  public redirectAfterLogin() {
+    let redirectUrl = this.localStorage.get<string>(this.REDIRECT_URL_KEY);
```
joshuawilson: we should work out how to do this without local storage, maybe Dana can help with that.

pmuir:
Why?

src/app/shared/login.service.ts
pmuir: Out of interest, why?

PR #156, opened 2017-02-16
Widgets integration

joshuawilson:
extract the shared components and import them via npm


joshuawilson:
[test]

joshuawilson:
used revert instead to bring back these commits

PR #166, opened 2017-02-20
162

pmuir:


src/app/shared/context.service.ts
```diff
@@ -197,6 +197,10 @@ export class ContextService {
   }
 
   private checkForReservedWords(arg: string): boolean {
+    // All words starting with _ are reserved
+    if (arg.startsWith('_')) {
```
dgutride: Is the underscore only being added to try to identify reserved words in the url?  Why not just use the named route params instead so :entity or :space values can be extracted.

pmuir:
I can't really parse your first comment. As for the second comment, we need
a way to disambiguate space names from views associated with the entity. We
have decided to use a leading underscore. I'm not aware of a way to
disambiguate without either reserving words or using an identifying prefix.
If you can explain another way, let me know. Nothing to do with angular
btw, this is just a logical problem.

On 20 Feb 2017 11:54 pm, "Dana Gutride" <notifications@github.com> wrote:

*@dgutride* commented on this pull request.

I'd recommend reconsidering this approach instead of just using the
built-in angular way of getting the set values on the URL.
------------------------------

In src/app/shared/context.service.ts
<https://github.com/fabric8io/fabric8-ui/pull/166#discussion_r102072223>:

> @@ -197,6 +197,10 @@ export class ContextService {
   }

   private checkForReservedWords(arg: string): boolean {
+    // All words starting with _ are reserved
+    if (arg.startsWith('_')) {

Is the underscore only being added to try to identify reserved words in the
url? Why not just use the named route params instead so :entity or :space
values can be extracted.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/fabric8io/fabric8-ui/pull/166#pullrequestreview-22817677>,
or mute the thread

src/app/shared/context.service.ts
pmuir: 1) Yes - though it doesn't only try, it succeeds.

pmuir: 2. :space is ambiguous with other pages under the entity

pmuir:
I think I understand your first comment now - it's not related to this PR AIUI, but rather the extraction of the :space and :entity from the URL? If so, issue + patch please - putting that comment on a PR will mean it gets lost in the long term.

dgutride:
@pmuir  - are you trying to solve the problem that /:entity/:space and /:entity/settings appear to be the same thing to the router?

If so - I've created a plnkr to demonstrate what I'm talking about:
https://embed.plnkr.co/EmL9snk036MvxZsNS4bc/

The router can distinguish between routes that it matches vs routes with parameters in them (see app.routes.ts in that example)

joshuawilson:
This work was done using the Angular router in #173. If there is still an issue or concern please raise it. I believe that the current solution is a better option, therefor I'm closing this PR.

pmuir:
@dgutride fixed another (similar, but not related) problem in #173 to #165 so we still need this PR. We still need to fix the URL scheme to the correct one, which is what #165 is about. I tried to explain in more detail in #173 but in essence right now we can't have spaces called settings, resources etc (so those words need reserving). We also will restrict the = routes underneath :entity apart from spaces once people start creating spaces (as there may be spaces created with the names we want). So we will prefix all routes under :entity with an _ and reserve all names starting with _.

pmuir:
I will rebase this on top of @dgutride changes.

pmuir:
@joshuawilson @dgutride rebase is done, please re-review :-D

dgutride:
@joshuawilson  - I'm good with this one.

PR #170, opened 2017-02-20
110

pmuir:


src/app/shared-component/wizard/wizard-navigator.ts
```diff
@@ -0,0 +1,21 @@
+import { WizardStepConfig } from './wizard-step-config';
+import { Wizard } from './wizard';
+
+// The navigator is used to inject wizard navigation into components and allow them to participate
+// in wizard navigation
+export class WizardNavigator {
```
joshuawilson: This looks generic enough, could this go into ngx-widgets instead?

src/app/shared-component/wizard/wizard-steps.ts
```diff
@@ -0,0 +1,9 @@
+import { WizardStep } from './wizard-step';
+
+export interface WizardSteps {
```
joshuawilson: This part looks to specific to be in ngx-widgets. If we want to move the other parts of "wizard" this will either need to stay or be changed. 
Or we keep all of it here but then is it really a shared component?

src/app/shared/entity.service.ts
```diff
@@ -0,0 +1,9 @@
+import { User } from 'ng-login';
```
joshuawilson: what is "ng-login"? are you importing a new module or is this a typo and should be ngx-login-client?

src/app/shared/entity.service.ts
```diff
@@ -0,0 +1,9 @@
...
+import { Entity } from 'ngx-login-client';
+export class EntityService {
```
joshuawilson: should the Entity service be in ngx-login-client

src/app/shared-component/wizard/wizard-navigator.ts
pmuir: @joshuawilson probably a good idea to move the wizard over, agreed. Suggest we merge this PR and then do that? (file an issue for the move).

src/app/shared/entity.service.ts
pmuir: Typo I think

src/app/shared/entity.service.ts
pmuir: We probably need to revisit this part of the model, it's exceedingly painful to use, and I'm not sure this is the right approach (probably better to have // model hierachies). I will try to work on this during the week.

PR #185, opened 2017-02-23
#183 Add build number to footer

pmuir:


joshuawilson:
There a lot of white space changes from 2 to 4 spaces. You also deleted several sections from the config, what was the reason to remove them?

pmuir:
Will fix - my laptop at home's vs code is oddly set up

Issue #190, opened 2017-02-23
Properly load styles from CSS files in production


pmuir:

pmuir:
Can't quite get this to work right now. Will try again once we have upgraded to newer webpack and angular, as I suspect some of the bugs will resolve themselves

pmuir:
Hmm, I seem to have made it work after all... Not sure if this *really* works or not...

pmuir:
This is now as complete as I can get it, I think (there is still some odd interaction with the sass loader, but we only have very small bits it style tags now

PR #215, opened 2017-02-28
fix(build): Fixes #214 Load CSS not sass

### LINKED ISSUE #214: Use patternfly compiled css rather than building from sass

pmuir:
Use the already compiled sass from patternfly


pmuir:
Part of #190

PR #217, opened 2017-02-28
fix(dependencies): add missing dependencies required for planner

dgutride:


config/webpack.dev.js
```diff
@@ -126,6 +126,7 @@ module.exports = function (options) {
           'ENV': JSON.stringify(METADATA.ENV),
           'NODE_ENV': JSON.stringify(METADATA.ENV),
           'HMR': METADATA.HMR,
+          'API_URL': JSON.stringify(METADATA.FABRIC8_WIT_API_URL),
```
joshuawilson: why is this needed?

dgutride: The planner is still using that value internally @joshuawilson so we had to have it available

joshuawilson: ok, for now but we need to fix it in Planner next.

PR #226, opened 2017-03-02
chore(build): allow running linked modules

pmuir:


config/webpack.common.js
```diff
@@ -125,7 +125,7 @@ module.exports = function (options) {
        *
        * See: https://webpack.js.org/configuration/resolve/#resolve-modules
        */
-      // modules: [ path.resolve(__dirname, 'src'), path.resolve(__dirname, 'node_modules')]
+      modules: [ helpers.root('src'), helpers.root('node_modules')]
```
joshuawilson: I get the error that modules is not allowed. That is why I commented it out.

pmuir: You mean that the configuration itself option is not allowed? Double check this one, as this solved the link problem for me.

joshuawilson: The problem is that the tests fail. The app will start but we can't test it.

pmuir: So let's run with this in dev but not in test :-) I'll set that up, as this really does work for me.

pmuir: BTW I don't understand the comment the tests fail - CI is passing, the unit tests pass for me locally...

pmuir:
[test]

pmuir:
@joshuawilson tests passing on CI - please shout if you still have problems.

joshuawilson:
it still fails for me locally but I'll merge it and then take a look again

PR #227, opened 2017-03-02
fix(routing): Fixed planner routing integration

sanbornsen:
These changes are needed in planner in order to get this working 
https://github.com/fabric8io/fabric8-planner/pull/1049


joshuawilson:
We are going to do this another way. Need to refactor the library not the app.

PR #228, opened 2017-03-02
Update the UI to use SpacesService to search for spaces

VineetReynolds:
Also updated ngx-fabric8-wit to 6.5.0.
Updated use of description attribute of spaces.


VineetReynolds:
@dgutride I removed the spaceService use from the header, because

* I believe it could be done a in different way, and in a different PR.
* The currentSpace method is not available in SpaceService, and is unlikely to be. So we'll have to find a way around this.

PR #232, opened 2017-03-03
fix(spaces): Upgrade to use Observables instead of Promises.

VineetReynolds:
Updated fabric8-ui to use the version of SpaceService that uses Observables.


VineetReynolds:
Please don't merge until this PR also includes a new release of ngx-fabric8-wit.

pmuir:
@VineetReynolds nice work!

PR #249, opened 2017-03-08
fix(build): exclude problematic sourcemaps

dgutride:
excludes sourcemaps for several libraries that
are not configured correctly


config/webpack.dev.js
```diff
@@ -119,6 +119,11 @@ module.exports = function (options) {
           exclude: [
             // Example helpers.nodeModulePath("fabric8-planner"),
             // Exclude any problematic sourcemaps
+            helpers.nodeModulePath("mydatepicker"),
+            helpers.nodeModulePath("ng2-completer"),
```
joshuawilson: looks like you can add ng2-dnd to this list

PR #265, opened 2017-03-12
fix(links): Fixed links in Header

VineetReynolds:
Fixed the username references in context menu and in dropdown.

**NOTE**: Like in the other PR, I haven't used `ngIf`. Maybe a router guard makes sense to verify that the user object for the currently logged in user is available from UserService.

### DIFF
--- a/src/app/header/header.component.html
+++ b/src/app/header/header.component.html
@@ -69,7 +69,7 @@
             </a>
           </li>
           <li>
-            <a [routerLink]="['pmuir/_spaces']">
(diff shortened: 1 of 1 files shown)


src/app/header/header.component.html
```diff
@@ -69,7 +69,7 @@
             </a>
           </li>
           <li>
-            <a [routerLink]="['pmuir/_spaces']">
+            <a [routerLink]="[loggedInUser.attributes?.username,'_spaces']">
```
pmuir: You shouldn't need the ? here...

PR #299, opened 2017-03-16
fix(contexts): properly reroute on guarded URLs

pmuir:
* Use a resolver rather than a guard to ensure the current context is set
* Use first() to ensure that a complete is always sent (odd behavior by Angular)
* Remove the auth part of this guard and apply that separately


pmuir:
[test]

PR #325, opened 2017-03-20
fix: update package.json fabric8-runtime-console to 1.0.15

rawlingsj:


rawlingsj:
Closing as the runtime pipeline has used an old git tag to use as the new version.

PR #335, opened 2017-03-22
fix(ngx-base): Use types from ngx-base.

VineetReynolds:
Using Broadcaster, Logger and Notification types from ngx-base.
Also updated ngx-login-client and ngx-fabric8-wit to versions that
depend on ngx-base.


pmuir:
We are holding on merge until @VineetReynolds get's this merged to runtime-console and planner.

PR #371, opened 2017-03-27
fix(login): Add local landing page

### LINKED ISSUE #311: Local development against local WIT server doesn't work
@alexeykazakov I will try to reproduce but any tips are most welcome.

pmuir:
Add a local landing page so that you can login from /

Fixes #311

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -11,7 +11,7 @@ export const routes: Routes = [
   // Only relevant locally, as the landing page sits on / in production
   {
     path: '',
-    redirectTo: '/rhn-support-pmuir',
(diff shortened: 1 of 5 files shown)


src/app/app-routing.module.ts
```diff
@@ -11,7 +11,7 @@ export const routes: Routes = [
   // Only relevant locally, as the landing page sits on / in production
   {
     path: '',
-    redirectTo: '/rhn-support-pmuir',
+    loadChildren: './landing-page/landing-page.module#LandingPageModule',
```
joshuawilson: this probably should not be lazy loaded

pmuir: No, it should as it's not loaded in production, only devel.

PR #385, opened 2017-03-28
fix(version): update property versions

fabric8cd:


joshuawilson:
[test]

joshuawilson:
@rawlingsj The npm version tracker is off, it is submitting an older version. 

closing this PR

rawlingsj:
@joshawilson I think I'm going to change the pipeline to make individual pull requests per version update. This will mean I can automatically close any unmerged PRs and replace it with a new PR if we have a newer version available. I'll do this tomorrow as I think that will prevent this situation of a PR with an old version included happening.

PR #445, opened 2017-04-02
fix(nav): remove PatternFly overrides

mindreeper2420:
remove PatternFly overrides from the navigation to better match the visual designs and PatternFly component guidelines.


pmuir:
Looks good!

PR #486, opened 2017-04-04
fix(nav): hide secondary nav when empty

mindreeper2420:
hide the secondary navigation when the active parent does not have any visible child elements

fixes https://github.com/fabric8io/fabric8-ui/issues/454


pmuir:
[test]

PR #490, opened 2017-04-04
Add OpenShift template

vpavlin:
We decided to move OpenShift templates to code repos to enable developer for faster iteration and easier maintainability. Service Delivery team will just keep track of what is deployed where, but all changes will go through this template.

### DIFF
--- a/openshift/fabric8-ui.app.yaml
+++ b/openshift/fabric8-ui.app.yaml
@@ -1,0 +1,89 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: f8ui
+  creationTimestamp: null
+objects:
+- apiVersion: v1
+  kind: DeploymentConfig
+  metadata:
+    creationTimestamp: null
+    generation: 1
+    labels:
+      service: f8ui
+    name: f8ui
+  spec:
+    replicas: 1
+    selector:
+      service: f8ui
+    strategy:
+      resources: {}
+      rollingParams:
+        intervalSeconds: 1
+        maxSurge: 25%
+        maxUnavailable: 25%
+        timeoutSeconds: 600
+        updatePeriodSeconds: 1
+      type: Rolling
+    template:
+      metadata:
+        creationTimestamp: null
+        labels:
+          service: f8ui
+      spec:
+        containers:
+        - image: registry.devshift.net/fabric8io/fabric8-ui:${IMAGE_TAG}
+          imagePullPolicy: Always
+          name: f8ui
+          ports:
+          - containerPort: 80
+            protocol: TCP
+          resources: {}
+          terminationMessagePath: /dev/termination-log
+        dnsPolicy: ClusterFirst
+        restartPolicy: Always
+        securityContext: {}
+        terminationGracePeriodSeconds: 30
+    test: false
+    triggers:
+    - type: ConfigChange
+  status: {}
+- apiVersion: v1
+  kind: Service
+  metadata:
+    creationTimestamp: null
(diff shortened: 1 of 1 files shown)


GitCop:
There were the following issues with your Pull Request

 * Commit: f88c7c6ed2eb88bcfb5e1343cb2e956e6f7bf3dc
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

Guidelines are available at https://github.com/fabric8io/fabric8-ui

---
This message was auto-generated by https://gitcop.com

rawlingsj:
FWIW during UI CI pipelines we use generated and released fabric8-ui YAML.  This means you don't need to maintain it as much and can be used in the CI+CD pipelines.  If you want I'm sure you could reuse this also https://github.com/fabric8io/fabric8-online/tree/master/apps/fabric8-ui

rawlingsj:
i.e. you get versioned YAML like this http://central.maven.org/maven2/io/fabric8/online/apps/fabric8-ui/1.0.75/fabric8-ui-1.0.75-openshift.yml

vpavlin:
@rawlingsj This looks like something we should be using in dsaas cluster, but it seems we are still using the one from housekeeping...

One issue we were trying to fix was that we didn't want to have a single repo containing all the templates, which we found is hard if developers need to make change in code which needs to be taken into account in the template. This kind of seems to be the case for fabric8-online repo, so not sure if we are actually solving anything by switching to this model here - why isn't the generation done from fabric8-ui repo?

rawlingsj:
we do tend to include the java project to generate the yaml when the repo is java based for example https://github.com/fabric8io/generator-backend/tree/master/src/main/fabric8

But not always people like a pom and java project in say an npm repo.  Its totally up to them though.

Personally I'd prefer to have a consistent approach and reuse what we've been doing for all the upstream and OSO tenant apps.  Maintaining openshift yaml by hand is really hard going, especially when upgrading major api versions or adding a new label for example.  We generate both kubernetes and openshift JSON + YAML it's also the same tooling that users use in their java quickstarts so it goes back to the dog fooding concept.

I'm just trying to save you a lot of pain :) we started off down this route and quickly realised maintaining YAML by hand is a pain.   The fabric8-maven-plugin generates the default annotations and labels again so apps are consistent and will work nicely in the runtime console too.

rawlingsj:
Also being a developer using this approach for the last 2 years it has allowed me and a community to iterate fast.  The Che developers seem to be finding it ok as well. https://github.com/fabric8io/fabric8-online/pulls?utf8=%E2%9C%93&q=is%3Apr%20author%3Al0rd%20.  It might be worth looking at.

vpavlin:
Yeah, I get the benefits. I am not against generation, I am only trying to figure out how to make things consistent - I'd like to avoid special cases, so the options from my PoV are:

1. Maintain full templates in a single repo - that's pain, we don't want that
2. Maintain full templates in project repos - should be easier than above, we just need to keep track of what is deployed where, which should be simple
3. Generate templates from a single repo - we "force" everyone to use fabric8-online repo, then tracking is easy, but it makes it more complicated for developer to find the template (I think)
4. Generate templates from project repos - IMO the ideal state, but as you said, people might be reluctant to f.e. get pom.xml in their golang project

@kbsingh What do you think?

I am probaly too old school and like option `2.` the most:D

pmuir:
[test]

vpavlin:
We had email conversation with @rawlingsj and I'd prefer to get this merged due to lack of time to pursue other options.

PR #700, opened 2017-04-12
fix(logout): only redirect after logout completes

dgutride:
This prevents the logout GET from being canceled due to page redirect


fabric8cd:
@dgutride snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-700.int.rdu2c.fabric8.io

PR #766, opened 2017-04-17
fix(logout): redirect to logout url

dgutride:


fabric8cd:
@dgutride snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-766-fabric8-ui.int.rdu2c.fabric8.io

PR #873, opened 2017-04-19
fix(codebases): GitHub service is broken

dlabrecq:
A GitHub token was recently added to the GitHub service; however, users are no longer able to add codebases. The new switchMap code fails to add the appropriate header.

### DIFF
--- a/src/app/create/codebases/services/github.service.ts
+++ b/src/app/create/codebases/services/github.service.ts
@@ -1,10 +1,8 @@
-import { LoginService } from './../../../shared/login.service';
 import { Injectable } from '@angular/core';
 import { Headers, Http } from '@angular/http';
-import { AuthenticationService, UserService } from 'ngx-login-client';
+import { AuthenticationService } from 'ngx-login-client';
 import { Logger } from 'ngx-base';
 import { Observable } from 'rxjs';
-import { cloneDeep } from 'lodash';
 
 import { Context, Contexts } from 'ngx-fabric8-wit';
 import {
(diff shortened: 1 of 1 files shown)


fabric8cd:
@dlabrecq snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-873-fabric8-ui.int.rdu2c.fabric8.io

src/app/create/codebases/services/github.service.ts
```diff
@@ -22,26 +20,26 @@ import {
...
+      private http: Http,
+      private logger: Logger,
+      private authService: AuthenticationService) {
     this.contexts.current.subscribe(val => this.context = val);
     this.gitHubUrl = 'https://api.github.com';
     this.cache = new Map();
+
+    authService.gitHubToken.map(token => {
+      this.headers.set('Authorization', `token ${token}`);
+    });
```
pmuir: I don't think this will work as you don't subscribe to the gitHubToken - if it's getting called, it's by chance...

src/app/create/codebases/services/github.service.ts
```diff
@@ -67,9 +65,7 @@ export class GitHubService {
     if (this.cache.has(url)) {
       return this.cache.get(url);
     } else {
-      let res = this.getHeaders(this.getUserName(fullName))
```
pmuir: This code is very dangerous like this, as you have no guarantee that the token is fetched before you start trying to use it. You need to make sure that you have the token before you start trying to use it, which is what the previous code was doing.

pmuir:
I've checked the build generated by the CD system above and we're now not using an authorization header.

src/app/create/codebases/services/github.service.ts
dlabrecq: Can you please suggest how to fix the previous code to ensure the authentication header is set?

PR #950, opened 2017-04-20
fix(usermenu): set fixed width of user menu

mindreeper2420:
set fixed width of the user menu and dropdown

### DIFF
--- a/src/app/header/header.component.ts
+++ b/src/app/header/header.component.ts
@@ -14,8 +14,6 @@ import { Navigation } from './../models/navigation';
 import { MenuItem } from './../models/menu-item';
 import { DummyService } from './../shared/dummy.service';
 
-
-
 interface MenuHiddenCallback {
   (headerComponent: HeaderComponent): Observable<boolean>;
 }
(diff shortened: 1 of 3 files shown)


src/app/header/header.component.ts
```diff
@@ -14,12 +14,6 @@ import { Navigation } from './../models/navigation';
 import { MenuItem } from './../models/menu-item';
 import { DummyService } from './../shared/dummy.service';
 
-
-
-interface MenuHiddenCallback {
```
joshuawilson: You can't remove this, it breaks the build. Or else you need to find all the places that call it and change them.

mindreeper2420: I don't recall doing anything with that, so I'll check the log again.

mindreeper2420: @joshuawilson pushed changes and ran tests - all passed.

fabric8cd:
@mindreeper2420 snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-950-fabric8-ui.int.rdu2c.fabric8.io

Issue #983, opened 2017-04-22
After logging in on prod-preview.openshift.io, user's home page is stuck at a developer survey


ldimaggi:
The user is able to navigate to other pages (such as 'view all spaces'), but the user's home page is always the survey:

![screenshot from 2017-04-22 08-48-27](https://cloud.githubusercontent.com/assets/642621/25304647/a320773a-2739-11e7-9634-3b0d2b0681dd.png)

dlabrecq:
Fixed by https://github.com/fabric8io/fabric8-ui/pull/987

ldimaggi:
Not being seen now - closing

PR #987, opened 2017-04-22
fix(home): Developer poll fix

dlabrecq:
Home page was inadvertently routed to the poll component because its child route path is empty.

Fixes issue #983


dlabrecq:
Fixes #983

fabric8cd:
@dlabrecq snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-987-fabric8-ui.int.rdu2c.fabric8.io

PR #1032, opened 2017-04-25
add a link to the openshift console on the pipelines page

jstrachan:


fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1032-fabric8-ui.int.rdu2c.fabric8.io

GitCop:
There were the following issues with your Pull Request

 * Commit: d73208c1f415fe80fe5f0114a2105b248c1a1b76
  * Your subject line is longer than 50 characters

Guidelines are available at https://github.com/fabric8io/fabric8-ui

---
This message was auto-generated by https://gitcop.com

fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1032-fabric8-ui.int.rdu2c.fabric8.io

joshuawilson:
[test]

config/webpack.dev.js
```diff
@@ -23,6 +23,7 @@ const HOST = process.env.HOST || 'localhost';
 const PORT = process.env.PORT || 3000;
 const HMR = helpers.hasProcessFlag('hot');
 // if env is 'inmemory', the inmemory debug resource is used
+const OPENSHIFT_CONSOLE_URL = process.env.OPENSHIFT_CONSOLE_URL || 'https://console.starter-us-east-2.openshift.com/console/';
```
joshuawilson: This is not the OpenShift that matches prod-preview.

config/webpack.prod.js
```diff
@@ -31,6 +31,7 @@ const OfflinePlugin = require('offline-plugin');
  * Webpack Constants
  */
 const ENV = process.env.NODE_ENV = process.env.ENV = 'production';
+const OPENSHIFT_CONSOLE_URL = process.env.OPENSHIFT_CONSOLE_URL || 'https://console.starter-us-east-2.openshift.com/console/';
```
joshuawilson: This should be set through the env var on prod.

config/webpack.dev.js
jstrachan: I specifically set it to prod as the pre-prod environment is completely unusable for the forseable future

config/webpack.prod.js
jstrachan: or to say that another way - we should always use the only working environment, prod, unless someone wants to point at something else. Right now there's only the prod OSO anyways

config/webpack.dev.js
joshuawilson: but the other links are to prod-preview

jstrachan: prod-preview right now is only for the SaaS - it doesn't have a suitable OSO or OSO console

jstrachan: if we had a pre-prod OSO that had any value, I'd definitely point to it outside of prod

joshuawilson: my point is that we need the other links to point to prop for this to work in dev mode

jstrachan: yeah - in dev I'm using this script right now in dev:
https://github.com/fabric8io/fabric8-ui/blob/master/environments/openshift-prod-cluster.sh

ideally we'd have a pre-prod ;). Then the next best thing is having pre-prod SaaS services that all point to the prod KeyCloak and OSO so at least we've pre-prod for forge + planner + core SaaS stuff

pmuir:
This should really be using runtime rather than build config. Will do a rewrite to that later on today.

pmuir:
Replaced by #1062

PR #1107, opened 2017-04-26
fix(approve): add missing env var

jstrachan:
lets link the approve UI to the forge API for now until a cleaner fix is made


fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1107-fabric8-ui.int.rdu2c.fabric8.io

PR #1109, opened 2017-04-26
fix(logging): remove logging

jstrachan:
its a bit noisy

### DIFF
--- a/src/app/create/pipelines/pipelines.component.ts
+++ b/src/app/create/pipelines/pipelines.component.ts
@@ -167,7 +167,7 @@ export class PipelinesComponent implements OnInit, OnDestroy {
           });
       })
       .subscribe(val => {
-        console.log('Updating build configs:', val);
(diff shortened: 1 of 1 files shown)


fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1109-fabric8-ui.int.rdu2c.fabric8.io

joshuawilson:
[test]

joshuawilson:
If it is a bit noisy, that tells me there is a memory leak we should fix instead of ignore by removing the logging.

jstrachan:
Neither log statements add any actual value - they are old debugging statements.

plus when a pipeline is pending/building/waiting approval you get 4 lines per second which makes the JavaScript console pretty hard to use. 

I can imagine it using up a fair bit of memory too - as it forces the browser to keep around at least a copy of every full BuildConfig object tree for every second!

PR #1158, opened 2017-04-27
fix(cards): fix truncation and overflow

mindreeper2420:
fixes the truncation and overflow issues when viewing on different resolutions


mindreeper2420:
@joshuawilson this fixes the overflow and truncation issues

mindreeper2420:
Note - I cannot get the pipelines card to show anything, even though my environment has a pipeline in it.

PR #1161, opened 2017-04-27
fix(version): update fabric8-planner to 0.23.44

fabric8cd:


joshuawilson:
merged another one that passed

PR #1173, opened 2017-04-27
fix(card): update Environments card

mindreeper2420:
Update the Environments card to be called Applications and adjust text.


fabric8cd:
@mindreeper2420 snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1173-fabric8-ui.int.rdu2c.fabric8.io

PR #1239, opened 2017-05-02
fix(): Getting started username read only

dlabrecq:
Disabled username change for summit, but require "Save" to set registration_complete=true.

- This will ensure the getting started page is not shown again. 
- This will also ensure that the username cannot be updated later, avoiding potential issues with a nonexistent username.


fabric8cd:
@dlabrecq snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1239-fabric8-ui.int.rdu2c.fabric8.io

PR #1240, opened 2017-05-02
fix(): Save username automatically

dlabrecq:
This will save the username automatically when the user visits the getting started page. However, users may jump right to the home page if their tokens are already linked (e.g., via Keycloak).


fabric8cd:
@dlabrecq snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1240-fabric8-ui.int.rdu2c.fabric8.io

PR #1250, opened 2017-05-04
fix(getting-started): Fix broken form tag

dlabrecq:
Due to recent edits, the form tag is not closed properly. The resulting errors in the page prevent login.


corinnekrych:
@dlabrecq @joshuawilson similar to the PR I did yesterday: 
https://github.com/fabric8io/fabric8-ui/pull/1248
do we want the form tag commented or not?

PR #1296, opened 2017-05-12
fix(environments): add environments page

jstrachan:
lets bring the environments page back again as its all lazily loaded now so doesn't perform too many network requests on startup any more


fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1296-fabric8-ui.int.rdu2c.fabric8.io

PR #1325, opened 2017-05-18
fix(runtime env vars): added runtime API URLs

### LINKED ISSUE #1316: load environment variables from the container at runtime
right now we are baking environment variables into the docker image via webpack for things like `FABRIC8_FORGE_API_URL`, `FABRIC8_WIT_API_URL`, `FABRIC8_SSO_API_URL`, `FABRIC8_RECOMMENDER_API_URL` and the [stuff in webpack](https://github.com/fabric8io/fabric8-ui/blob/master/config/webpack.dev.js#L26-L31).

This works great for SaaS hosted consoles in, say, pre-prod and prod. But it doesn't work at all for on premise or upstream fabric8 distros where we don't know these values until after we've installed the application (since they depend on the local cluster's URLs/routes/nodeports and whatnot).

The values are constants; but they should be loaded as environment variables the server side as the container starts up - rather than being baked into the JavaScript thats loaded.

BTW we hacked around this using relative domain logic in the code; but thats not very reliable for on premise as its common for routers on OpenShift to use names like 'foo-namespace.whatnot.com' rather than 'foo.namespace.whatnot.com'

I wonder whats the cleanest way to fix this in a way that minimises network chatter on startup so that for on premise and upstream we can make these values be loaded from the server side on startup using actual container environment variables.

e.g. could we combine webpack with some kind of server side templating so that the values can be replaced either at docker build time or at load time; so the values are embedded in the webpack JS as the app loads? Or maybe we just add something like this to the `index.html`

```html
<script>
process.ENV.FOO = "${FOO}";
process.ENV.BAR = "${BAR}";
</script>
```
and have the environment variables populated by nginx templating in the `index.html`? Or some separate `env.js` file?

@pmuir @joshuawilson got any bright ideas? Am hoping for a simple solution that avoids breaking things while keeping things flexible so we can use real environment variables

jstrachan:
lets add runtime environment variables for the API URLs for when running on
    premise or upstream fixes #1316

### DIFF
--- a/src/app/shared/api-locator.service.ts
+++ b/src/app/shared/api-locator.service.ts
@@ -1,4 +1,5 @@
-import { Injectable } from '@angular/core';
+import {Injectable} from "@angular/core";
+import {currentOAuthConfig} from "fabric8-runtime-console/src/app/kubernetes/store/oauth-config-store";
 
 @Injectable()
 export class ApiLocatorService {
@@ -34,7 +35,7 @@ export class ApiLocatorService {
   }
 
   get witApiUrl(): string {
-    return this.buildApiUrl('wit');
+    return this.buildApiUrl('wit', (config) => config.witApiUrl);
   }
 
   get realm(): string {
@@ -42,23 +43,32 @@ export class ApiLocatorService {
   }
 
   get forgeApiUrl(): string {
-    let tmp=this.buildApiUrl('forge');
+    let tmp=this.buildApiUrl('forge', (config) => config.forgeApiUrl);
     return tmp
   }
 
   get ssoApiUrl(): string {
-    return this.buildApiUrl('sso');
+    return this.buildApiUrl('sso', (config) => config.ssoApiUrl);
   }
 
   get recommenderApiUrl(): string {
-    return this.buildApiUrl('recommender');
+    return this.buildApiUrl('recommender', (config) => config.recommenderApiUrl);
   }
 
   private loadEnvVar(key: string): void {
     this.envVars.set(key, process.env[this.DEFAULT_API_ENV_VAR_NAMES.get(key)]);
   }
 
-  private buildApiUrl(key: string): string {
+  private buildApiUrl(key: string, fn: (OAuthConfig) => string): string {
+    // for on premise and upstream we need to be able to dynamically discover endpoints
+    // via environment variables on the server side
+    let oauth = currentOAuthConfig();
+    if (oauth != null && fn) {
+      let value = fn(oauth);
+      if (value) {
+        return value;
+      }
+    }
     // Return any environment specified URLs for this API
     if (this.envVars.get(key)) {
       return this.envVars.get(key);
--- a/src/config/oauth.json
+++ b/src/config/oauth.json
@@ -4,6 +4,10 @@
   "api_server_base_path": "{{ .Env.K8S_API_SERVER_BASE_PATH }}",
   "ws_api_server": "{{ .Env.WS_K8S_API_SERVER }}",
   "openshift_console_url": "{{ .Env.OPENSHIFT_CONSOLE_URL }}",
+  "forge_api_url": "{{ .Env.FABRIC8_FORGE_API_URL }}",
+  "wit_api_url": "{{ .Env.FABRIC8_WIT_API_URL }}",
+  "sso_api_url": "{{ .Env.FABRIC8_SSO_API_URL }}",
+  "recommender_api_url": "{{ .Env.FABRIC8_RECOMMENDER_API_URL }}",
   "oauth": {
     "oauth_authorize_uri": "{{ .Env.OAUTH_AUTHORIZE_URI }}",
     "oauth_client_id": "{{ .Env.OAUTH_CLIENT_ID }}",
--- a/environments/local-cluster.sh
+++ b/environments/local-cluster.sh
@@ -36,7 +36,7 @@ export NAMESPACE=`oc project -q`
 
 echo "Configured to connect to kubernetes cluster at https://${PROXIED_K8S_API_SERVER}/ with namespace ${NAMESPACE}"
 
-export FABRIC8_SSO_API_URL=`minishift openshift service keycloak -n ${NAMESPACE} --url`
+export FABRIC8_SSO_API_URL="`minishift openshift service keycloak -n ${NAMESPACE} --url`/"
 export FABRIC8_WIT_API_URL="`minishift openshift service wit -n ${NAMESPACE} --url`/api/"
 export FABRIC8_FORGE_API_URL=`minishift openshift service forge -n ${NAMESPACE} --url`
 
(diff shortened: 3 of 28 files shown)


fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1325-fabric8-ui.int.rdu2c.fabric8.io

fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1325-fabric8-ui.int.rdu2c.fabric8.io

src/config/oauth.json
```diff
@@ -4,6 +4,10 @@
   "api_server_base_path": "{{ .Env.K8S_API_SERVER_BASE_PATH }}",
   "ws_api_server": "{{ .Env.WS_K8S_API_SERVER }}",
   "openshift_console_url": "{{ .Env.OPENSHIFT_CONSOLE_URL }}",
+  "forge_api_url": "{{ .Env.FABRIC8_FORGE_API_URL }}",
```
pmuir: oauth.json is for fabric8-runtime-console, for fabric8-ui please use https://github.com/jstrachan/fabric8-ui/blob/999d4a6b13bc2ee3e059ae78d1b2571dee8b8781/src/config/fabric8-ui.config.json

src/app/shared/api-locator.service.ts
```diff
@@ -33,27 +34,36 @@ export class ApiLocatorService {
...
-    return this.buildApiUrl('recommender');
+    return this.buildApiUrl('recommender', (config) => config.recommenderApiUrl);
   }
 
   private loadEnvVar(key: string): void {
     this.envVars.set(key, process.env[this.DEFAULT_API_ENV_VAR_NAMES.get(key)]);
   }
 
-  private buildApiUrl(key: string): string {
+  private buildApiUrl(key: string, fn: (OAuthConfig) => string): string {
```
pmuir: This should just replace the process.env map, rather than suplement it - no need to have two ways to do the same thing.

src/config/oauth.json
jstrachan: it looks like the code still loads oauth.json and it seems to work there - but will have a look at fabric8-ui.config.json too

src/app/shared/api-locator.service.ts
pmuir: I *think* it should be ok, as in prod/prod-preview we don't use env vars for these, just relative vars. This might well break dev, but that is more easy to test. Given our goal right now is to remove tech debt I would prefer to break prod-preview or even prod and get this right ;-)

src/config/oauth.json
pmuir: It does, but I tried to keep some separation between the codebases, so that as we update the vars needed by the runtime console, we can easily just copy the updated oauth.json over to this codebase and not worry about patching stuff. TBH we should really just be reusing oauth.json from fabric8-runtime-console rather than hold a copy, but not sure how to do that atm, requires some build magic.

src/app/shared/api-locator.service.ts
jstrachan: sold - lets do it! :)

src/config/oauth.json
jstrachan: BTW the initial reason for fabric8-runtime-console was for something upstream if we didn't go with all the planner stuff; but now am thinking its gonna be easier just having the same UI both upstream and for openshift.io - so maybe we just merge fabric8-runtime-console into fabric8-ui? We could always split them apart later on to make them more modular (would require moving all the shared common stuff into a separate npm module really first)

jstrachan: that would at least help align everything more closely together; so one way to do auth + config?

pmuir: I think merging could make sense @joshuawilson ?

joshuawilson: I'm fine with merging it, but I want to get through this backlog of PRs and fix a few things first.

joshuawilson:
[test]

fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1325-fabric8-ui.int.rdu2c.fabric8.io

fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1325-fabric8-ui.int.rdu2c.fabric8.io

joshuawilson:
[test]

fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1325-fabric8-ui.int.rdu2c.fabric8.io

fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1325-fabric8-ui.int.rdu2c.fabric8.io

joshuawilson:
@pmuir can you please review

fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1325-fabric8-ui.int.rdu2c.fabric8.io

fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1325-fabric8-ui.int.rdu2c.fabric8.io

fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1325-fabric8-ui.int.rdu2c.fabric8.io

fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1325-fabric8-ui.int.rdu2c.fabric8.io

jstrachan:
it turns out its a very large complex change to make loading the env vars async as it changes ngx-login-client + fabric8-runtime-console too.

So I'm experimenting with a simpler solution in this PR
https://github.com/fabric8io/fabric8-ui/pull/1401

which requires eagerly loading a blob of JSON so that we can avoid doing async `Observable<T>` stuff which lets us use env var values with the injection stuff (which most code is using today)

jstrachan:
gonna close this PR as the approach is complex and doesn't yet work in favour of PR #1401 which is much simpler and seems to work fine ;)

fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1325-fabric8-ui.int.rdu2c.fabric8.io

src/config/oauth.json
joshuawilson: @jstrachan are you still good to have the runtime-console merged into fabric8-ui?

jstrachan: yes! once we've merged it all we can remove that old fabric8-runtime-console oauth.json stuff too

PR #1401, opened 2017-05-25
fix(runtime env vars): lets load env vars

jstrachan:
lets try another approach to #1316 by loading eagerly a blob of JSON on startup so that the env vars can be injected without blocking via @Inject which simplifies the code changes to all the fabric8-ui modules

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -77,7 +77,7 @@ import { AnalyticService }               from './shared/analytics.service';
 import { ApiLocatorService }             from './shared/api-locator.service';
 import { AuthGuard }                     from './shared/auth-guard.service';
 import { authApiUrlProvider }            from './shared/auth-api.provider';
-import { fabric8UIConfigProvider }       from './shared/config/fabric8-ui-config.service';
+import {fabric8UIConfigProvider, syncFabric8UIConfigProvider}       from './shared/config/fabric8-ui-config.service';
 import { AuthUserResolve }               from './shared/common.resolver';
 import { ContextService }                from './shared/context.service';
 import { ContextCurrentUserGuard }       from './shared/context-current-user-guard.service';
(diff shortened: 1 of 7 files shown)


fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1401-fabric8-ui.int.rdu2c.fabric8.io

pmuir:
+1 to this basic approach, but I didn't do a detailed review.

fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1401-fabric8-ui.int.rdu2c.fabric8.io

jstrachan:
@pmuir @joshuawilson fancy having a look at this PR - its seeming to work way better than the previous approach. 

The main issue is to work with all the gazillions of `@Inject(FOO_API_URL) someUrl: string` code thats littered throughout our codebases in this repo, `ngx-login-controller` and `fabric8-runtime-console` we can't have lazy loading of the `Fabric8UIConfig` object; so instead I load a [fabric8-ui.env.js](https://github.com/jstrachan/fabric8-ui/blob/43a097c635f5d161ca0990d1d9ea75c44a818b8d/src/config/fabric8-ui.env.js) file early in the [index.html](https://github.com/jstrachan/fabric8-ui/blob/43a097c635f5d161ca0990d1d9ea75c44a818b8d/src/index.ejs#L15) so that we can implement a simple synchronous provider of `Fabric8UIConfig` via [this code](https://github.com/fabric8io/fabric8-ui/pull/1401/commits/43a097c635f5d161ca0990d1d9ea75c44a818b8d#diff-9e0521b941b38a0a9cdfa15f55bc7034R24)

if you're happy with this approach (which on balance seems the simplest thing that could possibly work and avoids the startup timing issues of async loading of injected static strings), I can re-implement the `Observable<Fabric8UIConfig>` using the same underlying provider and remove the `fabric8-ui.config.json` file which is no longer required

fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1401-fabric8-ui.int.rdu2c.fabric8.io

joshuawilson:
I'm going to merge this, the issues can be fixed in another commit

PR #1448, opened 2017-06-06
fix(space): filter pipelines on space

jstrachan:
when viewing pipelines which have a space label lets filter out any
    pipelines not matching the current space


fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1448-fabric8-ui.int.rdu2c.fabric8.io

PR #1459, opened 2017-06-06
fix: only use the sync runtime var stuff

pmuir:


fabric8cd:
@pmuir snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1459-fabric8-ui.int.rdu2c.fabric8.io

PR #1492, opened 2017-06-11
fix(test): removed code that encounteed unpretictable results - will …

ldimaggi:
…add back in a separate more extensive scenario test


fabric8cd:
@ldimaggi snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1492-fabric8-ui.int.rdu2c.fabric8.io

PR #1504, opened 2017-06-13
fix(status): add new status list     lets add the new status list widget

jstrachan:


src/app/app.module.ts
```diff
@@ -30,7 +30,8 @@ import {
   PipelineModule,
   OnLogin,
   DevNamespaceScope,
-  NamespaceScope
+  NamespaceScope,
+§§  StatusListModule, SpaceNamespace,
```
joshuawilson: These extra characters are failing the build.

jstrachan: whoops butter fingers! Thanks for the heads up - fix committed...

fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1504-fabric8-ui.int.rdu2c.fabric8.io

PR #1519, opened 2017-06-15
polish docs

jstrachan:
lets add the new status list widget

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,25 +3,24 @@
 [![Build Status](https://ci.centos.org/buildStatus/icon?job=devtools-fabric8-ui-npm-publish-build-master)](https://ci.centos.org/job/devtools-fabric8-ui-npm-publish-build-master)
 [![codecov](https://codecov.io/gh/fabric8io/fabric8-ui/branch/master/graph/badge.svg)](https://codecov.io/gh/fabric8io/fabric8-ui)
 
-## Before you start
+## Before you start: Pointing to a cluster
 
-### Backend API
+You need to setup your shell to point to the right cluster so that it can talk to the required back end services like KeyCloak, WIT, Forge, OpenShift etc.
 
-Make sure you set the URL to the services. For UI development, we recommend connecting to the dev environment API server. 
+We provide various sample environments out of the box which make it easier to get started. They are all located as bash scripts in `environments`. 
 
-To connect to the dev environment instances: 
-* `export FABRIC8_WIT_API_URL="http://api.prod-preview.openshift.io/api/"`
-* `export FABRIC8_RECOMMENDER_API_URL="http://api-bayesian.dev.rdu2c.fabric8.io/api/v1/"`
-* `export FABRIC8_FORGE_API_URL="https://forge.api.prod-preview.openshift.io"`
-* `export FABRIC8_REALM="fabric8"`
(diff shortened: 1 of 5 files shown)


fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1519-fabric8-ui.int.rdu2c.fabric8.io

fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1519-fabric8-ui.int.rdu2c.fabric8.io

README.md
```diff
@@ -3,25 +3,24 @@
...
+You need to setup your shell to point to the right cluster so that it can talk to the required back end services like KeyCloak, WIT, Forge, OpenShift etc.
 
-Make sure you set the URL to the services. For UI development, we recommend connecting to the dev environment API server. 
+We provide various sample environments out of the box which make it easier to get started. They are all located as bash scripts in `environments`. 
 
-To connect to the dev environment instances: 
-* `export FABRIC8_WIT_API_URL="http://api.prod-preview.openshift.io/api/"`
-* `export FABRIC8_RECOMMENDER_API_URL="http://api-bayesian.dev.rdu2c.fabric8.io/api/v1/"`
-* `export FABRIC8_FORGE_API_URL="https://forge.api.prod-preview.openshift.io"`
-* `export FABRIC8_REALM="fabric8"`
```
joshuawilson: we still need to know how to set this to work with a local running version of the core

jstrachan: wouldn't we use the pre-prod environment script though? The reason I removed that section is setting those few env vars is no longer enough really. Plus those env vars are not about using a local WIT - its about using pre-prod which doesn't really work any more anyway

src/config/fabric8-ui.env.js
```diff
@@ -3,9 +3,9 @@ window.Fabric8UIEnv = {
   "analyticsWriteKey": "{{ .Env.ANALYTICS_WRITE_KEY }}",
   "forgeApiUrl": "{{ .Env.FABRIC8_FORGE_API_URL }}",
   "openshiftConsoleUrl": "{{ .Env.OPENSHIFT_CONSOLE_URL }}",
+  "openshiftProxiedApiServer": "{{ .Env.PROXIED_K8S_API_SERVER }}",
   "witApiUrl": "{{ .Env.FABRIC8_WIT_API_URL }}",
   "ssoApiUrl": "{{ .Env.FABRIC8_SSO_API_URL }}",
-  "forgeApiUrl": "{{ .Env.FABRIC8_FORGE_API_URL }}",
```
joshuawilson: I think this is being referenced multiple times, they would need to point to the new name.

README.md
joshuawilson: ok, we just need to set it then in the scripts

src/config/fabric8-ui.env.js
jstrachan: the line I removed was a duplicate (that you'd spotted in an early PR :) both keys were `forgeApiUrl` so I think we can safely delete this one?

joshuawilson: ah, thanks. missed that

PR #1548, opened 2017-06-20
fix(navbar): fix pipeline history nav bar

jstrachan:
selection should check for sub path matches if no direct match found and
    use reverse order to avoid matching 'Codebases' first (or the first tab in
    general when a more specific URL matches) fixes
    https://github.com/fabric8io/fabric8-ui/issues/870


fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1548-fabric8-ui.int.rdu2c.fabric8.io

PR #1589, opened 2017-06-23
fix(multiselect): fix multi-select widget for new import git repos

### LINKED ISSUE #1581: Add to Space Wizard: Import repositories to use multi-selection
if we change the back end to use multi-selection the UI (`UISelectMany`) the UI is much slicker and lets folks pick the repos they want to import using multi-selection and filtering - which also avoids accidentally importing all of them too :)

fabric8

the problem is the `Next` step then sends no selections for the `gitReposityPatterns` property so we can't implement the wizard on the server side.

fabric8

so we need to tweak the UI to handle this one

jstrachan:
when the selection changes we need to update the field.value so that we
    pass the selected values into the forge wizard on Next fixes #1581


fabric8cd:
@jstrachan snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1589-fabric8-ui.int.rdu2c.fabric8.io

PR #1622, opened 2017-07-03
fix(release): no need to use v prefix as it is already included in th…

rawlingsj:
…e tag


fabric8cd:
@rawlingsj snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1622-fabric8-ui.badger.fabric8.io

joshuawilson:
[test]

PR #1623, opened 2017-07-03
feat(preboot loading screen): added OpenShift.io logo with progress bar during preboot

kylebuch8:
Removed the Loading... text and replaced it with an inline OpenShift.io svg and an indeterminate
progress bar. Once the app loads, the HTML and style tag that were added will be replaced with the
booted app

screen shot 2017-07-03 at 12 18 00 pm


fabric8cd:
@kylebuch8 snapshot fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1623-fabric8-ui.badger.fabric8.io

joshuawilson:
[test]

joshuawilson:
Why include the SVG file and the data from the file in the html?

kylebuch8:
Including the SVG directly makes for the fastest load. If you included it as an img, then it would require a network request. Since it's just a preboot screen, I was trying to keep it simple and only require that CSS be downloaded before we show the loader. Really, the whole point is for speed to paint.

PR #1676, opened 2017-07-08
feat(dynamic page titles): Updating the title tag based on the page

### LINKED ISSUE #1377: HTML <Titles> should be according to the page content
HTML <Titles> should be according to the page content 
The Title Tag <title></title> should contain appropriate title with respect to page 
For example : if The page is for space dashboard , in that case the title should contain "space dashboard" keyword into it. 

This is minor enhancement in the pages

kylebuch8:
When a page finishes loading, update the title tag of the app to reflect the title of the current
page. If no title is found, just show the default branding name

fixes #1377

### DIFF
--- a/src/app/shared/branding.service.ts
+++ b/src/app/shared/branding.service.ts
@@ -1,0 +1,38 @@
+import { Injectable } from '@angular/core';
+import { Fabric8UIConfig } from '../shared/config/fabric8-ui-config';
+
+import openshiftLogo from '../../assets/images/OpenShift-io_logo.png';
+import fabric8Logo from '../../assets/images/fabric8_logo.png';
+
+@Injectable()
+export class BrandingService {
+  private isFabric8: boolean;
+
+  constructor(
+    private fabric8UIConfig: Fabric8UIConfig,
+  ) {
+    this.isFabric8 = this.fabric8UIConfig.branding && this.fabric8UIConfig.branding === "fabric8";
+    console.log('isFabric8', this.isFabric8);
+  }
+
+  get name(): string {
+    return this.isFabric8 ? "fabric8.io" : "OpenShift.io";
+  }
+
+  get logo(): string {
+    return this.isFabric8 ? fabric8Logo : openshiftLogo;
+  }
+
+  get backgroundClass(): string {
+    return this.isFabric8 ? "home-fabric8-background-image" : "home-header-background-image";
+  }
+
+  get description(): string {
+    return this.isFabric8 ? "A free, end-to-end, cloud-native development experience." : "A free, end-to-end, cloud-native development experience.";
+  }
+
+  get moreInfoLink(): string {
+    return this.isFabric8 ? "https://fabric8.io/" : "https://openshift.io";
+  }
+
+}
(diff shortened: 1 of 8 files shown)


kylebuch8:
I created a branding service that could be used elsewhere in the app notably in the recent changes in the home.component.ts module.

fabric8cd:
@kylebuch8 fabric8/fabric8-ui:SNAPSHOT-PR-1676-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1676-fabric8-ui.badger.fabric8.io

src/app/shared/branding.service.ts
```diff
@@ -0,0 +1,38 @@
...
+  private isFabric8: boolean;
+
+  constructor(
+    private fabric8UIConfig: Fabric8UIConfig,
+  ) {
+    this.isFabric8 = this.fabric8UIConfig.branding && this.fabric8UIConfig.branding === "fabric8";
+    console.log('isFabric8', this.isFabric8);
+  }
+
+  get name(): string {
```
joshuawilson: This feels a bit odd to me that you aren't using a private variable for these Getter methods.

joshuawilson:
[test]

joshuawilson:
Maybe I'm missing something but what does the branding service have to do with the page titles?

kylebuch8:
I built the branding service so it could replace the brandInformation stuff in home.component.ts. Looking through that code, I needed the same logic that setup the brandInformation.name value (fabric8.io or OpenShift.io). So I created the branding service so we don't have the logic in two places and then I used the name value from the branding service and append it to the title of the page. 

For example: 
Analyze - OpenShift.io or Analyze - fabric8.io
Profile - OpenShift.io or Profile - fabric8.io

Make sense?

src/app/shared/branding.service.ts
kylebuch8: I just updated this to private variables

fabric8cd:
@kylebuch8 fabric8/fabric8-ui:SNAPSHOT-PR-1676-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1676-fabric8-ui.badger.fabric8.io

joshuawilson:
[test]

src/app/shared/branding.service.ts
```diff
@@ -0,0 +1,58 @@
+import { Injectable } from '@angular/core';
+import { Fabric8UIConfig } from '../shared/config/fabric8-ui-config';
+
+import openshiftLogo from '../../assets/images/OpenShift-io_logo.png';
+import fabric8Logo from '../../assets/images/fabric8_logo.png';
+
+@Injectable()
+export class BrandingService {
+  private isFabric8: boolean;
+  private _logo: string;
```
joshuawilson: Based on this configuration, these private variable are now READONLY. They can only be set by the constructor. If we ever need to set this, they will need Setter methods.

kylebuch8: But *should* they ever be set from the outside? I thought all of these variables should be set based off of the environment.

joshuawilson: Eventually we will want to allow the user to set their own branding on an on-prem version. So these setting could be set from a config file or stored some place else. I don't consider the constructor as the "environment", but I do agree they should be set by their environment. :)

Issue #1758, opened 2017-07-20
Codebases: Layout problem with "open workspace" dropdown


ibuziuk:
![image](https://user-images.githubusercontent.com/1461122/28417732-6ce06962-6d59-11e7-95eb-e008d19d98c6.png)

joshuawilson:
@ibuziuk what is the problem and what do you expect it to do?

ibuziuk:
@joshuawilson the layout looks broken for me. Previously dropdown with available workspaces was per repository, now it is located in the table header

dlabrecq:
If you refresh the page, the combobox should be rendered in the appropriate row where it belongs.

I did some digging and work-item-widget is overriding the .combobox CSS selector -- it's positioned absolutely and sets the 'top' property. I'm guessing that's why it's at the top of the container.

I'm going to set the specificity of that selector so it won't effect anything else.

dlabrecq:
Fixed by https://github.com/fabric8-ui/fabric8-ui/pull/1775

PR #1775, opened 2017-07-21
Fix specificity of .combobox selector override

dlabrecq:
Fixes https://github.com/fabric8-ui/fabric8-ui/issues/1758.

The combobox selector was recently overridden for work-item-widget. This uses the ‘top’ property to position the select element absolutely. As a result, the select element used on the codebases page is incorrectly positioned at the top of its container. 

The codebases select element should appear in each row of the list component, not in the header.

### DIFF
--- a/src/app/home/work-item-widget/work-item-widget.component.html
+++ b/src/app/home/work-item-widget/work-item-widget.component.html
@@ -3,7 +3,7 @@
     <div class="card-pf-heading f8-card-heading">
       <h2 class="card-pf-title">My work items
       <span *ngIf="spaces.length !== 0">
-        <select class="col-xs-3 col-sm-6 combobox form-control"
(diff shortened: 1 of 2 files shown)


fabric8cd:
@dlabrecq fabric8/fabric8-ui:SNAPSHOT-PR-1775-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1775-fabric8-ui.badger.fabric8.io

mindreeper2420:
@dlabrecq does the class `.combobox` do anything? If not, it could just be simplified to `.work-item-combobox`

dlabrecq:
@mindreeper2420 I was wondering about .combobox myself? I found it used in three places; the codebases page, wizard, and work-item-widget. I left it there just in case.

I suspect .combobox is really meant to be used with the Bootstrap combobox, not a select element. We could probably revisit this and clean up, like you suggested?

PR #1779, opened 2017-07-24
fix(kubernetes): fixed k8s getting started

jstrachan:
lets poll the fabric8-tenant service until the kubernetes tenant is connected
as the creation of the jenkins external URL is lazy and only then can we register it into KC
so that the user can login to jenkins via KC

### DIFF
--- a/src/app/getting-started/getting-started.component.ts
+++ b/src/app/getting-started/getting-started.component.ts
@@ -7,6 +7,10 @@ import { AuthenticationService, UserService, User } from 'ngx-login-client';
 
 import { ExtUser, GettingStartedService } from './services/getting-started.service';
 import { ProviderService } from './services/provider.service';
+import {Fabric8UIConfig} from "../shared/config/fabric8-ui-config";
+import {Observable} from "rxjs/Observable";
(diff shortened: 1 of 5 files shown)


fabric8cd:
@jstrachan fabric8/fabric8-ui:SNAPSHOT-PR-1779-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1779-fabric8-ui.badger.fabric8.io

src/app/getting-started/getting-started.component.ts
```diff
@@ -232,4 +261,78 @@ export class GettingStartedComponent implements OnDestroy, OnInit {
       type: type
     } as Notification);
   }
+
+  /**
+   * Lets poll the fabric8-tenant service to see if the kubernetes tenant has connected the
+   * services such as Jenkins into KeyCloak which is typically lazy after the tenant
+   * starts to be created
+   */
+  private kubeTenantConnectPoll() {
```
joshuawilson: @jstrachan this does not appear to be related to the getting started page or service. I'm going to guess you put it here so it would get run first. If that is the case can you please create a service for this and let's find a better place for it.

PR #1870, opened 2017-08-16
Import runtime console

joshuawilson:
This will bring the runtime-console into the main UI. 
It also updates it to use LESS instead of SASS.


joshuawilson:
@rawlingsj why is this failing? Any ideas?

PR #1897, opened 2017-08-18
feat(cleanup): add support for tenant reset/cleanup

dgutride:
Add a new cleanup component (linked to from the user profile page).  This shows all spaces, deletes them after sufficient warning for the user and calls the new tenant cleanup API (which deletes all openshift objects).  

Upon success, the tenant reset API is called which adds back Jenkins and Che support.

### DIFF
--- a/src/app/profile/cleanup/cleanup.component.html
+++ b/src/app/profile/cleanup/cleanup.component.html
@@ -1,0 +1,114 @@
+<div id="overview" class="container-fluid content padding-top-15">
+  <div class="row">
+    <div class="col-xs-8 col-sm-8 col-sm-offset-1">
+      <div *ngIf="showNotification" class="alert {{notificationClass}}">
+        <span class="pficon {{notificationIcon}}"></span>
+        <strong>{{notificationTitle}}</strong> {{notificationText}}
+      </div>
+      <p *ngIf="cleanupStatus === 'notstarted'"><strong>You are about to erase all of your OpenShift.io spaces</strong>, reverting your OpenShift.io environment to it's default state.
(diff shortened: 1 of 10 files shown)


fabric8cd:
@dgutride fabric8/fabric8-ui:SNAPSHOT-PR-1897-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1897-fabric8-ui.badger.fabric8.io

src/app/profile/cleanup/cleanup.component.html
```diff
@@ -0,0 +1,114 @@
...
+                name="username"
+                [(ngModel)]="userName" />
+            </div>
+          </div>
+        </form>
+      </div>
+    </section>
+    <footer>
+      <div class="col-md-offset-4 col-md-8">
+        <button class="btn" [ngClass]="{ 'btn-danger': userNameMatches() }" type="button" (click)="confirm()" [disabled]="!userNameMatches()" type="submit"> I understand my actions - erase my environment</button>
```
joshuawilson: You have 2 "type" in this line. A button and a submit.

src/app/profile/services/tenent.service.ts
```diff
@@ -42,6 +42,22 @@ export class TenentService {
       });
   }
 
+  /**
+   * Cleanup tenant
+   * @returns {Observable<T>}
```
joshuawilson: Should the "T" be "any", just to keep the docs clear.

joshuawilson:
We are going to need to clean out the spaces that are deleted from the context menu and the list of work items that were in the space that was deleted.
I'm not allowed to click on them but they are there. 
However they go away with a hard refresh.

PR #1913, opened 2017-08-25
fix(acceptancetests): add teamVersion

jstrachan:
lets add a team version too so we can configure that in addition to jenkins & che versions in the acceptance tests in PRs

### DIFF
--- a/src/app/profile/tenant/tenant.component.html
+++ b/src/app/profile/tenant/tenant.component.html
@@ -42,6 +42,14 @@
           </div>
         </div>
         <div class="form-group">
+          <label for="teamVersion" class="control-label">Team version</label>
+          <div [ngClass]="{'has-error': teamVersionInvalid}">
+            <input type="text" class="form-control" id="teamVersion" name="teamVersion"
+                   placeholder="Team version"
+                   [(ngModel)]="teamVersion">
+          </div>
+        </div>
+        <div class="form-group">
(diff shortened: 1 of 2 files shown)


fabric8cd:
@jstrachan fabric8/fabric8-ui:SNAPSHOT-PR-1913-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1913-fabric8-ui.badger.fabric8.io

src/app/profile/tenant/tenant.component.html
```diff
@@ -42,6 +42,14 @@
           </div>
         </div>
         <div class="form-group">
+          <label for="teamVersion" class="control-label">Team version</label>
```
joshuawilson: @jstrachan where does this text appear in the UI?

jstrachan: on  the advanced OSIO developer/tester page: https://openshift.io/my-user-name/_tenant - its invisible to everyone who doesn't know that magic URL

one day we may even make that magic URL invisible to non 'OSIO testers'

joshuawilson: ah, thanks

jstrachan: background to this 'hidden' page here: https://github.com/fabric8-services/fabric8-tenant#testing-yaml

PR #1921, opened 2017-08-27
fix(status): Update app label used for jenkins

aslakknutsen:
"app" label has changed from "jenkins-openshift" to "jenkins"

Fixes openshiftio/openshift.io#665

This fixes the above problem, but unsure if it should be fixed in UI or in the templates?


fabric8cd:
@aslakknutsen fabric8/fabric8-ui:SNAPSHOT-PR-1921-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1921-fabric8-ui.badger.fabric8.io

PR #1935, opened 2017-08-30
fix(code): add code quality checks to build process

mindreeper2420:
This PR adds a proper stylelint configuration file and a proper htmlhint configuration file to fabric8-ui. These files are run as part of the build process.

Any failures to meet the various criteria as defined in the configuration files will fail the build.

**Note:** The runtime-console and wizard are not included as part of the linting/hint checks. A separate PR is needed to fix the multitude of issues in those folders.

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -27,6 +27,7 @@ const NormalModuleReplacementPlugin = require('webpack/lib/NormalModuleReplaceme
 const ProvidePlugin = require('webpack/lib/ProvidePlugin');
 const ScriptExtHtmlWebpackPlugin = require('script-ext-html-webpack-plugin');
 const UglifyJsPlugin = require('webpack/lib/optimize/UglifyJsPlugin');
+const StyleLintPlugin = require('stylelint-webpack-plugin');
 // const TsConfigPathsPlugin = require('awesome-typescript-loader');
 
 /*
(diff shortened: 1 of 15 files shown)


config/webpack.common.js
```diff
@@ -159,6 +160,19 @@ module.exports = function (options) {
...
         },
 
+        /* HTML Linter
+         * Checks all files against .htmlhintrc
+        */
+        {
+          enforce: 'pre',
+          test: /\.html$/,
+          loader: 'htmlhint-loader',
+          exclude: [/node_modules/,/src\/a-runtime-console/,/src\/app\/space\/wizard/],
```
corinnekrych: @mindreeper2420 since you use `StyleLintPlugin` with it ignore file do we need to specify the excluded directories here too?

mindreeper2420: @corinnekrych this is specifically for the HTML files, while the stylelint ignore file is only handling the `.less` files. Removing these ignores will throw errors in the build, as well as force each build to check all of our node modules.

corinnekrych: +1 thanks for the explanation!

fabric8cd:
@mindreeper2420 fabric8/fabric8-ui:SNAPSHOT-PR-1935-6 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-1935-fabric8-ui.badger.fabric8.io

joshuawilson:
How do I run the style linter and the html linter. As a dev I need to be able to run these before building so I can verify that I did everything correctly. Then they can be run in the builds.

mindreeper2420:
@joshuawilson I will include these steps as part of a README update. They are not npm commands, and require additional configuration.

PR #1977, opened 2017-09-02
fix(version): update ngx-base to 1.3.0

fabric8cd:


dgutride:
@joshuawilson - we are getting a ton of duplicates on these automated PR's - just stating the obvious, do you want me to open an issue for it?

PR #2001, opened 2017-09-13
fix(cards): remove padding overrides from cards

mindreeper2420:
remove the padding overrides from cards, returning them to default PatternFly style

related to https://github.com/fabric8-ui/fabric8-ux/issues/661


joshuawilson:
[test]

PR #2013, opened 2017-09-21
fix(applications): import icons

jstrachan:
lets import the icons so that the URLs work when using webpack image resources with dynamic file names


fabric8cd:
@jstrachan fabric8/fabric8-ui:SNAPSHOT-PR-2013-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2013-fabric8-ui.badger.fabric8.io

PR #2118, opened 2017-10-17
fix for patternfly 3.28

edewit:
fix for issue https://github.com/openshiftio/openshift.io/issues/1138 @corinnekrych

### DIFF
--- a/src/app/space/forge-wizard/quickstart-wizard.component.less
+++ b/src/app/space/forge-wizard/quickstart-wizard.component.less
@@ -1,0 +1,3 @@
+/deep/ .pfng-wizard-position-override {
+  overflow: auto;
+}
--- a/src/app/space/forge-wizard/quickstart-wizard.component.ts
+++ b/src/app/space/forge-wizard/quickstart-wizard.component.ts
@@ -9,7 +9,8 @@ import { CodebasesService } from '../create/codebases/services/codebases.service
 
 @Component({
   selector: 'quickstart-wizard',
-  templateUrl: './quickstart-wizard.component.html'
+  templateUrl: './quickstart-wizard.component.html',
+  styleUrls: ['./quickstart-wizard.component.less']
 })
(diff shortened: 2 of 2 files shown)


corinnekrych:
@edewit as we discussed it, that would do the trick for now. Although https://angular.io/guide/component-styles `deep` is going to be deprecated eventually.

@dlabrecq I think this should be fixed in patternfly-ng. wdyt?

src/app/space/forge-wizard/quickstart-wizard.component.less
```diff
@@ -0,0 +1,3 @@
+/deep/ .pfng-wizard-position-override {
+  overflow: auto;
+}
```
dlabrecq: The height and overflow can be set using a selector you define via the wizard config -- it should be set for both sidebar and main panel. Please see my comments here:
https://github.com/openshiftio/openshift.io/issues/1138#issuecomment-337140549

joshuawilson:
@edewit please address comments if you want me to merge this.

src/app/space/forge-wizard/quickstart-wizard.component.less
edewit: @dlabrecq This suggestion doesn't work, because the overflow must be on the parent of the main panel. The main panel already has `overflow: auto` set.

corinnekrych:
@dlabrecq @edewit @joshuawilson I've tried your suggested approach as per the comment on the issue.
see the branch https://github.com/fabric8-ui/fabric8-ui/compare/master...corinnekrych:wizard.css.not.working?expand=1
It does not work. this is an issue I've asked you before... it might be css encapsulation but the custom css is not applied.
Try it an propose a fix that work pls.

corinnekrych:
@dlabrecq @edewit @joshuawilson closing this on in favor of https://github.com/fabric8-ui/fabric8-ui/pull/2146

PR #2178, opened 2017-10-30
chore(build): fix the Dockerfile to be functional

debloper:
Fix the base `Dockerfile` to be useful and functional

This also makes the `Dockerfile.deploy` somewhat unnecessary, except
- it's already used in various orchestration and CI/CD
- it uses a different/custom nginx image with certain tag

So for that reason, I haven't removed the file. But unless the custom nginx image is necessary, that file can be removed and the main Dockerfile can be renamed to Dockerfile.deploy.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,0 @@
-# Builds a Docker to deliver dist/
-FROM nginx:latest
-COPY dist/ /usr/share/nginx/html
--- a/Dockerfile.deploy
+++ b/Dockerfile.deploy
@@ -1,11 +1,8 @@
 FROM fabric8/fabric8-openshift-nginx:vd83b3a1
-MAINTAINER "Pete Muir <pmuir@bleepbleep.org.uk>"
 
 USER root
 
-RUN rm -rf /usr/share/nginx/html/
+RUN rm -rf /usr/share/nginx/html
 COPY dist /usr/share/nginx/html
-RUN chmod -R 777 /usr/share/nginx/html/
-
 
 USER ${FABRIC8_USER_NAME}
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -13,6 +13,7 @@ services:
       - default
   ui:
     build: .
+    dockerfile: Dockerfile.deploy
     entrypoint:
       - /usr/local/bin/wait-for-it.sh
       - core:8080


fabric8cd:
@debloper fabric8/fabric8-ui:SNAPSHOT-PR-2178-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2178-fabric8-ui.badger.fabric8.io

Dockerfile
```diff
@@ -1,3 +1,6 @@
 # Builds a Docker to deliver dist/
 FROM nginx:latest
-COPY dist/ /usr/share/nginx/html
+USER root
```
kwk: I don't know where this Dockerfile is being used. If it is for testing, then this is fine, but if it is running in production I would advice to not run nginx as root.

joshuawilson: I think on this repo is it used for CICO. @kbsingh can you confirm?

kwk: @joshuawilson I don't think so. See https://github.com/fabric8-ui/fabric8-ui/search?utf8=%E2%9C%93&q=Dockerfile&type= . We only use the Dockerfile.builder and Dockerfile.deploy.

debloper: Like in the other dockerfile, the user can be switched back (and should be IMO, I'm unsure whether to use env_var like the other Dockerfile (Dockerfile.deploy) or the username itself.

The reason to use root is to have enough priv to remove and write content to `/usr/share/nginx/`

Please let me know what you guys think of it and I'll update the PR.

kwk: @deb please tell us where the Dockerfile is used first before we answer your questions.

kwk: @deb ah, we have a docker-compose file that uses it: https://github.com/fabric8-ui/fabric8-ui/blob/master/docker-compose.yaml#L15

vpavlin:
Yeah, it does not seem to be used in CICO context, so feel free to do whatever you want with it:). I don't expect UI to do something requiring root privileges, but just to keep things consistent - maybe switching to some non-root user might be good practice

debloper:
PR updated according to the feedback.

- Removed the old `Dockerfile.deploy`
- Added the necessary steps/layers to `Dockerfile`
- Renamed `Dockerfile` to `Dockerfile.deploy` (so that it's readily usable by cico builds)
- Updated docker-compose file to use the new file as there's no default `Dockerfile` anymore

Dockerfile.deploy
```diff
@@ -1,11 +1,9 @@
...
 
-RUN rm -rf /usr/share/nginx/html/
+RUN rm -rf /usr/share/nginx/html
 COPY dist /usr/share/nginx/html
-RUN chmod -R 777 /usr/share/nginx/html/
+RUN chown -R fabric8:fabric8 /usr/share/nginx/html
 
-
-USER ${FABRIC8_USER_NAME}
+USER fabric8
```
kwk: Why not use `${FABRIC8_USER_NAME}` as is done [here](https://github.com/fabric8-ui/fabric8-ui/search?utf8=%E2%9C%93&q=FABRIC8_USER_NAME&type=)?

debloper: That env_var is set by the `Dockerfile.builder`. In cases where you want to run the `Dockerfile.deploy` in host without the `Dockerfile.builder` setting up that var (manually update some content in `/dist` post-build for testing, build locally without builder image or something else) then it'll work just fine, as it's primary goal is just to create an image by putting `/dist` contents in `nginx` doc-root.

So the upside is it doesn't need to be dependent on the another Dockerfile to operate.
Downside is, none actually.

> We should IMHO keep a dedicated Dockerfile for deployments and one for building. This way we can control what goes into what build.

That's how it is.

fabric8cd:
@debloper fabric8/fabric8-ui:SNAPSHOT-PR-2178-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2178-fabric8-ui.badger.fabric8.io

debloper:
Update: Root user is needed for nginx container to run properly; and as running nginx server is _only_ what we're doing here, leaving it up as root user shouldn't be an issue. Following is one of the various possible situations faced when trying to run nginx without superuser:

```
fabric8-planner-platform | nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
fabric8-planner-platform | nginx: [emerg] mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
fabric8-planner-platform exited with code 1
```

This can be easily verified/reproduced by running it with or without root user:
<pre>
docker run -it <b>--user=root</b> -p 8080:80 --name=fabric8-planner-platform fabric8-planner-platform
</pre>

The previous `Dockerfile.deploy` worked around some of that by setting permission 777 (a+rwx) to the content, but from security perspective, that's not a better solution (meaning, the content is breachable even without root user).

Given the scenario, I think running this deploy container is perfectly valid (against common/best practices) as the root user.

Following up with another commit to remove the step to switch to non-priv user.

fabric8cd:
@debloper fabric8/fabric8-ui:SNAPSHOT-PR-2178-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2178-fabric8-ui.badger.fabric8.io

debloper:
> - You can't run as root on openshift

Do you mean it's (an enforced/technical) limitation or good-common-practice?

> - Removing the parent image removes the OSO Cors proxy

If we're [talking of this](https://github.com/fabric8-ui/fabric8-ui-openshift-nginx/blob/master/Dockerfile#L26), then I can work on adding the necessary layers on top of the new Dockerfile to make it work just like that, while using/building from latest `nginx` base image. Should there be any problem?

aslakknutsen:
> Do you mean it's (an enforced/technical) limitation or good-common-practice?

It's a technical enforced good common practice.

"This prevents root actions such as chown or chmod from being run and is a sensible security precaution as, should a user be able to perform a local exploit to break out of the container, then they would not be running as root on the underlying container host."

> if we're talking of this..

https://github.com/fabric8-ui/fabric8-ui-openshift-nginx/blob/master/root/etc/nginx/nginx.conf#L79

Also, this new image is missing things like json format access logging

> @i don't think we should be sticking to a private/inaccessible and old base image (registry.centos.org/kbsingh/openshift-nginx:1.10.2).

Private/inaccessible? I think you're confusing this with 'custom'. The Dockerfile is publicly available on GitHub and the built image publically available from our repository.

We also have a requirement from SD to run CentOS(soon RHEL) based images in prod. "nginx:latest" does not follow that.

debloper:
> It's a technical enforced good common practice. No public facing OS Cluster will allow you to run as root.

At the risk of taking the discussion sidewise, we're using the root user **inside the container** which is different than running something on a cluster as root. Also the root user inside a container is different and limited compared to host root user by what it's capable of doing (which is why the `--cap-add` flag exists). So, this seems to be superstitious practice at best.

Especially when we leave our doc-roots with 777 permissions (one of many vulnerabilities which are already in there, and are more severe than the one in the focus of discussion), meaning even non-root users can do significant damage. Which means ANY user having access to the container already compromises the integrity of the container, root or not.

Please give me an example of vulnerabilities we may run into, to prove that I'm just a dumb guy with a large mouth. Otherwise this looks like a tradition at best, instead of security best practice.

---

> Private/inaccessible? I think you're confusing this with 'custom'.

Any Dockerfile built from a base image is a custom Dockerfile, on top of the base image. Which is an integral part of the containers' layering system.

What I meant was, the registry for the base image `http://registry.centos.org` is publicly inaccessible and documentation around what's the objective and tactics (compared to the official nginx image) is missing or is undiscoverable. If you can point me towards where I can get the answers to that, it'll help.

If custom layering is needed (including, centos/rhel as base image), we can do that, as long as what are all the problems we're targeting/solving are clear and transparent.

aslakknutsen:
> doc-roots with 777 permissions

That any user can read/write to the filesystem of an immutable/disposable container is a strange argument for running the whole thing as root... Your solution to your brought up security issue is to give the user even wider rights?

FIx the permissions

> Please give me an example of vulnerabilities we may run into, to prove that I'm just a dumb guy with a large mouth. Otherwise this looks like a tradition at best, instead of security best practice.

You personally assessment is completely irrelevant.

    No public facing OS Cluster will allow you to run as root.


>  the registry for the base image http://registry.centos.org is publicly inaccessible 

```
$ docker pull registry.centos.org/kbsingh/openshift-nginx:1.10.2
Trying to pull repository registry.centos.org/kbsingh/openshift-nginx ... 
sha256:67c330a824a0242f2d617c3d9bcd85661b5a71430a9c3d156df4fdfb91db0de0: Pulling from registry.centos.org/kbsingh/openshift-nginx
66881a2148ed: Already exists 
bbcc78f8def8: Pull complete 
0beff9780015: Pull complete 
17a6f284b362: Pull complete 
aad9be34ed7a: Pull complete 
6398397e84f5: Pull complete 
624eaf58bc97: Pull complete 
f62f8a3be2e9: Pull complete 
Digest: sha256:67c330a824a0242f2d617c3d9bcd85661b5a71430a9c3d156df4fdfb91db0de0
Status: Downloaded newer image for registry.centos.org/kbsingh/openshift-nginx:1.10.2
```

> is missing or is undiscoverable

Read the Dockerfile and nginx config.

> If custom layering is needed (including, centos/rhel as base image)

There is no IF

debloper:
First and foremost, you earned my respect for not taking the bait and handling the situation better than I expected and set it up for. Kudos.

So, here's what I'm going to do:

- Use `registry.centos.org/kbsingh/openshift-nginx:1.10.2` > `fabric8/fabric8-openshift-nginx:vd83b3a1` base image
- Switch the user back to non-priv user

Please confirm whether that's good enough (I know you've mentioned these two points in your initial review, just need confirmation that they're all there are in need of a revert).

Now the topic of if using root user inside a container realistically widens the security risk - maybe someday we can catch back on that over a beer.

New commit incoming in a bit.

Dockerfile.deploy
```diff
@@ -1,11 +1,6 @@
-FROM fabric8/fabric8-openshift-nginx:vd83b3a1
-MAINTAINER "Pete Muir <pmuir@bleepbleep.org.uk>"
+FROM nginx:latest
```
aslakknutsen: Using :latest is bad practice and cause indeterministic behavior as what the image contains depends on when it was built causing irreproducible builds and you have given up control over which version you're using in your production env.

aslakknutsen:
fabric8/fabric8-openshift-nginx:vd83b3a1 is missing parts of the nginx config if I remember correctly.

debloper:
> fabric8/fabric8-openshift-nginx:vd83b3a1 is missing parts of the nginx config if I remember correctly.

I'm not sure which missing parts are you talking of, and should I have to fix that in this commit itself. FWIW, the previous image (that I'm changing) is working in CI/CD, so I'm inclined to leave that as is and fix that as a follow up if the need be.

aslakknutsen:
> FWIW, the previous image (that I'm changing) is working in CI/CD,

That's not worth much at all. The image is never used, only built. 

> I'm not sure which missing parts are you talking of, 

Sorry, my bad. I mixed the images. Was thinking it was this one registry.centos.org/kbsingh/openshift-nginx:1.10.2

debloper:
Alright, updated patch is up!

> That's not worth much at all. The image is never used, only built.

`Dockerfile` was not used elsewhere (removing, hence), but `Dockerfile.deploy` is used both in [cico](https://github.com/fabric8-ui/fabric8-ui/blob/master/cico_build_deploy.sh#L70) and [f8ci](https://github.com/fabric8-ui/fabric8-ui/blob/1a5f8964a0ad350f64d706152f5cf00393399e59/release.groovy#L28) both places. Or, do you mean that the built images are never pushed anywhere for real usage. I'm guessing the later?

Either way, please review this and merge if all is well. Addressed both the points mentioned.

fabric8cd:
@debloper fabric8/fabric8-ui:SNAPSHOT-PR-2178-6 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2178-fabric8-ui.badger.fabric8.io

debloper:
@aslakknutsen please update review

joshuawilson:
What is the status on this?

debloper:
@joshuawilson can be aborted.

The ideal course of action would be to send a PR to [CentOS Dockerfiles](https://github.com/CentOS/CentOS-Dockerfiles) repo for an official CentOS-nginx image; then document what modifications/customization needed for our purpose, and then build a custom image from the official base image and use it. But,

PR #2011, opened 2017-09-20
fix(auth): auth service for all auth and user API endpoints

sbose78:
Note to the reviewer : 
please use the PR title as the commit message after squashing commits during merge.

- Use `auth.openshift.io` for login+user calls instead of `api.openshift.io`
- update login path from `/api/login/authorize` to `/api/login`

This PR has to be merged with-or-after https://github.com/fabric8-ui/ngx-login-client/pull/86 .

### What this PR does ?
- Introduces the usage of auth service endpoints for existing user/token api calls.
- Reads Auth API url from the environment variable.
- The default is https://prod-preview.openshift.io 
- When tested with https://github.com/fabric8-ui/ngx-login-client/pull/86  , the token for github url would no longer be fetched from keycloak but from the auth service.
- When tested with https://github.com/fabric8-ui/ngx-login-client/pull/86  , the refresh token would be fetched from `https://auth.prod-preview.openshift.io/api/token/refresh` instead of  `https://api.prod-preview.openshift.io/api/login/refresh` .
 ### How to test ?
- Checkout https://github.com/fabric8-ui/ngx-login-client/pull/86 , build it and do an `npm link` from the location where the current fabric8-ui branch has been checked out.
- Do an `npm start`
- Verify that all `/api/user` related urls are constructed with the auth service domain. ( except for /api/user/services ) 
- Verify that github token is not fetched from keycloak.

PR #2193, opened 2017-10-31
WIP : Link github using the native auth service API

sbose78:
Must be merged *after* https://github.com/fabric8-ui/fabric8-ui/pull/2011  - merged now.

At the moment,
We will support 'connecting'  `github` the new way and support connecting `openShift-v3` the old way
Because we haven't switched openshift-v3 to the new 'Get token' endpoint  - which is because of the OSO proxy work that is in progress.

There are 3 scenarios:

1. User chooses to connect **only** github , and so the UI makes a POST call to the new API to connect github account with form parameters
- the access token 
- for=https://github.com
( Note: we got rid of the need to pass client_session etc in this new API )

2. User chooses to connect **only** openshift-v3 
The UI does this the old/existing way by redirection.

3. User chooses both github and openshift-v3
The UI does a form POST as (1) , and in the redirect param , the url for (2) is passed.


joshuawilson:
cc @dlabrecq @dgutride so that you are aware of changes here

src/app/getting-started/getting-started.component.ts
```diff
@@ -144,14 +170,44 @@ export class GettingStartedComponent implements OnDestroy, OnInit {
...
+        let parsedToken = jwt_decode(this.auth.getToken());
+      
+        // code repition, but temporary.
+        // Will soon migrate this to use the new way of 
+        // linking using the auth service.
+        let url = `${this.oldLinkingServer}/link/session?`
+          + "clientSession=" + parsedToken.client_session
+          + "&sessionState=" + parsedToken.session_state
+          + "&redirect=" + finalRedirect // brings us back to Getting Started.
+          + "&provider=openshift-v3";
```
dlabrecq: Looks like this code probably belongs in the prvider service, not in the getting started compponent?

alexeykazakov:
https://github.com/fabric8-ui/fabric8-ui/pull/2011 is merged. We now can move forward with this PR.

fabric8cd:
@sbose78 fabric8/fabric8-ui:SNAPSHOT-PR-2193-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2193-fabric8-ui.badger.fabric8.io

src/app/getting-started/getting-started.component.ts
sbose78: Thanks @dlabrecq , moved it to `provider service`

fabric8cd:
@sbose78 fabric8/fabric8-ui:SNAPSHOT-PR-2193-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2193-fabric8-ui.badger.fabric8.io

fabric8cd:
@sbose78 fabric8/fabric8-ui:SNAPSHOT-PR-2193-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2193-fabric8-ui.badger.fabric8.io

src/app/getting-started/services/provider.service.ts
```diff
@@ -63,6 +65,21 @@ export class ProviderService {
     this.redirectToAuth(url);
   }
 
+  /**
+   * 
+   * @param url 
+   */
+
+   getLegacyLinkingUrl(provider: string, redirect: string): string{
```
dlabrecq: It's not clear why getLegacyLinkingUrl was created? It's almost identical to the link method -- why not just change that instead to use the correct URL params?

src/app/getting-started/getting-started.component.ts
dlabrecq: I see a lot of new code, to support a form submit, but existing code has not been refactored or removed.

src/app/getting-started/getting-started.component.html
```diff
@@ -113,16 +114,18 @@ <h2 class="circle">1</h2>
...
                 <label for="authOpenShift" class="required-pf">OpenShift.com</label>
               </div>
             </ng-template>
           </div>
           <div class="form-group">
-            <button class="btn btn-lg btn-default" id="connect"
-                    [disabled]="isConnectAccountsDisabled"
-                    (click)="connectAccounts()">Connect accounts</button>
+              <input type="hidden"  name="token" [(ngModel)]="accessToken"/>
+              <input type="hidden"  name="redirect" [(ngModel)]="redirect"/>
```
dlabrecq: Should we be hiding an access token in plain text?

src/app/getting-started/getting-started.component.ts
```diff
@@ -47,9 +57,20 @@ export class GettingStartedComponent implements OnDestroy, OnInit {
...
       private userService: UserService) {
 
+    // will be used to connect OpenShift Online.
+    this.oldLinkingServer = apiUrl;
+
+    // will be used in the non-angular html FORM's ACTION during the POST request
+    // to connect GitHub
+    this.linkingApiURL = apiUrl + 'token/link';
+
+    this.redirect = window.location.origin + "/_gettingstarted?wait=true";
```
dlabrecq: All this new code appears to support a form submit. Why is a form submit preferable to using the existing provider service to generate requests? 

We created that service to keep that code out of the getting started component and keep the HTML relatively clean. For example, we don't need to hide the access token and redirect URL in the HTML template.

sbose78: Thanks for the review @dlabrecq - I updated the description above to show the use-cases this change is handling.

> All this new code appears to support a form submit. Why is a form submit preferable to using the existing provider service to generate requests?

The new account linking API gets rid of the need to pass `clientSession` & `sessionState` info for linking. We've simplified the API by requiring only the access token to be passed. Passing the access token in a GET request might break things since the access token is long. That's why we need the form POST.

> For example, we don't need to hide the access token and redirect URL in the HTML template.
Agreed, however we need it in the HTML because of scenario (3) in the description above.
In this scenario the user is trying to connect OSO and Github together. So, the new github is linked by a form POST and the old OSO is linked by the redirect after connecting github.

src/app/getting-started/services/provider.service.ts
sbose78: I didn't want to touch and remove existing code because I'm unsure what the fate of the GettingStarted page would finally be :)

src/app/getting-started/getting-started.component.ts
dlabrecq: Thanks. Makes more sense now. 

Although, we could create a simple POST request using code (from codebases.service) similar to this:

this.http
&nbsp;&nbsp;.post(url, payload, { headers: this.headers })
&nbsp;&nbsp;.map(response => {
&nbsp;&nbsp;return response.json().data as Codebase;
&nbsp;&nbsp;});

That said, the getting started page is going away soon. Not sure these details will matter? However, it would be preferable to have this functionality in a service, so we can call it from any page without a lot of setup in the HTML template -- probably from the user profile.

sbose78: Thanks @dlabrecq ,
That was the initial attempt to keep the component html off the excessive code but the issue is that the POST request returns a 303 which gets intercepted ( & rejected by the browser because its a redirect to github ). So that's where I realized than an AJAX POST with a 303 redirect can only be done using an html form post.

https://stackoverflow.com/questions/19952518/303-redirection-not-working-with-angular-http-post

fabric8cd:
@sbose78 fabric8/fabric8-ui:SNAPSHOT-PR-2193-6 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2193-fabric8-ui.badger.fabric8.io

src/app/getting-started/getting-started.component.html
sbose78: Umm. yeah let me see if i can avoid this.. maybe fill/bind the input field just before the form actually gets submitted.

sbose78:
Hi @dlabrecq , are you OK with my approach ( and the explanation ) ?

dlabrecq:
Yes, I better understand why you used a form submit now. Thank you. I don't want to hold up the PR, so I'll defer to @joshuawilson here. 

I'll just say that I have some reservations after seeing the latest UXD designs. Like I said, the getting started page will be removed shortly. Instead, we plan on using this to link accounts from multiple pages, waiting until the account linking is required to proceed. 

Ultimately, we will provide the ability to link accounts from the user profile page. However, we won't force the user to link accounts until they reach a page like the codebases or planner pages. In order to use codebases; for example, the user will be prompted to link their GitHib account from that page -- similar to page that shows empty state. 

My concern is that each one of these pages must embed hidden input variables and perform a form submit, which is a bit complicated to setup. Thus, it would be ideal if the functionality could be added to a service in order to make this easier to implement across the platform?

If we can duplicate the form submit using Http, that would be best. However, if @joshuawilson agrees that a form submit is the way to go, I'm fine with merging.

alexeykazakov:
I would like to add that we need to switch the getting started page to the new linking API ASAP to unblock a few important features like ability to link the same GitHub account to multiple OSIO accounts.
We could do refactoring later when implementing the new account linking flow.

sbose78:
replaced by https://github.com/fabric8-ui/fabric8-ui/pull/2270

PR #2198, opened 2017-11-02
chore(build): update version discovery with `npm view` call

debloper:
RIP `version_number.js`
Mar 2017 - Nov 2017 :sob:


fabric8cd:
@debloper fabric8/fabric8-ui:SNAPSHOT-PR-2198-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2198-fabric8-ui.badger.fabric8.io

joshuawilson:
Why make this change? What made you think of it?

corinnekrych:
Something I've noticed if I run `npm run build:prod` locally I got:
`BUILD_VERSION: '0.0.0-development',`
Running @debloper changes, we got the latest npmorg release ie:
`  BUILD_VERSION: '1.52.6',`
I guess this target is called to get latest released, some other target might be called prior to that one to get somethig else than `0.0.0-development`. Do you know more @joshuawilson ?

debloper:
@corinnekrych because this command reports the actual published version number of the package (ref: https://www.npmjs.com/package/fabric8-ui).

Which makes me realize, this wont work offline (but `npm install` won't work offline as well, so I guess that's a non-issue).

> Why make this change? What made you think of it?

The way it was handled seemed like a workaround for the lack of a better method. It's also always stuck to `0.0.0-development` (unless that's how it's meant to be, as a placeholder version in dev-env - which is the questions I initially had... is it so?).

If so, then hang on and please don't merge the PR yet. I think I'll poke `@npm/npm` once to see if they're ready to fix https://twitter.com/Debloper/status/923878707692896257 and report `npm view` details for `npm view .` from local directory's `package.json` and not from the published package.

joshuawilson:
The use of `0.0.0-development` is for use with semantic release. We have semantic release auto change based on the commit syntax. This is more of a place holder then anything. If that is the only thing you were trying to change then I think we should close this as that is what semantic release is looking for.

debloper:
@joshuawilson the command doesn't do anything different (it'll still fetch the version number that's defined i.e. `0.0.0-development`), it just changes the way it's fetched from a work-around script to a npm command.

Don't merge this PR though, `npm view` is still undercooked and devs are being lazy to do anything about it. Considering sending a PR to npm repo with this fixed. We can revisit this after that.

joshuawilson:
@debloper if you can confirm that `npm view` is safe or that you got that PR accepted with the npm devs, please reopen this.

PR #2200, opened 2017-11-02
fix(docs): Update Help link

mindreeper2420:
Update the Help link to point to the proper documentation URL

fixes https://github.com/openshiftio/openshift.io/issues/912

screen shot 2017-11-02 at 10 26 53 am

### DIFF
--- a/src/app/layout/header/header.component.html
+++ b/src/app/layout/header/header.component.html
@@ -152,6 +152,12 @@
                 </a>
               </li>
               <li class="divider"></li>
+              <li id="header_loggedinHelp">
+                <a href="https://docs.openshift.io" target="top" class="pointer">
+                  Documentation
+                </a>
+              </li>
+              <li class="divider"></li>
               <li id="header_logout">
                 <a *ngIf="loggedInUser" (click)="logout();" class="pointer">
                   Log Out


mindreeper2420:
@rkratky Here is the PR adding the Help link with the proper URL.

fabric8cd:
@mindreeper2420 fabric8/fabric8-ui:SNAPSHOT-PR-2200-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2200-fabric8-ui.badger.fabric8.io

rkratky:
@mindreeper2420, I'm wondering whether "Help" is a good title for the link. We will have online, contextual *help* in the app. So, I think this link should be either "Docs" (maybe a bit too slangy) or "Documentation".

mindreeper2420:
@rkratky that makes sense - I'll update it with "Documentation"

mindreeper2420:
screen shot 2017-11-02 at 11 07 10 am

rkratky:
@mindreeper2420, one more thing... Shouldn't the link indicate that it opens an external site? Or perhaps have `target="_blank"` (though that's ugly, admittedly).

mindreeper2420:
@rkratky IMO, as it is a subset of openshift.io, it really isn't an external site. Ideally, docs.openshift.io would be styled similarly to the parent domain. I'm also not a big fan of adding icons to dropdown, unless absolutely necessary.

With the current implementation, it will open in a new tab, but that can be changed to just a new window, though that loses some context.

rkratky:
@mindreeper2420, right. New tab is cool.

(We're working on the styling. You can see a preview at https://docs.prod-preview.openshift.io/)

fabric8cd:
@mindreeper2420 fabric8/fabric8-ui:SNAPSHOT-PR-2200-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2200-fabric8-ui.badger.fabric8.io

PR #2236, opened 2017-11-15
Add ability to remove codebase

ashumilova:
feat(codebases): add ability to remove codebase from the list

The kabob menu contains Remove item, clicking on which - displays the confirm dialog for codebase
removal. After success deletion - the codebase list is updated

Closes openshiftio/openshift.io#192

screenshot from 2017-11-15 16-34-35
screenshot from 2017-11-15 16-34-44

### DIFF
--- a/src/app/space/create/codebases/codebases-delete/codebase-delete-dialog.component.html
+++ b/src/app/space/create/codebases/codebases-delete/codebase-delete-dialog.component.html
@@ -1,8 +1,6 @@
-<div>
-  <div>You are about to remove a codebase {{codebase.name}} from your space.
-    Please, verify this action by clicking Remove below.</div>
-  <div class="remove-footer pull-right">
-    <button class="btn btn-default" (click)="cancel()">Cancel</button>
-    <button class="btn btn-danger" (click)="confirmDelete()">Remove</button>
-  </div>
+<span>You are about to remove a codebase {{codebase.name}} from your space.
+    Please, verify this action by clicking Remove below.</span>
+<div class="remove-footer pull-right">
+  <button class="btn btn-default" (click)="cancel()">Cancel</button>
+  <button class="btn btn-danger" (click)="confirmDelete()">Remove</button>
 </div>
--- a/src/app/space/create/codebases/codebases-delete/codebase-delete-dialog.component.html
+++ b/src/app/space/create/codebases/codebases-delete/codebase-delete-dialog.component.html
@@ -1,4 +1,4 @@
-<span>You are about to remove a codebase {{codebase.name}} from your space.
+<span>You are about to remove codebase <b>{{codebase.name}}</b> from your space.
     Please, verify this action by clicking Remove below.</span>
 <div class="remove-footer pull-right">
   <button class="btn btn-default" (click)="cancel()">Cancel</button>
--- a/src/app/space/create/codebases/codebases-delete/codebase-delete-dialog.component.html
+++ b/src/app/space/create/codebases/codebases-delete/codebase-delete-dialog.component.html
@@ -1,0 +1,8 @@
+<div>
+  <div>You are about to remove a codebase {{codebase.name}} from your space.
+    Please, verify this action by clicking Remove below.</div>
+  <div class="remove-footer pull-right">
+    <button class="btn btn-default" (click)="cancel()">Cancel</button>
+    <button class="btn btn-danger" (click)="confirmDelete()">Remove</button>
+  </div>
+</div>
--- a/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.less
+++ b/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.less
@@ -1,6 +1,2 @@
 @import (reference) '../../../../../assets/stylesheets/shared/osio.less';
 .margin-right-20 { margin-right: 20px; }
-
-.remove-codebase-dialog .modal-body {
-  padding: 0px 15px;
-}
--- a/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.less
+++ b/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.less
@@ -1,2 +1,6 @@
 @import (reference) '../../../../../assets/stylesheets/shared/osio.less';
 .margin-right-20 { margin-right: 20px; }
+
+.remove-codebase-dialog .modal-body {
+  padding: 0px 15px;
+}
--- a/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
+++ b/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
@@ -17,11 +17,11 @@ import {IModalHost} from "../../../wizard/models/modal-host";
   styleUrls: ['./codebases-item-actions.component.less']
 })
 export class CodebasesItemActionsComponent implements OnDestroy, OnInit {
+  @Input() cheRunning: boolean;
   @Input() codebase: Codebase;
   @Input() index: number = -1;
   @ViewChild('deleteCodebaseDialog') deleteCodebaseDialog: IModalHost;
 
-  cheRunning: boolean = false;
   subscriptions: Subscription[] = [];
   workspaceBusy: boolean = false;
   dialog: Dialog;
@@ -41,13 +41,6 @@ export class CodebasesItemActionsComponent implements OnDestroy, OnInit {
   }
 
   ngOnInit(): void {
-    this.subscriptions.push(this.broadcaster
-      .on('cheStateChange')
-      .subscribe((che: Che) => {
-        if (che !== undefined && che.running === true) {
-          this.cheRunning = true;
-        }
-      }));
   }
 
   // Actions
--- a/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
+++ b/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
@@ -115,7 +115,7 @@ export class CodebasesItemActionsComponent implements OnDestroy, OnInit {
    * Disassociate codebase from current space
    */
   deleteCodebase(): void {
-    this.subscriptions.push(this.codebasesService.delete(this.codebase).subscribe((codebase: Codebase) => {
+    this.subscriptions.push(this.codebasesService.deleteCodebase(this.codebase).subscribe((codebase: Codebase) => {
       this.broadcaster.broadcast('codebaseDeleted', {
         codebase: codebase
       });
--- a/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
+++ b/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
@@ -1,14 +1,13 @@
-import {Component, Input, OnDestroy, OnInit, ViewChild, ViewEncapsulation} from '@angular/core';
+import { Component, Input, OnDestroy, OnInit, ViewChild, ViewEncapsulation } from '@angular/core';
 import { Subscription } from 'rxjs';
 
-import { Che } from '../services/che';
 import { Codebase } from '../services/codebase';
 import { CodebasesService } from '../services/codebases.service';
 import { Broadcaster, Notification, NotificationType, Notifications } from 'ngx-base';
 import { Dialog } from 'ngx-widgets';
 import { WindowService } from '../services/window.service';
 import { WorkspacesService } from '../services/workspaces.service';
-import {IModalHost} from "../../../wizard/models/modal-host";
+import { IModalHost } from '../../../wizard/models/modal-host';
 
 @Component({
   encapsulation: ViewEncapsulation.None,
--- a/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
+++ b/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, OnDestroy, OnInit, ViewEncapsulation } from '@angular/core';
+import {Component, Input, OnDestroy, OnInit, ViewChild, ViewEncapsulation} from '@angular/core';
 import { Subscription } from 'rxjs';
 
 import { Che } from '../services/che';
@@ -8,6 +8,7 @@ import { Broadcaster, Notification, NotificationType, Notifications } from 'ngx-
 import { Dialog } from 'ngx-widgets';
 import { WindowService } from '../services/window.service';
 import { WorkspacesService } from '../services/workspaces.service';
+import {IModalHost} from "../../../wizard/models/modal-host";
 
 @Component({
   encapsulation: ViewEncapsulation.None,
@@ -18,12 +19,12 @@ import { WorkspacesService } from '../services/workspaces.service';
 export class CodebasesItemActionsComponent implements OnDestroy, OnInit {
   @Input() codebase: Codebase;
   @Input() index: number = -1;
+  @ViewChild('deleteCodebaseDialog') deleteCodebaseDialog: IModalHost;
 
   cheRunning: boolean = false;
   subscriptions: Subscription[] = [];
   workspaceBusy: boolean = false;
   dialog: Dialog;
-  showDialog = false;
 
   constructor(
       private broadcaster: Broadcaster,
@@ -86,29 +87,14 @@ export class CodebasesItemActionsComponent implements OnDestroy, OnInit {
    * @param {MouseEvent} event mouse event
    */
   confirmDeleteCodebase(event: MouseEvent): void {
-    event.stopPropagation();
-    this.dialog = {
-      'title': 'Confirm codebase deletion',
-      'message': 'Are you sure you want to deleteCodebase codebase?',
-      'actionButtons': [
-        {'title': 'Confirm', 'value': 1, 'default': false},
-        {'title': 'Cancel', 'value': 0, 'default': true}
-      ]
-    } as Dialog;
-    this.showDialog = true;
+    this.deleteCodebaseDialog.open();
   }
 
   /**
    * Process the click on confirm dialog button.
-   *
-   * @param {number} value
    */
-  onDialogButtonClick(value: number) {
-    // callback from the confirm deleteCodebase dialog
-    if (value === 1) {
-      this.deleteCodebase();
-    }
-    this.showDialog = false;
+  onDeleteCodebase() {
+    this.deleteCodebase();
   }
 
   /**
@@ -116,10 +102,12 @@ export class CodebasesItemActionsComponent implements OnDestroy, OnInit {
    */
   deleteCodebase(): void {
     this.subscriptions.push(this.codebasesService.deleteCodebase(this.codebase).subscribe((codebase: Codebase) => {
+      this.deleteCodebaseDialog.close();
       this.broadcaster.broadcast('codebaseDeleted', {
         codebase: codebase
       });
     }, (error: any) => {
+      this.deleteCodebaseDialog.close();
       this.handleError('Failed to deleteCodebase codebase ' + this.codebase.name, NotificationType.DANGER);
     }));
   }
--- a/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
+++ b/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
@@ -3,7 +3,9 @@ import { Subscription } from 'rxjs';
 
 import { Che } from '../services/che';
 import { Codebase } from '../services/codebase';
+import { CodebasesService } from '../services/codebases.service';
 import { Broadcaster, Notification, NotificationType, Notifications } from 'ngx-base';
+import { Dialog } from 'ngx-widgets';
 import { WindowService } from '../services/window.service';
 import { WorkspacesService } from '../services/workspaces.service';
 
@@ -20,12 +22,15 @@ export class CodebasesItemActionsComponent implements OnDestroy, OnInit {
   cheRunning: boolean = false;
   subscriptions: Subscription[] = [];
   workspaceBusy: boolean = false;
+  dialog: Dialog;
+  showDialog = false;
 
   constructor(
       private broadcaster: Broadcaster,
       private notifications: Notifications,
       private windowService: WindowService,
-      private workspacesService: WorkspacesService) {
+      private workspacesService: WorkspacesService,
+      private codebasesService: CodebasesService) {
   }
 
   ngOnDestroy(): void {
@@ -38,7 +43,7 @@ export class CodebasesItemActionsComponent implements OnDestroy, OnInit {
     this.subscriptions.push(this.broadcaster
       .on('cheStateChange')
       .subscribe((che: Che) => {
-        if (che != undefined && che.running === true) {
+        if (che !== undefined && che.running === true) {
           this.cheRunning = true;
         }
       }));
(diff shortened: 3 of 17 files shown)


fabric8cd:
@ashumilova fabric8/fabric8-ui:SNAPSHOT-PR-2236-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2236-fabric8-ui.badger.fabric8.io

joshuawilson:
[test]

src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
```diff
@@ -76,10 +81,47 @@ export class CodebasesItemActionsComponent implements OnDestroy, OnInit {
...
   /**
+   * Confirmation dialog for codebase removal.
+   *
+   * @param {MouseEvent} event mouse event
+   */
+  confirmDeleteCodebase(event: MouseEvent): void {
+    event.stopPropagation();
+    this.dialog = {
+      'title': 'Confirm codebase deletion',
+      'message': 'Are you sure you want to deleteCodebase codebase?',
```
joshuawilson: `deleteCodebase codebase` is a typo

src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.less
```diff
@@ -1,2 +1,6 @@
 @import (reference) '../../../../../assets/stylesheets/shared/osio.less';
 .margin-right-20 { margin-right: 20px; }
+
+.remove-codebase-dialog .modal-body {
+  padding: 0px 15px;
```
mindreeper2420: remove the trailing `px` from `0px`

src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.less
```diff
@@ -1,2 +1,6 @@
 @import (reference) '../../../../../assets/stylesheets/shared/osio.less';
 .margin-right-20 { margin-right: 20px; }
+
+.remove-codebase-dialog .modal-body {
```
mindreeper2420: break this out as follows:

```
.remove-codebase-dialog {
  .modal-body {
  }
}
```

src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
```diff
@@ -76,10 +81,47 @@ export class CodebasesItemActionsComponent implements OnDestroy, OnInit {
...
 
   /**
+   * Confirmation dialog for codebase removal.
+   *
+   * @param {MouseEvent} event mouse event
+   */
+  confirmDeleteCodebase(event: MouseEvent): void {
+    event.stopPropagation();
+    this.dialog = {
+      'title': 'Confirm codebase deletion',
```
mindreeper2420: Change to 'Remove Codebase'

src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
mindreeper2420: Change to 'You are about to remove a codebase from you Space. Please verify this action by clicking Remove below.'

This statement matches others in OpenShift.io that are used when removing/deleting an item.

src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
```diff
@@ -76,10 +81,47 @@ export class CodebasesItemActionsComponent implements OnDestroy, OnInit {
...
+      'message': 'Are you sure you want to deleteCodebase codebase?',
+      'actionButtons': [
+        {'title': 'Confirm', 'value': 1, 'default': false},
```
mindreeper2420: Change "Confirm" to "Remove" to match usage throughout OpenShift.io

mindreeper2420:
@joshuawilson will removing a Codebase have any ill-effects on other parts of OpenShift.io? Namely Workspaces/Che, pipelines, etc? I believe there was a reason why it was not implemented, but I cannot recall (or find) an issue around it.

src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.module.ts
```diff
@@ -3,23 +3,26 @@ import { CommonModule } from '@angular/common';
...
     BsDropdownModule.forRoot(),
     CommonModule,
+    DialogModule,
     FormsModule,
     TooltipModule.forRoot()
   ],
   declarations: [ CodebasesItemActionsComponent ],
   exports: [ CodebasesItemActionsComponent ],
-  providers: [ BsDropdownConfig, TooltipConfig, WindowService, WorkspacesService ]
+  providers: [ BsDropdownConfig, TooltipConfig, WindowService, WorkspacesService, CodebasesService ]
```
joshuawilson: Please list alphabetically. Also we now have enough in this list to break it out per line.

src/app/space/create/codebases/codebases.component.ts
```diff
@@ -59,6 +59,12 @@ export class CodebasesComponent implements OnDestroy, OnInit {
       .subscribe((codebase: Codebase) => {
         this.updateCodebases();
       }));
+
+    this.subscriptions.push(this.broadcaster
+      .on('codebaseDeleted')
+      .subscribe((codebase: Codebase) => {
```
joshuawilson: `codebase` appears to be unused, why is it included?

src/app/space/create/codebases/services/codebases.service.ts
```diff
@@ -110,6 +112,24 @@ export class CodebasesService {
       });
   }
 
+  /**
+   * Delete codebase
+   *
+   * @param {Codebase} codebase codebase to delete
+   * @returns {Observable<Codebase>}
+   */
+  delete(codebase: Codebase): Observable<Codebase> {
```
joshuawilson: `delete` is a reserved word in Typescript and so should not be used alone. Can you can it to something like `deleteCodebase`. 

Can you please update the unit test for this function.

joshuawilson:
@mindreeper2420 that is a good question. We might need to add a conditional to verify that the codebase isn't being used. 
@catrobson do we have a user flow on codebases?

src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.less
ashumilova: Dialogs were taken from planner ones - https://github.com/fabric8-ui/fabric8-planner/blob/master/src/app/components/work-item-list-entry/work-item-list-entry.component.html#L2
Will update to proper ones, thank you for the links.

ashumilova:
@mindreeper2420 , @joshuawilson the codebase removal should trigger the related workspaces removal as well (https://github.com/fabric8-services/fabric8-wit/issues/1749). The confirm message could contain that information as well. I think that could be enough.

fabric8cd:
@ashumilova fabric8/fabric8-ui:SNAPSHOT-PR-2236-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2236-fabric8-ui.badger.fabric8.io

joshuawilson:
[test]

src/app/space/create/codebases/services/codebases.service.ts
```diff
@@ -110,6 +112,24 @@ export class CodebasesService {
       });
   }
 
+  /**
+   * Delete codebase
+   *
+   * @param {Codebase} codebase codebase to delete
+   * @returns {Observable<Codebase>}
+   */
+  deleteCodebase(codebase: Codebase): Observable<Codebase> {
```
joshuawilson: Please add a unit test to the service.spec.ts.

src/app/space/create/codebases/codebases.component.ts
```diff
@@ -268,6 +274,7 @@ export class CodebasesComponent implements OnDestroy, OnInit {
           this.codebases.unshift({} as Codebase); // Add empty object for row header
           this.applyFilters(this.appliedFilters);
         }
+        this.startIdleChe();
```
dlabrecq: I understand why we're starting Che here, but it's too late at this point. 

We call updateCodebases when the codebaseDeleted event is receieved. And if we cannot update, we call startIdelChe. However,  updateCodebases will never be called again -- the codebases list won't be updated.

We may need to test if Che is running before calling updateCodebases? Or, add something in the page to ping Che periodically and keep it alive?

joshuawilson:
Now that we have a pattern for delete confirmation, it should be used.

src/app/space/create/codebases/codebases.component.ts
ashumilova: startIdleChe - fetches the Che status and propagates it to other components (if status is not running - starts che). Otherwise, now any update of the list of codebases (add or delete) leads to a bug , when workspaces component and menu - are not displayed, though the che is running, cause all components were reseted to default state - (cheRunning - false). What can be done here for the code to be more explicit - call method smth like - fetchCheState/Status.
Attaching screenshot of the screen after adding new codebase:
screenshot from 2017-11-17 14-23-51

ashumilova:
@dlabrecq save filter deletion feature looks really cool, but there we perform delete action on item, here we have action itself. 
I mean codebases kebab menu has action items:
screenshot from 2017-11-17 15-31-50
filter sample shows list of items and ability to perform deletion on it:
screenshot from 2017-11-17 15-32-00

joshuawilson:
@catrobson can you have someone update the design if we are going to use the new pattern?

@ashumilova can you create a new issue for the delete menu pattern and we can take care of it there.

src/app/space/create/codebases/services/codebases.service.ts
ashumilova: done https://github.com/fabric8-ui/fabric8-ui/pull/2236/files#diff-6037b30f05a1462d22651009442ba9a8R251

fabric8cd:
@ashumilova fabric8/fabric8-ui:SNAPSHOT-PR-2236-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2236-fabric8-ui.badger.fabric8.io

ashumilova:
Here is the issue https://github.com/openshiftio/openshift.io/issues/1408

joshuawilson:
[test]

joshuawilson:
@mindreeper2420 @dlabrecq is there something that still needs to be done for you to sign off?

src/app/space/create/codebases/codebases-delete/codebase-delete-dialog.component.html
```diff
@@ -0,0 +1,8 @@
+<div>
```
mindreeper2420: this `<div>` is not needed

src/app/space/create/codebases/codebases-delete/codebase-delete-dialog.component.html
```diff
@@ -0,0 +1,8 @@
...
+  <div>You are about to remove a codebase {{codebase.name}} from your space.
```
mindreeper2420: would it make more sense for this to either be a `<p>` or a `<span>`?

src/app/space/create/codebases/services/codebases.service.spec.ts
```diff
@@ -7,6 +7,7 @@ import { WIT_API_URL } from 'ngx-fabric8-wit';
 import { CodebasesService } from './codebases.service';
 import { Codebase } from './codebase';
 import { cloneDeep } from 'lodash';
+import {error} from "shelljs";
```
mindreeper2420: spaces needed around "error"

dlabrecq:
@ashumilova Regarding the save filter feature, I was envisioning that the trash can would not be there, only the confirmation slide-in. Although, that was just an idea and not something we need to implement now. I'll pass that along to UXD and see if that's something we want to implement later.

@joshuawilson There is still an issue related to the codebaseDeleted event calling updateCodebases, which in turn always calls startIdleChe() regardless if the API call fails or not. At this point, it's too late to start Che as updateCodebases won't be called again. We should test if Che is started before calling updateCodebases.

ashumilova:
@dlabrecq your suggestion about confirmation is neat.
About updateCodebases - after codebases list is refetched with new data - i call [fetchCheState](https://github.com/fabric8-ui/fabric8-ui/pull/2236/files#diff-79434ea6c0d7476cee3cbe5c09eb2a10R295) as far as Che state is not stored anywhere and need to pass it's current value to the all codebases components.

src/app/space/create/codebases/codebases-delete/codebase-delete-dialog.component.html
```diff
@@ -0,0 +1,6 @@
+<span>You are about to remove a codebase {{codebase.name}} from your space.
```
mindreeper2420: Remove the "a" from "remove a codebase" and bold the {{codebase.name}}.

mindreeper2420:
@ashumilova just the small change to the wording and styling the codebase name, and then it is good to go from my viewpoint :)

ashumilova:
@mindreeper2420 yes, sure - updated, thank you.

fabric8cd:
@ashumilova fabric8/fabric8-ui:SNAPSHOT-PR-2236-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2236-fabric8-ui.badger.fabric8.io

src/app/space/create/codebases/codebases.component.ts
```diff
@@ -267,7 +286,13 @@ export class CodebasesComponent implements OnDestroy, OnInit {
           this.codebases = cloneDeep(codebases);
           this.codebases.unshift({} as Codebase); // Add empty object for row header
           this.applyFilters(this.appliedFilters);
+        } else {
+          // clear the codebases list:
+          this.allCodebases = [];
+          this.codebases = [];
         }
+        // re-fetch Che state to have consistent information inside all related components:
+        this.fetchCheState();
```
dlabrecq: I don't believe we want to call fetchCheState() here? The intention was that we broadcast the cheStateChange event only if the Che state changed from idle to running. 

The cheStateChange event now appears to be used to update "related components"? Can you please clarify what components need to be updated?

Perhaps a different event name would be more appropriate, like codebasesUpdate? Then the related components can subscribe to that if necessary.

ashumilova: Place number 1: codebases actions: 
https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts#L20
https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts#L39

Place number 2: codebase item:
https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/space/create/codebases/codebases-item/codebases-item.component.ts#L21
https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/space/create/codebases/codebases-item/codebases-item.component.ts#L47

As long as the codebases list is re-rendered each time on codebases updation - what will happen to those components? they will all be recreated and will have 
`cheRunning: boolean = false;`
and it's not the current actual true state of che. How should they be informed? Where those components have to take that value? Fetch the state on their own?

If user goes to openshift.io and adds a codebase he will not be able to use kebab menu or see the list of workspaces any more. Here is the screenshot:
screenshot from 2017-11-22 09-44-42

About the "intention was that we broadcast the cheStateChange event only if the Che state changed from idle to running" - the name of the event in such case is not expressive/explicit, by Che state change it could change from idle to running, as well as from running to stopped. I think cheRunning - would avoid such misunderstandings.

About sending "codebasesUpdate" - no need for now cause need the che state. So how it could be solved: 
- maybe place in CheService - the current state of Che - for easy access without additional request to server?

dlabrecq: I understand why the codebase items are left in a disabled state and realize that we need to fix that. However, the point I'm trying to convey is that the Che state has not changed, it's still running, so broadcasting cheStateChange in this case is misleading.

At least, the name of the cheStateChange event isn't appropriate because we're not using it to reflect a change in state. Instead, we're now using it to force the components to redraw even though Che was already running.

I'll defer to @joshuawilson if we want to merge as is? It works, but could use some refactoring.

joshuawilson: I agree with @dlabrecq we need to keep the use of the events separate.

bdellasc:
Hi all: 

I just read through this thread and I'd have concerns with using the same pattern as the filter delete because it sounds like deleting this codebase would be a lot more destructive and affect other areas of the system beyond just the codebase. When the user is opting to delete a codebase, showing the other parts of the system that will also be affected by the deletion is very useful information for the user to have before committing it. I think the confirmation dialog is more appropriate here.

@catrobson @joshuawilson @ashumilova @mindreeper2420 @dlabrecq

src/app/space/create/codebases/codebases.component.ts
ashumilova: @dlabrecq will be ok if i pass cheRunning property to components:
`[cheRunning]="cheRunning"`
for example:
`<codebases-item-actions [codebase]="item" [index]="index" [cheRunning]="cheRunning"></codebases-item-actions>`

or you prefer other event to be used?

dlabrecq: I think that would be ok. Whatever is cleaner?

joshuawilson:
@ashumilova what is the status on this?

joshuawilson:
@dlabrecq do you still have an issue with this PR?

ashumilova:
I have to provide the fix of feedbacks - will do

ashumilova:
done, updated the PR

fabric8cd:
@ashumilova fabric8/fabric8-ui:SNAPSHOT-PR-2236-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2236-fabric8-ui.badger.fabric8.io

src/app/space/create/codebases/codebases.component.ts
```diff
@@ -228,31 +230,20 @@ export class CodebasesComponent implements OnDestroy, OnInit {
...
-
   /**
    * Start the Che server
    */
   private startChe(): void {
     // Get state for Che server
     this.subscriptions.push(this.cheService.start()
       .subscribe(che => {
+        this.cheRunning = che && che.running;
         this.broadcaster.broadcast('cheStateChange', che);
```
dlabrecq: Please remove the cheStateChange event broadcast

src/app/space/create/codebases/codebases.component.ts
```diff
@@ -228,31 +230,20 @@ export class CodebasesComponent implements OnDestroy, OnInit {
...
         if (che == undefined || che.running !== true) {
           this.cheStatePoll();
         }
       }, error => {
+        this.cheRunning = false;
         this.broadcaster.broadcast('cheStateChange');
```
dlabrecq: Same -- please remove the cheStateChange event broadcast

src/app/space/create/codebases/codebases.component.ts
```diff
@@ -265,11 +256,13 @@ export class CodebasesComponent implements OnDestroy, OnInit {
     this.subscriptions.push(this.cheService.getState()
       .subscribe(che => {
         if (che !== undefined && che.running === true) {
+          this.cheRunning = true;
           this.broadcaster.broadcast('cheStateChange', che);
```
dlabrecq: Same

src/app/space/create/codebases/codebases.component.ts
```diff
@@ -265,11 +256,13 @@ export class CodebasesComponent implements OnDestroy, OnInit {
...
         } else {
           this.startChe();
         }
       }, error => {
+        this.cheRunning = false;
         this.broadcaster.broadcast('cheStateChange');
```
dlabrecq: Same

src/app/space/create/codebases/codebases-item/codebases-item.component.ts
```diff
@@ -14,11 +14,11 @@ import { NotificationType as NotificationTypes } from 'patternfly-ng';
   styleUrls: ['./codebases-item.component.less']
 })
 export class CodebasesItemComponent implements OnDestroy, OnInit {
+  @Input() cheRunning: boolean;
```
dlabrecq: Where are we using cheRunning in this class? Please replace the broadcast subscription (cheStateChange') in ngOnInit. The cheRunning variable should be enough?

fabric8cd:
@ashumilova fabric8/fabric8-ui:SNAPSHOT-PR-2236-6 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2236-fabric8-ui.badger.fabric8.io

ashumilova:
Removed broadcast and also had to rework a little this part https://github.com/fabric8-ui/fabric8-ui/pull/2236/files#diff-66483b2a9297e91d74e72a521c57ace9R55.

src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
```diff
@@ -1,14 +1,13 @@
...
 import { CodebasesService } from '../services/codebases.service';
 import { Broadcaster, Notification, NotificationType, Notifications } from 'ngx-base';
 import { Dialog } from 'ngx-widgets';
 import { WindowService } from '../services/window.service';
 import { WorkspacesService } from '../services/workspaces.service';
-import {IModalHost} from "../../../wizard/models/modal-host";
+import { IModalHost } from '../../../wizard/models/modal-host';
 
 @Component({
   encapsulation: ViewEncapsulation.None,
```
dlabrecq: There is an unused cheRunning input in this class that should be removed.

https://github.com/ashumilova/fabric8-ui/blob/4a59ebc3452e43902418fc118a462a4c98e39a4f/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts#L19

src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
```diff
@@ -17,11 +17,11 @@ import {IModalHost} from "../../../wizard/models/modal-host";
   styleUrls: ['./codebases-item-actions.component.less']
 })
 export class CodebasesItemActionsComponent implements OnDestroy, OnInit {
+  @Input() cheRunning: boolean;
```
dlabrecq: This appears to be unused

joshuawilson:
@ashumilova i'm going to merge this, you can make the change that @dlabrecq suggests in a new PR.

src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
ashumilova: used in template https://github.com/ashumilova/fabric8-ui/blob/4a59ebc3452e43902418fc118a462a4c98e39a4f/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.html#L7

src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.ts
ashumilova: used in template https://github.com/ashumilova/fabric8-ui/blob/4a59ebc3452e43902418fc118a462a4c98e39a4f/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.html#L7

dlabrecq: Ah... I was expecting each component would use a similar Che input and the template would reference che.running.

PR #2238, opened 2017-11-16
refactor(deployments): rename app-card to deployment-card

andrewazores:
related to openshiftio/openshift.io#1382

This does also move the component's files into a subdirectory, since there will be several more components split out and moved here in the near future.


fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2238-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2238-fabric8-ui.badger.fabric8.io

joshuawilson:
Due to the way git tracks files, it is best to keep deletions or renaming in a separate commit. Please refactor this into 2 commits. One that changes all the file names that you want to change and another that changes content.

andrewazores:
Okay. Does this mean that the renaming commit would also leave relative import paths broken, and the next commit would fix those as well as changing class names etc.?

joshuawilson:
You can change the paths with the rename, just don't change anything else. Keep the commits atomic.

PR #2239, opened 2017-11-16
refactor(deployments): Use component properties instead of static values

c-koehler:
fixes https://github.com/openshiftio/openshift.io/issues/1373

PR #2240, opened 2017-11-16
refactor(deployments): extract Resource Usage panel into component

andrewazores:
fixes openshiftio/openshift.io#1385


andrewazores:
Ah, this conflicts with https://github.com/fabric8-ui/fabric8-ui/pull/2239 . That one is more involved, so I will wait and rebase the work from this PR until after that one.

PR #2247, opened 2017-11-20
fix(deployments): resource-card unsubscribes on destroy

andrewazores:
fixes openshiftio/openshift.io#1412

### DIFF
--- a/src/app/space/create/apps/components/resource-card.component.ts
+++ b/src/app/space/create/apps/components/resource-card.component.ts
@@ -9,7 +9,10 @@ import { DeploymentsService } from '../services/deployments.service';
 import { Environment } from '../models/environment';
 import { Stat } from '../models/stat';
 
-import { Observable } from 'rxjs';
+import {
+  Observable,
+  Subscription
+} from 'rxjs';
 
 @Component({
   selector: 'resource-card',
@@ -25,14 +28,21 @@ export class ResourceCardComponent implements OnInit {
   usedPercent: number;
   unusedPercent: number;
 
+  private statSubscription: Subscription;
+
   constructor() { }
 
   ngOnInit(): void {
-    this.stat.subscribe(val => {
+    this.statSubscription = this.stat.subscribe(val => {
       this.used = val.used;
       this.total = val.total;
       this.usedPercent = (this.total !== 0) ? Math.floor(this.used / this.total * 100) : 0;
       this.unusedPercent = 100 - this.usedPercent;
     });
   }
+
+  ngOnDestroy(): void {
+    this.statSubscription.unsubscribe();
+  }
+
 }


fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2247-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2247-fabric8-ui.badger.fabric8.io

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2247-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2247-fabric8-ui.badger.fabric8.io

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2247-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2247-fabric8-ui.badger.fabric8.io

src/app/space/create/apps/components/resource-card.component.ts
```diff
@@ -25,14 +28,24 @@ export class ResourceCardComponent implements OnInit {
   usedPercent: number;
   unusedPercent: number;
 
+  private readonly unsubscribe: Subject<void> = new Subject<void>();
```
joshuawilson: Why are you creating a variable called `unsubscribe`? RxJS has a function on `Observable` called `unsubscribe()`.

What are you using `Subject` for? This does not look the correct use of this.

andrewazores: It's taken from the example here:

http://brianflove.com/2016/12/11/anguar-2-unsubscribe-observables/

It looked like a nice solution for the possibility of having to manage unsubscribing multiple Observables at a time, although it's definitely overkill here where we only have one.

joshuawilson: You are right, unsubscribe is a part of subscription. 

I would rather you figure out the right way to use Observables then use an example that is not doing what you really need. You are not in a rush.

andrewazores: The ways I've discovered to manage Observables are:

1) use the `async` pipe in the template, which documentation says will have Angular properly unsubscribe when the component is destroyed along with avoiding having to write TS to assign values to properties etc.

2) keep reference to each `Subscription` after subscribing, and call `unsubscribe` on each on destroy. Possibly by keeping the subscriptions in a collection and `forEach`ing it.

3) use an operator like `first` so that the Subscription is automatically closed as it's "finite"

4) use an operator like `takeWhile` or `takeUntil` to use some other event stream (ex. component destruction) to signal when the primary event stream is no longer needed (the original approach in this PR)

Are there any others that I've missed? I've listed them in the order that I believe they're preferred, judging by other examples in the codebase and this conversation.

joshuawilson: I can't think of any but I'm not an RxJS expert.

joshuawilson:
@dgutride can you please review this, you have more experience with RxJS then I do at this point.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2247-6 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2247-fabric8-ui.badger.fabric8.io

joshuawilson:
[test]

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2247-8 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2247-fabric8-ui.badger.fabric8.io

PR #2249, opened 2017-11-20
enhancement(deployments): add patternfly charts to deployment cards

chrislessard:
Our deployment card components no longer use a static message "Detailed
information placeholder", but instead display dummy patternfly charts
that will eventually be used to show real data. Polling mock data will be added in
the charts in a follow up patch.

This patch is a first step towards resolving https://github.com/openshiftio/openshift.io/issues/1366

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -57,7 +57,7 @@ import {
   UserService
 }                             from 'ngx-login-client';
 import { WidgetsModule }      from 'ngx-widgets';
-import { ActionModule, ListModule, TreeListComponent, PatternFlyNgModule } from 'patternfly-ng';
+import { ActionModule, ListModule, TreeListComponent, PatternFlyNgModule, ChartModule } from 'patternfly-ng';
 
 /*
  * Platform and Environment providers/directives/pipes
@@ -145,6 +145,7 @@ export type StoreType = {
     BrowserAnimationsModule,
     BrowserModule,
     BsDropdownModule.forRoot(),
+    ChartModule,
     FormsModule,
     HttpModule,
     KubernetesRestangularModule,
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -57,7 +57,12 @@ import {
   UserService
 }                             from 'ngx-login-client';
 import { WidgetsModule }      from 'ngx-widgets';
-import { ActionModule, ListModule, TreeListComponent, PatternFlyNgModule, ChartModule } from 'patternfly-ng';
+import {
+  ActionModule,
+  ChartModule,
+  ListModule,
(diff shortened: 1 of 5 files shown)


src/app/space/create/apps/apps.module.ts
```diff
@@ -24,7 +26,8 @@ const imports = USE_RUNTIME_CONSOLE ?
     CollapseModule.forRoot(),
     AccordionModule.forRoot(),
     CommonModule,
-    AppsRoutingModule
+    AppsRoutingModule,
+    ChartModule
```
joshuawilson: Not sure if this holds true on this sub-component but at the parent level the AppsRoutingModule must appear last.

src/app/space/create/apps/components/deployment-card/deployment-card.component.spec.ts
```diff
@@ -51,6 +51,14 @@ describe('DeploymentCardComponent', () => {
     component.environment = { environmentId: 'mockEnvironmentId', name: 'mockEnvironment'};
 
     fixture.detectChanges();
+
+    it('should generate a unique chartid for each DeploymentCardComponent instance', () => {
+      let depCard1 = new DeploymentCardComponent(null);
+      let depCard2 = new DeploymentCardComponent(null);
+      let depCard3 = new DeploymentCardComponent(null);
+
+      expect(depCard1.config.chartId !== depCard2.config.chartId !== depCard3.config.chartId);
```
joshuawilson: I don't think you can do a triple equals and have it evaluate correctly. There are also better ways of checking this with functions on `expect` like `toEqual()`.

jiekang:
Some unit tests fail with these changes:

```
✖ 2 tests failed

FAILED TESTS:
  DeploymentCardComponent
    podCountLabel
      ✖ should be set from mockSvc.getPodCount result
        PhantomJS 2.1.1 (Linux 0.0.0)
      Error: Template parse errors:
      Can't bind to 'config' since it isn't a known property of 'pfng-chart-sparkline'.
```

Looks like the test now needs patternfly chart visibility to work.

src/app/app.module.ts
```diff
@@ -57,7 +57,7 @@ import {
   UserService
 }                             from 'ngx-login-client';
 import { WidgetsModule }      from 'ngx-widgets';
-import { ActionModule, ListModule, TreeListComponent, PatternFlyNgModule } from 'patternfly-ng';
+import { ActionModule, ListModule, TreeListComponent, PatternFlyNgModule, ChartModule } from 'patternfly-ng';
```
joshuawilson: Please list these alphabetically. Also after 2 or 3 just make it a vertical list. It should have been updated a while ago.

PR #2270, opened 2017-11-28
fix(link): support for linking github using the auth service

sbose78:
At the moment,
We will support 'connecting'  `github` the new way and support connecting `openShift-v3` the old way.

( Because we haven't switched openshift-v3 to the new 'Get token' endpoint  - which is because of the OSO proxy work that is in progress. )

There are 3 **scenarios**

1. User chooses to connect **only** github , and so the UI makes a GET call to the new API to connect github account. The API returns the redirect url for logging in to Github in the response ( as a 200OK)

2. User chooses to connect **only** openshift-v3 
The UI does this the old/existing way by redirection.

3. User chooses both github and openshift-v3
The UI does (1) , and in the redirect param , the url for (2) is passed.

**How to test**

For the above 3 scenarios you will need to -

- Login to https://prod-preview.openshift.io and grab you access token.

- Unlink `github` 
```
curl -i -H "Authorization: Bearer $ACCESS_TOKEN" -X DELETE https://auth.prod-preview.openshift.io/api/token?for=https://github.com
```

- Unlink `openshift-v3`

```
curl -i -H "Authorization: Bearer $AUTH_TOKEN" -X DELETE https://auth.prod-preview.openshift.io/api/token?for=https://api.free-int.openshift.com
```


sbose78:
replaced by https://github.com/fabric8-ui/fabric8-ui/pull/2193

fabric8cd:
@sbose78 fabric8/fabric8-ui:SNAPSHOT-PR-2270-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2270-fabric8-ui.badger.fabric8.io

src/app/getting-started/services/provider.service.ts
```diff
@@ -31,8 +36,42 @@ export class ProviderService {
...
+  }
+
+   /**
+   * Link a GitHub and OpenShift Online account to the user account
+   *
+   * @param redirect URL to be redirected to after successful account linking
+   */
+  linkGitHubAndOpenShift(redirect: string): void {
+
+    let parsedToken = jwt_decode(this.auth.getToken());
```
dgutride: this variable isn't used anywhere in this function

src/app/getting-started/services/provider.service.ts
```diff
@@ -31,8 +36,42 @@ export class ProviderService {
...
    * @param redirect URL to be redirected to after successful account linking
    */
-  linkGitHub(redirect: string): void {
-    this.link("github", redirect);
+  linkGitHub(redirect: string): void{
+    let headers = new Headers();
+    let githubLinkURL = this.linkUrl + `redirect=` + redirect ;
+
+    this.http
+    .get(this.linkUrl, { headers: headers })
```
dgutride: What value do these headers provide here - nothing is set in them.

src/app/getting-started/services/provider.service.ts
```diff
@@ -31,8 +36,42 @@ export class ProviderService {
    *
    * @param redirect URL to be redirected to after successful account linking
    */
-  linkGitHub(redirect: string): void {
-    this.link("github", redirect);
+  linkGitHub(redirect: string): void{
+    let headers = new Headers();
+    let githubLinkURL = this.linkUrl + `redirect=` + redirect ;
```
dgutride: There is no reason to use backticks here - single quotes are sufficient if you aren't going to use string interpolation.

src/app/getting-started/services/provider.service.ts
```diff
@@ -31,8 +36,42 @@ export class ProviderService {
...
+   * Link a GitHub and OpenShift Online account to the user account
+   *
+   * @param redirect URL to be redirected to after successful account linking
+   */
+  linkGitHubAndOpenShift(redirect: string): void {
+
+    let parsedToken = jwt_decode(this.auth.getToken());
+    // the new url is /api/link/session 
+    let opensShiftLinkingRedirectUrl = this.getLegacyLinkingUrl("openshift-v3",redirect);
+    this.linkGitHub(opensShiftLinkingRedirectUrl);
```
dgutride: This doesn't really make sense to me - you are asking for the openshift url and then calling linkGitHub.  What should the logic be?

src/app/getting-started/services/provider.service.ts
```diff
@@ -31,8 +36,42 @@ export class ProviderService {
...
+  }
+
+  getLegacyLinkingUrl(provider: string, redirect: string): string{
+    let parsedToken = jwt_decode(this.auth.getToken());    
+    let url = this.loginUrl + `/session?`
+    + "clientSession=" + parsedToken.client_session
+    + "&sessionState=" + parsedToken.session_state
+    + "&redirect=" + redirect // brings us back to Getting Started.
+    + "&provider="+provider;
+    return url
```
dgutride: There are a number of linting issues with this function - please end line 74 with a semicolon at a minimum.  It is good to also space consistently (see the provider assignment on line 73 and the function call only 63).  All of those things help lead to code that is readable and will not break our linting rules once Joshua turns them on globally.

sbose78:
Thanks @dlabrecq and @dgutride  :) 
This was a Work-in-progress PR ( and it's only 50% done ) and you caught it in the middle of a mess :D 

Let me add a description, and also address the design decisions.

Btw, thanks for pointing out the lint-related stuff - I wouldn't have caught them easily.

src/app/getting-started/services/provider.service.ts
sbose78: When we link Github + OpenShift together:
We call the new linking API for github and pass the old linking api url for OpenShift in the redirect.

Since we are yet to move over openshift-v3 while we wait for OSO proxy to be in place.

sbose78:
Hi @dgutride @dlabrecq  , Could you please check at the current progress and my comments - and let me know if they make sense?

fabric8cd:
@sbose78 fabric8/fabric8-ui:SNAPSHOT-PR-2270-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2270-fabric8-ui.badger.fabric8.io

alexeykazakov:
> This looks good - why is it still a WIP?

Because of missing tests I believe.

fabric8cd:
@sbose78 fabric8/fabric8-ui:SNAPSHOT-PR-2270-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2270-fabric8-ui.badger.fabric8.io

sbose78:
Hi @dgutride  , @dlabrecq  - 
I've added a test to the only method among my changes which doesn't end up in a remote call.
also, I've updated the descriptions with a 'how to test' section.

src/app/getting-started/services/provider.service.ts
```diff
@@ -23,16 +28,43 @@ export class ProviderService {
    * @param redirect URL to be redirected to after successful account linking
    */
   linkAll(redirect: string): void {
-    this.link(null, redirect);
+        let openShiftLinkingRedirectUrl = this.getLegacyLinkingUrl("openshift-v3",redirect);
+        // after linking github, proceed with linking openshift-v3,
+        // hence passing openshift linking url as a redirect.
+        this.linkGitHub(openShiftLinkingRedirectUrl);
```
dlabrecq: It appears that linkAll no longer links OpenShift accounts as originally intended. And as described by the method's comments above. FYI, a null value defaulted to linking both.

If this is no longer linking both OpenShift and GitHub, perhaps it should be removed or renamed? I believe there is still logic in the getting started component that links OS, GitHub, or both at the same time.

src/app/getting-started/services/provider.service.ts
```diff
@@ -23,16 +28,43 @@ export class ProviderService {
...
+  
+
+  getLegacyLinkingUrl(provider: string, redirect: string): string{
+    let parsedToken = jwt_decode(this.auth.getToken());    
+    let url = this.loginUrl + "/session?"
+    + "clientSession=" + parsedToken.client_session
+    + "&sessionState=" + parsedToken.session_state
+    + "&redirect=" + redirect // brings us back to Getting Started.
+    + "&provider=" + provider;
+    return url;
```
dlabrecq: 1. I'm still curious about the getLegacyLinkingUrl() name. Why is it a legacy URL?
2. It appears that the getLegacyLinkingUrl() and link() methods are nearly identical. Can the link method get its URL from getLegacyLinkingUrl() instead?
3. Are we still using the link() method (e.g., to link openshift)?

src/app/getting-started/services/provider.service.ts
alexeykazakov: It does link OpenShift by passing openShiftLinkingRedirectUrl as a redirect to the GitHub API.
We have to use both old and new API to link all accounts. OSO is using the old one and GitHub the new one. So we if we need to link both OSO and GH we call the new API to link GH but we use the old API link as a redirect.

New API links GH and then redirects to the old API to link OSO. This is why we in this method there is only one call. The rest will be done by backend and both OSO and GH will be linked.

src/app/getting-started/services/provider.service.ts
alexeykazakov: link() ? There is no link() function method anymore. Could you elaborate please?
There are linkAll() links both OSO and GitHub
linkGitHub() uses new API to link GH
getLegacyLinkingUrl() to get a link to the legacy/depricated API which we will remove later.

They all are different.

fabric8cd:
@sbose78 fabric8/fabric8-ui:SNAPSHOT-PR-2270-7 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2270-fabric8-ui.badger.fabric8.io

src/app/getting-started/services/provider.service.ts
dlabrecq: I was referring to the link() method in provider.service here: https://github.com/sbose78/fabric8-ui/blob/f84fad16eab2e2f0969163012e0873fa169af709/src/app/getting-started/services/provider.service.ts#L85

The code (below) to generate the url appears duplicated in getLegacyLinkingUrl() and link().

```
let parsedToken = jwt_decode(this.auth.getToken());    
    let url = this.loginUrl + "/session?"
    + "clientSession=" + parsedToken.client_session
    + "&sessionState=" + parsedToken.session_state
    + "&redirect=" + redirect // brings us back to Getting Started.
+ "&provider=" + provider;
```

alexeykazakov: Oh, right. I missed the old link() :)
Yes, we should reuse getLegacyLinkUrl() in linkOpenShift() and remove link(). Good catch.

sbose78: Since this service will again undergo refactoring ( and there aren't too many tests ), I was avoiding touching the existing code :) Let me do it nevertheless. Thanks.

alexeykazakov:
@dlabrecq, @dgutride thanks for review!

src/app/getting-started/services/provider.service.ts
sbose78: Done.

fabric8cd:
@sbose78 fabric8/fabric8-ui:SNAPSHOT-PR-2270-9 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2270-fabric8-ui.badger.fabric8.io

PR #2284, opened 2017-12-01
fix(scrollbar): fix empty scrollbar in cards

mindreeper2420:
Fixed an issue where scrollbars always appeared inside of card components.


mindreeper2420:
@joshuawilson can this be merged before https://github.com/openshiftio/openshift.io/issues/1617 is tackled?

Issue #1372, opened 2017-05-22
Update Cache Settings to Match OSIO Marketing


mindreeper2420:
Update OSIO product to utilize caching changes that are utilized on the marketing page.

OpenShiftio/OpenShift.io PR: https://github.com/openshiftio/openshift.io/pull/176

mindreeper2420:
PR here: https://github.com/fabric8io/fabric8-ui/pull/1373

mindreeper2420:
Closing, as PR was merged.

PR #2293, opened 2017-12-04
refactor(deployments): split elements into separate directives/components

c-koehler:
fixes https://github.com/openshiftio/openshift.io/issues/1372

Also related to https://github.com/openshiftio/openshift.io/issues/1385


fabric8cd:
@c-koehler fabric8/fabric8-ui:SNAPSHOT-PR-2293-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2293-fabric8-ui.badger.fabric8.io

c-koehler:
Updated by moving them into better named folders.

I still can't think of a good name, I feel the same way you do about `appenvs` but don't have anything immediately popping into my head for a replacement name.

andrewazores:
Thanks.

Looking back at the comment I made on issue #1372, I don't know if we want to go the way I just asked for or what I said then (https://github.com/openshiftio/openshift.io/issues/1372#issuecomment-345861165). The old comment mirrors how it's done in the codebases area though, so maybe that's the way forward? I don't really know - I think it's going to depend on how much interdependency there ends up being between components. The less there is the better the old comment seems. I suppose we do want to aim for minimal interdependency.

Either way, a follow-up commit in this series to move the current contents of deployments/components/* into less-nested component directories makes sense.

jiekang:
Just AppsComponent or ApplicationsComponent would work for me. It's showing information for each application. The [0,N] environment(s) the application is deployed in is under this umbrella.

jiekang:
@c-koehler also while we're here, can you change `deployments.component.spec.ts`

```
describe('AppsComponent', () => {
```

to 

```
describe('DeploymentsComponent', () => {
```

c-koehler:
Updated name, also reworded first few commits to start with a lower case letter for consistency

andrewazores:
I guess you're right @jiekang, it just sounded too generic, but at least it's more sensible than "appenvs." @c-koehler please rename appenvs -> apps (all instances, including the directory), otherwise this looks pretty good to me. We still might want to pursue the structure I left in that comment on #1372, but if so, I think that can be done as a follow-up PR. At least here the components are all split out the right way, so reorganizing the directories later should be a small body of work.

PR #2313, opened 2017-12-12
feat(test): add createMock test util

andrewazores:
closes https://github.com/openshiftio/openshift.io/issues/1644

I have a use-case in mind here already (DeploymentsComponent spec), but I'm unsure of how best to demo that. Should I include a follow-up commit on this branch/PR to do that small refactoring?


jiekang:
Thanks, looks nice to me.

PR #2317, opened 2017-12-14
fix(test): force a build to generate a new PR for testing

ldimaggi:


maxandersen:
len - do you still need this as there already been other merges done on the master ?

maxandersen:
never mind - seems you just hadn't rebased. I've rebased and pushed.

mental note: we really should have a way to trigger a rebuild without doing a push.

joshuawilson:
we do have a way, it is on Jenkins

PR #2350, opened 2017-12-28
fix(env-vars): Use realm value from config instead of environment

jarifibrahim:
fixes https://github.com/openshiftio/openshift.io/issues/1734

### DIFF
--- a/src/app/shared/api-locator.service.ts
+++ b/src/app/shared/api-locator.service.ts
@@ -37,7 +37,7 @@ export class BaseApiLocatorService {
   }
 
   get realm(): string {
-    return this.envVars.get('realm') || 'fabric8';
+    return this.config.realm || 'fabric8';
   }
 
   get branding(): string {
--- a/src/app/shared/config/fabric8-ui-config.ts
+++ b/src/app/shared/config/fabric8-ui-config.ts
@@ -9,6 +9,7 @@ export class Fabric8UIConfig {
   ssoApiUrl: string;
   statusApiUrl: string;
   witApiUrl: string;
+  realm: string;
   tenantApiUrl: string;
   kubernetesMode: string;
   authApiUrl: string;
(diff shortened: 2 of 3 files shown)


fabric8cd:
@jarifibrahim fabric8/fabric8-ui:SNAPSHOT-PR-2350-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2350-fabric8-ui.badger.fabric8.io

joshuawilson:
@sbose78 @alexeykazakov @xcoulon how are we using the REALM variable? Do you agree with this change?

sbose78:
@jarifibrahim what is the difference between reading the env var in the method, and reading the env var in the config?

Change looks fine to me.

sbose78:
@joshuawilson the REALM variable is used to construct the openshift token url.

alexeykazakov:
For clarification. We won't need any realm settings at all when we switch to Multicluster. But for now, yes, we use it to construct the OSO token url as @sbose78 said.

jarifibrahim:
@sbose78 
Here's how it works
1. The `envVars` map is set at **build time**. When you do a `npm run build:prod`, the `envVar` is populated the current environment variable values. 
2. The `config` map is set at **runtime**. The `config` is populated at runtime by the docker. When you do a `docker run`,  the entry point (https://github.com/kbsingh/openshift-nginx/blob/master/Dockerfile) adds the appropriate environment variables to the config file.

 More info - https://github.com/fabric8-ui/fabric8-planner/issues/2368#issuecomment-353884487

src/app/shared/api-locator.service.ts
```diff
@@ -36,7 +36,7 @@ export class BaseApiLocatorService {
   }
 
   get realm(): string {
-    return this.envVars.get('realm') || "fabric8";
```
jarifibrahim: I think we should rename `envVars` to `buildTimeEnvVars`. That would make more sense since `envVars` doesn't contain actual environment variable values.

joshuawilson:
please rebase

joshuawilson:
please remove the space in `env vars` in the commit. The subject can not have any spaces.

fabric8cd:
@jarifibrahim fabric8/fabric8-ui:SNAPSHOT-PR-2350-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2350-fabric8-ui.badger.fabric8.io

joshuawilson:
This appears out of date now that we have multi-cluster. If this is still needed, please reopen and rebase it.

PR #2357, opened 2018-01-03
feat(deployments): hide undeployed app cards

andrewazores:
hide app cards when app is not deployed in a given environment

fixes https://github.com/openshiftio/openshift.io/issues/1749

### DIFF
--- a/src/app/space/create/deployments/apps/deployment-card.component.html
+++ b/src/app/space/create/deployments/apps/deployment-card.component.html
@@ -1,43 +1,45 @@
-<div class="card-pf" (click)="collapsed = !collapsed">
-  <div class="card-pf-body">
-    <div class="row mini-card-content">
-      <div class="col-sm-2">
-        <span class="pficon pficon-ok fa-lg" tooltip="Everything looks okay" placement="top"></span>
-      </div>
-      <div class="col-sm-7">
-        <deployments-donut [applicationId]="applicationId" [environment]="environment" [spaceId]="spaceId" [mini]="true"></deployments-donut>
-      </div>
-      <div class="col-sm-1">
-        <span id="versionLabel">{{ version | async }}</span>
-      </div>
-      <div class="col-sm-2">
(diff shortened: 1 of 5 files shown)


jiekang:
These code changes are a bit confusing for me. I'd expect that when using the API to get a list of deployed applications, it would only return us the list of application deployments which are obviously, actually deployed. From there the `*ngFor` over the data would naturally only show applications which are deployed and not show applications that aren't deployed.

I think the current placeholder API exhibits this issue solely because it returns a list of environments and a list of applications separately, while the 'true' API wouldn't (shouldn't?) be done in this manner.

I'd expect some request to provide, for example (extremely loosely):

```
{
vertx-hello : {
  deployments: {
    stage : {
    // deployment info for stage
    },
    run : {
    }
  }
}

vertx-blob : {
  deployments: {
  // empty
  }

  vertx-blah: {
    stage: {

    }
  }
}
```

Using something like this, we should naturally be able to show the cards we have deployment info for, and not show cards that don't have any deployment info. Thoughts?

andrewazores:
Sure, that would work too and be nicer. This design is based off of what happens with the current unmerged WIT backend work however, and with the service mostly mirroring the format (but pared down) returned there. If we want the (UI) service to return information about applications as well as the environments they are deployed in in one call then we'll need to redo some UI logic - this approach was simply less invasive. It's also very easy to adapt to the format currently returned by the WIT backend.

I think the idea of simply excluding applications from the service response if they do have any deployments will not be as cut-and-dry as the ngFor simply not showing them for us. What if the environments are ordered "Stage, Run", and some application is only deployed in "Run"? I imagine the service would return a singleton list response, and the ngFor would place that as the first element - visually, under "Stage". So as I said, this would take more work on the UI and layout code as well.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2357-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2357-fabric8-ui.badger.fabric8.io

PR #2373, opened 2018-01-09
refactor(deployments): Move iconStatus logic to parent component.

chrislessard:
This patch moves the logic for determing a deployment card's icon class to the parent component, deployment-card.

Solves https://github.com/openshiftio/openshift.io/issues/1659


lint/src/app/space/create/deployments/apps/deployment-card.component.ts.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<checkstyle version="5.7">[120, 1]: trailing whitespace
+</checkstyle>
```
andrewazores: What is this file?

src/app/space/create/deployments/apps/deployment-card.component.html
```diff
@@ -3,7 +3,7 @@
     <div class="card-pf-body">
       <div class="row mini-card-content">
         <div class="col-sm-2">
-          <deployment-status-icon [cpuDataStream]='cpuStat'></deployment-status-icon>
+          <deployment-status-icon [iconClass]='iconClass' [toolTip]='toolTip'></deployment-status-icon>
```
andrewazores: I believe these should be double-quotes, not singles.

src/app/space/create/deployments/apps/deployment-card.component.spec.ts
```diff
@@ -33,6 +33,11 @@ import { Environment } from '../models/environment';
 import { DeploymentsService } from '../services/deployments.service';
 import { DeploymentCardComponent } from './deployment-card.component';
 
+const ICON_OK = 'pficon-ok';
+const ICON_WARN = 'pficon-warning-triangle-o';
+const ICON_ERR = 'pficon-error-circle-o';
+const MSG_OK = 'Everything is ok.';
```
andrewazores: Should these be duplicates or can you just export the values from the component and reference them here?

src/app/space/create/deployments/apps/deployment-card.component.ts
```diff
@@ -13,6 +13,12 @@ import { CpuStat } from '../models/cpu-stat';
 import { Environment } from '../models/environment';
 import { DeploymentsService } from '../services/deployments.service';
 
+const ICON_OK = 'pficon-ok';
+const ICON_WARN = 'pficon-warning-triangle-o';
+const ICON_ERR = 'pficon-error-circle-o';
+const MSG_OK = 'Everything is ok.';
+const STAT_THRESHOLD = .6;
+
```
andrewazores: Maybe these can be made into StatusIcon and StatusMessage enums or something similar, so that the things are grouped together?

src/app/space/create/deployments/apps/deployment-status-icon.component.ts
```diff
@@ -23,38 +23,13 @@ const STAT_THRESHOLD = .6;
...
-      this.iconClass = ICON_WARN;
-      this.toolTip = 'CPU usage is approaching or at capacity.';
-    }
-
-    if (stat.used > stat.quota) {
-      this.iconClass = ICON_ERR;
-      this.toolTip = 'CPU usage has exceeded capacity.';
-    }
-  }
+  ngOnInit(): void { }
```
andrewazores: May as well remove the constructor, ngOnDestroy, and ngOnInit if they're just going to be empty for this very simple component. If they're ever needed in the future then they can be re-added then.

src/app/space/create/deployments/apps/deployment-card.component.ts
```diff
@@ -13,6 +13,12 @@ import { CpuStat } from '../models/cpu-stat';
 import { Environment } from '../models/environment';
 import { DeploymentsService } from '../services/deployments.service';
 
+const ICON_OK = 'pficon-ok';
+const ICON_WARN = 'pficon-warning-triangle-o';
+const ICON_ERR = 'pficon-error-circle-o';
```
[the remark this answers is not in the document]
jiekang: I think Andrew touched upon these too.

src/app/space/create/deployments/apps/deployment-card.component.ts
```diff
@@ -13,6 +13,20 @@ import { CpuStat } from '../models/cpu-stat';
 import { Environment } from '../models/environment';
 import { DeploymentsService } from '../services/deployments.service';
 
+export const CLASSES = Object.freeze({
+  ICON_OK: 'pficon-ok',
+  ICON_WARN: 'pficon-warning-triangle-o',
+  ICON_ERR: 'pficon-error-circle-o'
+});
```
andrewazores: Why use a frozen object rather than an actual Enum?

https://www.typescriptlang.org/docs/handbook/enums.html

And, rather than exporting this value separately in the file, how about moving it inside of the component class? Then it doesn't require an additional import for consumers of the component, and the intention/purpose of the values is more obvious at a glance. Also, as Jie said, these enums should live in the icon component class, not its parent. The idea is that there could be various, even potentially unknown third party, parents to the icon, but the icon should expect to only be used with some specific set of values that it already knows about.

fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2373-7 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2373-fabric8-ui.badger.fabric8.io

fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2373-9 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2373-fabric8-ui.badger.fabric8.io

PR #2379, opened 2018-01-10
fix(deployments): add missing osio.less import

jiekang:
I believe all less files are supposed to import the osio.less file. This adds it to one that was missing the import.


fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2379-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2379-fabric8-ui.badger.fabric8.io

PR #2388, opened 2018-01-12
Deployments Network linechart

andrewazores:
This PR adds the network utilization linechart to deployment cards. The "Requests" sparkline is left out for a follow-up PR.

The chart component itself is essentially the upstream patternfly-ng sparkline component but modified for a linechart: https://github.com/patternfly/patternfly-ng/blob/master/src/app/chart/sparkline/sparkline.component.ts .

The design for the chart can be seen here: https://redhat.invisionapp.com/share/YSDL0KRKW#/screens/260764293 .

screenshot-20180112120239


andrewazores:
@jiekang the largest difference (other than config values) is that the data set is expected to contain more than one Y-axis value, so the `getChartData` function is modified for this case, and the `tooltip` function is also reimplemented to suit multiple Y-axis values as well.

It's probably doable to extend the Sparkline component rather than ChartBase and then selectively override specific things, but this seemed more fragile than going off of ChartBase directly.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2388-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2388-fabric8-ui.badger.fabric8.io

andrewazores:
Actually - extending the SparklineComponent and overriding wouldn't be as easy, because the defaultConfig property is marked private. So overriding/modifying the chart config would depend on modifying just the input config, which would get hairy when also trying to detect changes to the config from consumers of the component - or duplicating a second defaultConfig and trying to merge them all together.

jiekang:
The input config is manipulated via:

```
defaultsDeep(this.config, this.defaultConfig);
```

After, the data property is manipulated via: 
```
this.config.data = merge(this.config.data, this.getChartData());
```

The tooltip is only used for the default config via:
```
this.defaultConfig.tooltip = this.tooltip();
```

If you input a config object with the tooltip property, it will use that instead of the default.

If you input a config.data property, it gets merged with the results of `this.getChartData()`. I suppose the merge would make it impossible to input chartData you want. I'm not sure about the details of lodash `merge` but this seems strange in itself. Anyways, the hacky way would be to make sure the check:

```
if (this.chartData && this.chartData.dataAvailable !== false && this.chartData.xData && this.chartData.yData) {
```

is false. Then it won't merge and you're free to do whatever you want. Though it's hacky it would reduce the code addition by quite a sizable chunk.

Do you think there's any way to go about this? Otherwise I'll review it as it is.

andrewazores:
I mean... that's shorter in terms of LOC, but IMO would be longer in terms of developer time to understand what exactly is being done and why if somebody else comes back to maintain that in the future.

src/app/space/create/deployments/apps/deployment-details.component.ts
```diff
@@ -49,6 +63,11 @@ export class DeploymentDetailsComponent {
     chartId: uniqueId('mem-chart-') + '-'
```
jiekang: Can we clean these up while we're here? The extra + '-' seems strange.

src/app/space/create/deployments/deployments-linechart/deployments-linechart-data.ts
```diff
@@ -0,0 +1,7 @@
+export abstract class DeploymentsLinechartData {
```
jiekang: Linechart vs LineChart? Not really a big deal though.

src/app/space/create/deployments/deployments-linechart/deployments-linechart.component.ts
```diff
@@ -0,0 +1,178 @@
...
+  ngOnInit(): void {
+    this.setupConfigDefaults();
+    this.setupConfig();
+    this.generateChart(this.config, true);
+  }
+
+  ngDoCheck(): void {
+    if (!isEqual(this.config, this.prevConfig) || !isEqual(this.chartData, this.prevChartData)) {
+      this.setupConfig();
+      this.generateChart(this.config, true);
```
jiekang: This is one aspect of the patternfly charts that I really wish was improved. Using `this.generate(foo, true)` will cause it to run `chart.generate` even though in the situation where only the chart data changes, it would suffice to run `chart.load` with the data object instead of rebuilding the entire chart (which can be done via `this.generate(foo, false)`. There's so much unnecessary work being done here; what with rebuilding the whole config object when most of the time, only the chartData will change.

There is also a possible memory leak described when calling `chart.generate` repeatedly: https://github.com/c3js/c3/issues/926 which doesn't seem to have a clear resolution (does the problem still exist?)

I'd like to also see the use of `ngOnChanges` instead of `ngDoCheck` as the former only gets called if there are changes to any input values, while the latter is called every change detection cycle, even if nothing relevant to this component has changed. I believe Angular documentation says to only use one, not both though (just fyi).

However, if using `ngOnChanges` you have to be careful that you don't modify any input values or it may get called twice accidentally (I haven't verified the truth of this though). You'd probably want the inputs to be readonly to be safe. I think the donut charts could see this a fix for the inputs too now that I think about it.

andrewazores: `ngOnChanges` will only fire when the `@Input` *reference* changes, AFAIK, so it'd take some extract work. The `ChangeDetectionStrategy` on the component would need to be changed and some event handling added so that the chart could notify Angular as needed. It seems like it would be more complex than the current solution at least.

Would something simple like this help with your performance concerns re. regenerating the chart?

```
  ngDoCheck(): void {
    if (!isEqual(this.config, this.prevConfig) || !isEqual(this.chartData, this.prevChartData)) {
      this.setupConfig();
      this.generateChart(this.config, isEqual(this.chartData, this.prevChartData));
    }
  }
```

This still recreates the config object, but that to me sounds like it should be a much lesser performance concern than regenerating the chart.

jiekang: Hmm right; a stack overflow I read says ngOnChanges does an === comparison? For what it's worth, it's being used in the donut chart code I wrote against the config and chartData values though and worked when I tested it. I am not sure if it's any better performance-wise than ngDoCheck now though.

You're right most important issue is the chart regeneration. I'll be happy if that alone is improved.

andrewazores: Huh. Not sure about that then. I tested it locally with ngOnChanges (with the if-statement removed) and added a console.log. The console message was only printed once and the chart was never updated past the first data point.

andrewazores: Anyway I'll commit the little piece I proposed above to try to avoid regenerating the charts.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2388-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2388-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/deployments-linechart/deployments-linechart.component.ts
jiekang: Ah, the reason it works in the donut charts is because it takes the 'pods' object as input from DeploymentsService. When the Observable emits again, it's always a new reference. For the sparkline, when you receive from the DeploymentsService you push that into an array so it's not a new reference. It wouldn't really work for the line charts as data grows. The donut chart is a little different since it's just showing the 'current state'. Sounds good re. avoiding regenerating charts.

src/app/space/create/deployments/deployments-linechart/deployments-linechart.component.ts
```diff
@@ -0,0 +1,178 @@
...
+  ngOnInit(): void {
+    this.setupConfigDefaults();
+    this.setupConfig();
+    this.generateChart(this.config, true);
+  }
+
+  ngDoCheck(): void {
+    if (!isEqual(this.config, this.prevConfig) || !isEqual(this.chartData, this.prevChartData)) {
+      this.setupConfig();
+      this.generateChart(this.config, isEqual(this.chartData, this.prevChartData));
```
jiekang: @andrewazores Oh. I just realized; I think when loading data via chart.load, by default, it doesn't remove any of the old data. You can specify to unload all prior to loading or to load only the new data. I'm not sure how easy it is to do in this situation. Can you verify if this is the case?

andrewazores: I see. Is the concern that without unloading the old data, the c3 component will render duplicates of each old data point? I see the documentation about the "unload" parameter when calling the load function of a c3 chart, but it's not exactly clear when or why this would be required.

jiekang: Yes. I ran into this with the c3js donut chart and though I haven't tested it, I think the other c3js charts follow the same behavior. The load's default behavior is more of an 'append'. For the donut chart, it was okay for me to just set 'config.data.unload' to true to unload all the previous data before adding the replacement data. It's not that expensive; I suppose if these charts are limited to 15 data points it wouldn't be that expensive either.

andrewazores: I manually tested to check for this behaviour with this patch:

https://paste.fedoraproject.org/paste/A5sAGbyfgV-xkQBq-xkqOw

Examining the console output, I don't see any difference in the behaviour of the internal data state of the c3 component with or without the "unload" property set. Either way, the data set grows by a single data point per update as expected, with no duplicates in the array or apparent changes to any prior data points. Perhaps this is just not the right internal state to examine.

I can include this property if you insist, but I'm not entirely sure of what exact effect it has on this component.

jiekang: @andrewazores It looks like it's fine. I did notice another issue where

```
isEqual(this.chartData, this.prevChartData)
```

is always true because there is a clone deep in the `setupConfig` call prior so it always regenerates anyways. The result prior to the `setupConfig` call needs to be extracted to a boolean.

andrewazores: Ah thanks, nice catch.

andrewazores: I re-tested the `unload` property with that change and the behaviour still looks correct with or without the property. The chart visibly behaves more nicely with the constant extracted since that allows us to actually avoid the regeneration as intended, so there's no visual flicker on update - just a smooth update of the chart dataset.

Perhaps we should make a PR to upstream pf-ng to enhance the sparkline component similarly?

jiekang: Yeah my tests had a similar result. It doesn't reload the dataset and only adds what's new.

jiekang:
@andrewazores I think a PR to upstream for the sparkline charts would be good. Your change to the implementation of ngDoCheck is small and effective.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2388-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2388-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/apps/deployment-details.component.spec.ts
```diff
@@ -68,15 +71,26 @@ class FakePfngChartSparkline {
   @Input() chartData: any;
 }
 
+@Component({
+  selector: 'deployments-linechart',
```
joshuawilson: Please add the `f8-` namespace prefix to the selector.

src/app/space/create/deployments/deployments-linechart/deployments-linechart.component.ts
```diff
@@ -0,0 +1,179 @@
...
+  merge,
+  uniqueId
+} from 'lodash';
+
+import { DeploymentsLinechartConfig } from './deployments-linechart-config';
+import { DeploymentsLinechartData } from './deployments-linechart-data';
+
+@Component({
+  encapsulation: ViewEncapsulation.None,
+  selector: 'deployments-linechart',
```
joshuawilson: Please add the f8- namespace prefix to the selector.

andrewazores: Done. When are we expected to use this convention for selectors? Should it always be done?

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2388-6 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2388-fabric8-ui.badger.fabric8.io

PR #2391, opened 2018-01-15
fix(profile): Fix double scroll bar in profile

chrislessard:
This patch remedies a bug in the profile space's view, in which
two scrollbars are loaded if the list of users' spaces is sufficiently
long. It removes scrolling from a parent div in favour of our
infinite scroll component.

Issue: https://github.com/openshiftio/openshift.io/issues/1885

### DIFF
--- a/src/app/profile/overview/spaces/spaces.component.html
+++ b/src/app/profile/overview/spaces/spaces.component.html
@@ -25,7 +25,7 @@
       </div>
     </ng-template>
     <ng-template #showSpaces>
-      <div class="col-xs-12 f8-card-list">
+      <div class="no-vertical-scroll col-xs-12 f8-card-list">
         <div class="spaces-list-scroll" almInfiniteScroll
              (fetchMore)='fetchMoreSpaces($event)'>
           <ul class="list-group list-view-pf list-view-pf-view list-view-pf-striped">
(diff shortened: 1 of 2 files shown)


fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2391-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2391-fabric8-ui.badger.fabric8.io

src/app/profile/overview/spaces/spaces.component.less
```diff
@@ -10,6 +10,9 @@
   overflow-y: auto;
   overflow-x: hidden;
 }
+.no-scroll {
```
mindreeper2420: I'm thinking this might be better served in the `src/assets/stylesheets/shared` directory, so that it can be reused in the future? Might be worth naming it so that it also easily declares that this is for the vertical axis, and not all axes.

chrislessard: Good call, I've renamed and moved it out !

fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2391-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2391-fabric8-ui.badger.fabric8.io

PR #2396, opened 2018-01-15
feat(deployments): automatically scale memory stats

andrewazores:
This PR adds a class "ScaledMemoryStat" which implements the existing MemoryStat interface, which when given a value for the "used" and "quota" values, will automatically scale these values down and increase the "units" through bytes -> KB -> MB -> GB for display (ex. 2048 bytes will become 2 KB). This will be useful when the UI is hooked up to live data, where the memory stat numbers will be very large compared to the current mock data due to being sent in raw bytes.

### DIFF
--- a/src/app/space/create/deployments/models/scaled-memory-stat.ts
+++ b/src/app/space/create/deployments/models/scaled-memory-stat.ts
@@ -1,0 +1,35 @@
+import {
+  MemoryStat,
+  MemoryUnit
+} from './memory-stat';
+
+import { round } from 'lodash';
+
+export class ScaledMemoryStat implements MemoryStat {
+
+  private static readonly UNITS = ['bytes', 'KB', 'MB', 'GB'];
+
+  public units: MemoryUnit;
+
+  constructor(
+    public used: number,
+    public quota: number
+  ) {
+    let scale = 0;
+    if (this.used !== 0) {
+      while (this.used > 1024 && scale < ScaledMemoryStat.UNITS.length) {
+        this.used /= 1024;
+        this.quota /= 1024;
+        scale++;
+      }
+    } else {
+      while (this.quota > 1024 && scale < ScaledMemoryStat.UNITS.length) {
+        this.quota /= 1024;
+        scale++;
+      }
+    }
+    this.used = round(this.used, 1);
+    this.quota = round(this.quota, 1);
+    this.units = ScaledMemoryStat.UNITS[scale] as MemoryUnit;
+  }
+}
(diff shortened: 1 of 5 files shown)


src/app/space/create/deployments/models/scaled-memory-stat.ts
```diff
@@ -0,0 +1,35 @@
...
+  constructor(
+    public used: number,
+    public quota: number
+  ) {
+    let scale = 0;
+    if (this.used !== 0) {
+      while (this.used > 1024 && scale < ScaledMemoryStat.UNITS.length) {
+        this.used /= 1024;
+        this.quota /= 1024;
+        scale++;
```
jiekang: Do you think the algorithm should be tweaked not to scale in the situation where `quota < used && quota < scalepoint && used > scalepoint`?

E.g. if quota is 1023 bytes and used is 1025. Then it would have the 'same' scaling behavior when used is 1023 bytes and quota is 1025 bytes.

Or maybe it should scale based on the larger of the two. E.g. if quota is 5 GB, used is 1 MB, it would still be weird to see a quota in MB.

Mm... Maybe the two should just scale independently? In most cases the result would be the same as your current algorithm, but the edges might look more natural. Thoughts?

jiekang: I suppose this also suggests a common MemoryStat that's used twice, one to represent used, and one for quota.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2396-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2396-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/models/scaled-memory-stat.ts
andrewazores: How would a split "used" vs "quota" stat be used? The designs for the page don't show any place for independent units between the two, and without displaying independent units it doesn't make sense to present them with different scales.

There is definitely room for improvement in the algorithm, but I'm not exactly sure what would work best. In the case of 1 MB / 5 GB for example, the current algorithm would display (1 / 5120) MB. If it scales based on the larger unit, then this would just round down to 0 / 5 GB. In fact, anything under 50 MB would scale and round down to 0 when the quota is "natural in" GB.

andrewazores: (the rounding to 0 is based on the design where values are rounded to the nearest tenth - if we can go to the nearest hundredth that isn't so bad.)

jiekang: Right, the UI elements don't really cover this well.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2396-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2396-fabric8-ui.badger.fabric8.io

PR #2399, opened 2018-01-16
feat(deployments): retrieve list of apps and environments from WIT en…

andrewazores:
…dpoint

related to https://github.com/openshiftio/openshift.io/issues/1370

This is the initial commit for integration the Deployments area with the backend in WIT (as of https://github.com/fabric8-services/fabric8-wit/pull/1798 ). There are more commits to come in separate PRs to replace the other mock implementations within DeploymentsService with real HTTP requests.

In this commit, the DeploymentsService#isDeployedInEnvironment implementation is stubbed to always return 'true' - I haven't come up with a good way to implement this with the current state of the API responses from the backend. It'll be addressed with a follow-up commit so the rest of this work is not blocked, since the solution may need to involve backend support.

### DIFF
--- a/src/app/space/create/deployments/services/deployments.service.spec.ts
+++ b/src/app/space/create/deployments/services/deployments.service.spec.ts
@@ -1,151 +1,154 @@
 import {
   discardPeriodicTasks,
   fakeAsync,
+  TestBed,
   tick
 } from '@angular/core/testing';
 
+import {
+  HttpModule,
+  Response,
+  ResponseOptions,
+  XHRBackend
+} from '@angular/http';
+
+import {
+  MockBackend,
+  MockConnection
+} from '@angular/http/testing';
+
+import { createMock } from 'testing/mock';
+
+import { Observable } from 'rxjs';
+
+import {
+  AuthenticationService,
+  UserService
+} from 'ngx-login-client';
+
+import { WIT_API_URL } from 'ngx-fabric8-wit';
+
 import { Environment } from '../models/environment';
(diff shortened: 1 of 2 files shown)


andrewazores:
Also - this PR should probably not be merged until after the WIT PR (https://github.com/fabric8-services/fabric8-wit/pull/1798) has bubbled up to at least prod-preview as well.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2399-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2399-fabric8-ui.badger.fabric8.io

jiekang:
@andrewazores Regarding the implementation for `isDeployedInEnvironment`, isn't this information provided by `/apps/spaces/{spaceId}`?

E.g. from
```
{"data":{"applications":[{"name":"osio-prod","pipeline":[{"name":"stage","pods":{"running":1,"starting":0,"stopping":0,"total":1},"version":"1.0.1"}]}]}}
```

we can tell the application osio-prod is deployed in stage and nowhere else.

andrewazores:
@jiekang ah, I had it in my head that the response there would still include environments (pipeline stages) even when not deployed, but that may have been totally incorrect. I'll take a closer look with that follow-up PR, but it does seem like what you say would work. Thanks for the lead.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2399-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2399-fabric8-ui.badger.fabric8.io

PR #2403, opened 2018-01-17
fix(codebases): fix filter by creation date and last commit

ashumilova:
fixes openshiftio/openshift.io#1346

There is an error on filtering codebase details by creation date and last commit:
screenshot from 2018-01-17 12-44-40

because the codebase-item.component and codebases.component use different instances on codebase object, thus Github details fetched by one - is not accessed by another. To solve the problem - the GitHub data retrieval was moved to the codebases.component and this also solves that we keep fetching GitHub data each time on filtering.


fabric8cd:
@ashumilova fabric8/fabric8-ui:SNAPSHOT-PR-2403-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2403-fabric8-ui.badger.fabric8.io

PR #2420, opened 2018-01-19
fix(deployments): hide empty statuses from pod legend

jiekang:
This fixes the pod legend to not display statuses that have 0 pods. For example, if data comes back as 1 running, 0 stopped, it no longer shows 0 stopped.

Before:
legend-broken

After:
legend-fixed


fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2420-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2420-fabric8-ui.badger.fabric8.io

PR #2422, opened 2018-01-19
fix(deployments): tweak deployments toolbar width

aptmac:
This PR extends the width of the deployments toolbar in an attempt to more closely follow the UX design [0].

There was a line in the patternfly css that was fixing the width of the toolbar to 275px. Because of this, the default text inside of the toolbar ("Filter by Application Name ...") gets cut-off inside of the input box.

Here is what the toolbar looks like currently:
tweak-toolbar-width-1

And here's what the toolbar would look like after this PR:
tweak-toolbar-width-2

[0] https://redhat.invisionapp.com/share/YSDL0KRKW#/screens/260764293

### DIFF
--- a/src/app/space/create/deployments/deployments-toolbar/deployments-toolbar.component.less
+++ b/src/app/space/create/deployments/deployments-toolbar/deployments-toolbar.component.less
@@ -1,0 +1,5 @@
+@import (reference) '../../../../../assets/stylesheets/shared/osio.less';
+
+deployments-toolbar pfng-filter-fields .filter-pf.filter-fields .form-group {
+  width: 100%;
+}
--- a/src/app/space/create/deployments/deployments-toolbar/deployments-toolbar.component.ts
+++ b/src/app/space/create/deployments/deployments-toolbar/deployments-toolbar.component.ts
@@ -22,7 +22,8 @@ import {
 @Component({
   encapsulation: ViewEncapsulation.None,
   selector: 'deployments-toolbar',
-  templateUrl: './deployments-toolbar.component.html'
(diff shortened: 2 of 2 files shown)


mindreeper2420:
If the toolbar can’t accommodate a variable length of placeholder text, then I’d prefer that this be a PR against the main PatternFly repo, than a specific change to this one instance of the toolbar.

aptmac:
Sure that sounds good. The hard-coded value is in a pf-ng less file [1], so I'll patch it up there and open a PR for it.

[1] https://github.com/patternfly/patternfly-ng/blob/d93cd4176aa91cdfa8242bbe4c711af0e71e7bdb/src/app/filter/filter-fields.component.less#L8

jiekang:
@aptmac can this PR be closed then or is some part of it still needed?

aptmac:
@jiekang Yes this can be closed, although I'm not quite sure the best way to handle it in pf-ng is. After playing around a bit with the pf-ng repo, it feels like the hardcoded value was included so the size of the toolbar stays constant when cycling through the different toolbar options. My original idea of 100% width doesn't quite work because it'll cause the toolbar to vary in width based on the filter selected. Perhaps I should just open an issue for this on the pf-ng repo and request feedback.

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2422-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2422-fabric8-ui.badger.fabric8.io

jiekang:
@aptmac Okay that sounds good. If it doesn't come out well we can reopen this PR.

PR #2423, opened 2018-01-19
fix(deployments): fix deployments cards sparkline chart resizing when…

aptmac:
… window resizing

This PR addresses https://github.com/openshiftio/openshift.io/issues/1859, in which the size of the sparkline charts inside of deployment cards do not adapt until the charts themselves are updated.

After applying this one line of less, the chart will resize to fill it's appropriate width. When resizing the browser window now, the sparkline charts will resize, and then will flicker as they are updated and redrawn when new data comes in (this was previously the way the charts would resize themselves).

Also to note, I tested this with a recent sparkline fix in patternfly-ng [0]. With commit [0] applied, the sparkline charts don't regenerate on data update, and instead reload themselves (which looks nice and smooth). However, because the entire chart is no longer recreated on an update, when changing the browser size from large to small (or visa-versa) the chart width stays static and won't change on update (like how it currently does). As a result, going from a large to small screen size will result in something that looks like this (and the chart will stay at this width while it reloads on data update): 
fix-sparkline-charts-width-1

This PR will also fix the above image.

[0] https://github.com/patternfly/patternfly-ng/commit/a908031917b1a237ebe4260c605e6db00dc76efb#diff-7388380f9f16c2007382b5e0e255ba2c


fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2423-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2423-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/apps/deployment-details.component.less
```diff
@@ -1,5 +1,9 @@
 @import (reference) '../../../../../assets/stylesheets/shared/osio.less';
 
+pfng-chart-sparkline {
```
aptmac: errrr .. I think this line is better off as: 
`.deployment-chart pfng-chart-sparkline {`
to make it more specific?

mindreeper2420:
Is this something that should be applied to PatternFly-ng, so that this does not need to be repeated elsewhere?

aptmac:
Sure, that sounds like a plan to me.

aptmac:
So after a bit more digging, the fix that I had wasn't the fix I actually wanted. It's not an issue with the PatternFly stylesheets, but rather a `float: left`that was fixing the width of the sparkline chart.

To test, I had added a function to the example page of the pf-ng Sparkline [1] that would continuously update the chart in an attempt to recreate the environment of the deployments chart. The chart was able to resize appropriately, so I started looking on our side of things. After inspecting the chart dimensions, I noticed that `float: left`was causing the width of the chart to become a fixed value (and would only reset on a chart redraw), and without it the dimensions become auto x auto (similar to the network chart) - which allows the chart to resize appropriately. 

[1] https://github.com/aptmac/patternfly-ng/commit/bde317d65ffe16badd9e36d8b6334403eda280c2

PR #2424, opened 2018-01-19
Add null checks before accessing space references

jiekang:
This adds some null checks to code that accesses space references. The space can become null when Context changes (e.g. in the process of routing to a non-space related page). The code may still run before the component gets cleaned up, so they should be robust against a null space.

The commit `fix(runtime-console-f8ui-space-service): add null check to space references` is the direct fix for the issue [1] of accessing the Profile page from the Environments page. The rest do not fix any known issues (afaik) but should increase safety regardless. I'm open to separating into various PRs if needed. Addressing [1] is a main concern.

[1] https://github.com/openshiftio/openshift.io/issues/1940

Note: Tested with local instance targeting production. I verified this by clicking the Profile button in the top left username dropdown after visiting the Environments page.

### DIFF
--- a/src/app/dashboard-widgets/edit-space-description-widget/edit-space-description-widget.component.html
+++ b/src/app/dashboard-widgets/edit-space-description-widget/edit-space-description-widget.component.html
@@ -10,7 +10,7 @@
               *ngIf="isEditing && (isEditable() | async)"
               (keyup.enter)="onUpdateDescription(description.value)"
               (keydown.enter)="preventDef($event)"
-              #description>{{space.attributes.description}}</textarea>
+              #description>{{space?.attributes.description}}</textarea>
     <div *ngIf="isEditing">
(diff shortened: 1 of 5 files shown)


fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2424-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2424-fabric8-ui.badger.fabric8.io

jiekang:
After some thought over the weekend: Another possibility is adding a filter at the right spot(s) in the Observable chain that will prevent the emission of a `null` space value. This needs more codebase knowledge to determine if acceptable and is a significant change in behavior. However if `null` isn't a useful value to emit, then it makes sense not to emit it instead of asking users to handle that possibility.

joshuawilson:
cc @dgutride

PR #2431, opened 2018-01-22
fix(deployments): clicking details does not collapse cards

andrewazores:
prevent click even propagation on deployment-details child component, so that
the "collapsed" property is not toggled, which causes the details to be hidden
when attempting to select card text or click on charts

closes https://github.com/openshiftio/openshift.io/issues/1967

### DIFF
--- a/src/app/space/create/deployments/apps/deployment-card.component.html
+++ b/src/app/space/create/deployments/apps/deployment-card.component.html
@@ -38,8 +38,8 @@
           </div>
         </div>
       </div>
-      <deployment-details [collapsed]="collapsed" [applicationId]="applicationId"
-        [environment]="environment" [spaceId]="spaceId"></deployment-details>
+      <deployment-details (click)="$event.stopPropagation()" [collapsed]="collapsed"
+        [applicationId]="applicationId" [environment]="environment" [spaceId]="spaceId"></deployment-details>
     </div>
   </div>
 </ng-container>


src/app/space/create/deployments/apps/deployment-card.component.html
```diff
@@ -38,8 +38,8 @@
           </div>
         </div>
       </div>
-      <deployment-details [collapsed]="collapsed" [applicationId]="applicationId"
-        [environment]="environment" [spaceId]="spaceId"></deployment-details>
+      <deployment-details onclick="event.preventDefault(); event.stopPropagation();" [collapsed]="collapsed"
+        [applicationId]="applicationId" [environment]="environment" [spaceId]="spaceId"></deployment-details>
```
jiekang: Some other component views [1] also stop propagation but only calling `event.stopPropagation()`. What does `preventDefault` do? Is it necessary elsewhere as well?

[1] `deployment-card.component.html` and `deployments-donut.component.html`

andrewazores: I just tested it quickly and `stopPropagation()` on its own seems sufficient, but I don't know what's best practice here really. I was hoping Adam might have some input here.

There's this discussion that has quite a bit of info: https://stackoverflow.com/questions/5963669/whats-the-difference-between-event-stoppropagation-and-event-preventdefault

But I still don't have a clear picture...

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2431-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2431-fabric8-ui.badger.fabric8.io

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2431-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2431-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/apps/deployment-card.component.html
mindreeper2420: from an interaction perspective, if this card is clickable, there should be a cursor change and some kind of highlight on the card when the user hovers over it.

The Single-Select Card View demo has the best example: http://www.patternfly.org/pattern-library/content-views/card-view/. The classes are included with the framework.

jiekang: The cursor change and highlight are good things to have, but I think it would fall into a separate issue and PR.

I think if only `stopPropagation()` is sufficient, we should leave it at that for now.

mindreeper2420: I'm good with the style pieces being a separate PR/issue.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2431-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2431-fabric8-ui.badger.fabric8.io

PR #2434, opened 2018-01-23
Refactor Deployments observables

jiekang:
This addresses [1], moving spaceId subscription down to the points where it's data is used.

[1] https://github.com/openshiftio/openshift.io/issues/1980

### DIFF
--- a/src/app/space/create/deployments/resource-usage/resource-card.component.ts
+++ b/src/app/space/create/deployments/resource-usage/resource-card.component.ts
@@ -1,44 +1,59 @@
 import {
   Component,
   Input,
+  OnDestroy,
   OnInit
 } from '@angular/core';
 
-import { Observable } from 'rxjs';
+import { Observable, Subscription } from 'rxjs';
 
-import { MemoryStat } from 'app/space/create/deployments/models/memory-stat';
 import { Environment } from '../models/environment';
+import { MemoryStat } from '../models/memory-stat';
+import { Stat } from '../models/stat';
 import { DeploymentsService } from '../services/deployments.service';
 
 @Component({
   selector: 'resource-card',
   templateUrl: 'resource-card.component.html'
 })
-export class ResourceCardComponent implements OnInit {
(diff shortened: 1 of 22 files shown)


src/app/space/create/deployments/resource-usage/resource-card.component.ts
```diff
@@ -1,44 +1,59 @@
...
+      .subscribe((active: boolean) => {
+        this.active = active;
+        if (active) {
+          this.memUnit = this.deploymentsService.getEnvironmentMemoryStat(spaceId, this.environment.name)
+          .first()
+          .map((stat: MemoryStat) => stat.units);
+        }
+      }));
+      this.cpuStat = this.deploymentsService.getEnvironmentCpuStat(spaceId, this.environment.name);
+      this.memStat = this.deploymentsService.getEnvironmentMemoryStat(spaceId, this.environment.name);
```
andrewazores: Assigning Observable properties only after some other Observable has emitted a value is what got us into trouble the first time around when the spaceId schedular changed in Dana's `modify-rxjs-behavior` branch. If we can't work under the assumption that these things resolve synchronously on subscribe, then it's possible that we will be instantiating child components with not-yet-defined bindings for Observables (like `memStat` here).

IMO the clearest solution to this problem is to ensure that all component properties are defined synchronously by the end of `ngOnInit` at the latest. For the case of `memStat` for example, this would mean immediately defining it as a `Subject`, which would then be passed to the component's children, and they can subscribe immediately even if the spaceId has not yet resolved. This component would then create its own subscription via `deploymentsService.getEnvironmentMemoryStat(spaceId, this.environment.name).subscribe(this.memorySubject)`, so that service emissions are relayed to child components.

jiekang: Good point. I'll rework this and use Dana's change to test it. I'll ping you again when it's ready for review

jiekang:
After further work in this area, it looks like the changes will need to be more involved. I'm closing this PR for now. I'll reopen or open another PR when ready.

PR #2441, opened 2018-01-24
fix(stylelint): Upgrade and set failure rules

mindreeper2420:
Update stylelint-webpack-plugin to use the latest version that allows for console printed errors with the webpack dev server running. Previous versions failed on error, even when setting the specific dev config to not fail.


fabric8cd:
@mindreeper2420 fabric8/fabric8-ui:SNAPSHOT-PR-2441-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2441-fabric8-ui.badger.fabric8.io

PR #2442, opened 2018-01-24
fix(deployments): fix deployments donut chart to properly apply c3js css

aptmac:
This PR addresses [osio issue 1939](https://github.com/openshiftio/openshift.io/issues/1939) [0], in which the donut chart css gets masked due to Angular's stylesheet encapsulation.

This patch changes the ViewEncapsulation of both the deployments donut component and donut chart component to be 'none'. As a result, the stylesheet values for the donut chart are now visible and applied.

For example, here is the current stylesheet applied to the c3 tooltip:
head

And here it is if only the chart component is given view encapsulation none:
chart-only

And here is the result of having both the donut and chart components having view encapsulation none:
both

This allows for the c3 tooltip-container values[1] to be applied.

[0] https://github.com/openshiftio/openshift.io/issues/1939
[1] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/space/create/deployments/deployments-donut/deployments-donut-chart/deployments-donut-chart.component.less#L25

### DIFF
--- a/src/app/space/create/deployments/deployments-donut/deployments-donut.component.ts
+++ b/src/app/space/create/deployments/deployments-donut/deployments-donut.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, OnInit } from '@angular/core';
+import { Component, Input, OnInit, ViewEncapsulation } from '@angular/core';
 
 import { debounce, isNumber } from 'lodash';
 import { NotificationType } from 'ngx-base';
@@ -11,6 +11,7 @@ import { Pods } from '../models/pods';
 import { DeploymentsService } from '../services/deployments.service';
 
 @Component({
+  encapsulation: ViewEncapsulation.None,
   selector: 'deployments-donut',
   templateUrl: './deployments-donut.component.html',
   styleUrls: ['./deployments-donut.component.less']
--- a/src/app/space/create/deployments/deployments-donut/deployments-donut-chart/deployments-donut-chart.component.ts
+++ b/src/app/space/create/deployments/deployments-donut/deployments-donut-chart/deployments-donut-chart.component.ts
@@ -5,7 +5,8 @@ import {
   OnChanges,
   OnDestroy,
   OnInit,
-  SimpleChanges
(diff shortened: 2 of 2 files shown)


fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2442-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2442-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/deployments-donut/deployments-donut.component.ts
```diff
@@ -11,6 +11,7 @@ import { Pods } from '../models/pods';
 import { DeploymentsService } from '../services/deployments.service';
 
 @Component({
+  encapsulation: ViewEncapsulation.None,
```
jiekang: If there is any css targeting child classes in `deployments-donut.component.less` they should be moved into `deployments-donut-chart.component.less`. I'd like the `ViewEncapsulation` to remain as `Default` for `deployments-donut.component.ts`.

As an aside, after investigation the css for `deployments-donut-chart.component.ts` needs a bit of tweaking but I can do that in a separate PR and issue.

mindreeper2420: All other components have (or should have) view encapsulation to none, as we do not include the shared stylesheets in every component (just referenced). I would leave this as none, until it is decided whether the entire UI can/should/needs to be switched.

jiekang: @mindreeper2420 Oh. I wasn't aware of that. This isn't currently the case in the components under the deployments page. Most of them have the default view encapsulation. We can fix this though.

mindreeper2420: If everything is working now, I wouldn't worry about it. Something that can be brought up with Joshua, as there was an issue on it (ViewEncapsulation) at one point

PR #2448, opened 2018-01-25
fix(link): multi-cluster:start using native openshift linking service

sbose78:
**--- do not merge  ---**

**Depends on**

1. https://github.com/fabric8-ui/ngx-login-client/pull/98

2. The `Getting Started` page currently uses the `https://sso.prod-preview.openshift.io/auth/realms/fabric8/broker/openshift-v3/token` url to determine if the user is connected. This has to be changed to `GET /api/token?for=<user's cluster url` . 

@dgutride , I see that we have helper methods to fetch tokens in the Provider service as well as the `ngx-login-client`'s AuthenticationService . 

the method called here https://github.com/fabric8-ui/ngx-login-client/blob/master/src/app/auth/authentication.service.ts#L41 now needs the user's cluster url too.
Do we make changes in that method? Or shall we get the helper methods in provider ts ?

**What this PR accomplishes**
Previously, we linked Github using the Auth service & OpenShift using keycloak.
In this PR, we are linking OpenShift using Auth - this would be critical for multi-cluster support.

### DIFF
--- a/src/app/shared/account/provider.service.ts
+++ b/src/app/shared/account/provider.service.ts
@@ -25,13 +25,14 @@ export class ProviderService {
   /**
    * Link an OpenShift.com account to the user account
    *
+   * @param cluster URL of the openshift cluster the user is associated with
    * @param redirect URL to be redirected to after successful account linking
    */
-  linkAll(redirect: string): void {
-        let openShiftLinkingRedirectUrl = this.getLegacyLinkingUrl('openshift-v3', redirect);
+  linkAll(cluster: string, redirect: string): void {
+        let redirectToGithubLinkURL = window.location.origin + '/_gettingstarted?wait=true&link=' + encodeURIComponent('https://github.com');
         // after linking github, proceed with linking openshift-v3,
         // hence passing openshift linking url as a redirect.
-        this.linkGitHub(openShiftLinkingRedirectUrl);
+        this.linkOpenShift(cluster, redirectToGithubLinkURL);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


sbose78:
tests pending!

dgutride:
We need to make token retrieval all go through the ngx-login-client/authenticationservice helper method - where is the provider method you referred to @sbose78 ?

maxandersen:
related to https://github.com/openshiftio/openshift.io/issues/1685

sbose78:
I see we are calling `getGithubStatus()` here, https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/shared/account/provider.service.ts#L73

The URL pattern for getting the oso token is the exactly the same as above ( except of course the extra param `force_pull` ). So that makes me wonder if we were moving things here..

If we are gonna keep using the ngx-login-client, then we would need to change the way we do https://github.com/fabric8-ui/ngx-login-client/blob/master/src/app/auth/authentication.service.ts#L41

` this.openShiftToken = this.createFederatedToken(this.openshift, (response: Response) => response.json() as Token);`

needs to be something like

` this.openShiftToken = this.createFederatedToken(this.openshift, user_cluster, (response: Response) => response.json() as Token);`

sbose78:
@dgutride I'm a bit unsure if we can pass the user's cluster url when we call `this.openshiftToken` this.openShiftToken = this.createFederatedToken(this.openshift, user_cluster, (response: Response) => response.json() as Token);

I'm a bit unsure how to do so, any pointers or design suggestions ?

src/app/shared/account/provider.service.ts
```diff
@@ -115,8 +105,22 @@ export class ProviderService {
    * @param redirect URL to be redirected to after successful account linking
    */
   link(provider: string, redirect: string): void {
-    let url = this.getLegacyLinkingUrl(provider, redirect);
-    this.redirectToAuth(url);
+    let linkURL = this.linkUrl + '?for=' + provider + '&redirect=' +  encodeURIComponent(redirect) ;
+    console.log('attempting to connect ' + provider + ' using ' + linkURL);
```
dgutride: console.log should be removed - there is a log service

src/app/shared/account/provider.service.ts
```diff
@@ -115,8 +105,22 @@ export class ProviderService {
...
+    this.http
+    .get(linkURL)
+    .map(response => {
+      // TODO: what happens to this when the response is not a pure json
+      let redirectInfo = response.json() as Link;
+      this.redirectToAuth(redirectInfo.redirect_location);
+
+      // todo - handle redirect info?
+      console.log(redirectInfo);
```
dgutride: more console.logs - pls remove or replace

src/app/shared/account/provider.service.ts
```diff
@@ -115,8 +105,22 @@ export class ProviderService {
...
+    })
+    .catch((error) => {
+      console.log('error while linking ' + linkURL);
```
dgutride: See above...

joshuawilson:
[test]

dgutride:
This should be closed now @sbose78 - moving to closed, just so it doesn't get merged.

sbose78:
Thanks!

PR #2455, opened 2018-01-29
Refactor deployments observables

jiekang:
This addresses https://github.com/openshiftio/openshift.io/issues/1980

Observables are now moved down the component tree to be subscribed in as late as possible. Certain observables are now also Subjects instead, so that they will be available on component creation, preventing null exceptions when asynchronous.

### DIFF
--- a/src/app/space/create/deployments/apps/deployment-card.component.ts
+++ b/src/app/space/create/deployments/apps/deployment-card.component.ts
@@ -27,7 +27,7 @@ export class DeploymentCardComponent implements OnDestroy, OnInit {
   private static readonly DEBOUNCE_TIME: number = 5000; // 5 seconds
   private static readonly MAX_DEBOUNCE_TIME: number = 10000; // 10 seconds
 
-  @Input() spaceId: string;
+  @Input() spaceId: Observable<string>;
   @Input() applicationId: string;
   @Input() environment: Environment;
 
@@ -46,6 +46,7 @@ export class DeploymentCardComponent implements OnDestroy, OnInit {
   subscriptions: Array<Subscription> = [];
 
   private readonly debouncedUpdateDetails = debounce(this.updateDetails, DeploymentCardComponent.DEBOUNCE_TIME, { maxWait: DeploymentCardComponent.MAX_DEBOUNCE_TIME });
+  private spaceIdReference: string;
 
   constructor(
     private deploymentsService: DeploymentsService,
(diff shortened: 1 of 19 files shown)


fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2455-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2455-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/apps/deployment-card.component.ts
```diff
@@ -59,32 +60,35 @@ export class DeploymentCardComponent implements OnDestroy, OnInit {
...
+                this.deploymentsService.getVersion(spaceId, this.applicationId, this.environment.name);
+
+              this.logsUrl =
+                this.deploymentsService.getLogsUrl(spaceId, this.applicationId, this.environment.name);
+
+              this.consoleUrl =
+                this.deploymentsService.getConsoleUrl(spaceId, this.applicationId, this.environment.name);
+
+              this.appUrl =
+                this.deploymentsService.getAppUrl(spaceId, this.applicationId, this.environment.name);
```
andrewazores: cpuStat, version, and the URL Observables would ideally also be backed by Subjects here. This can be done in a follow-up commit on top of this patch series

src/app/space/create/deployments/deployments-donut/deployments-donut.component.ts
```diff
@@ -44,19 +44,33 @@ export class DeploymentsDonutComponent implements OnInit {
...
-    this.pods = this.deploymentsService.getPods(this.spaceId, this.applicationId, this.environment.name);
-    this.pods.subscribe(pods => {
-      this.replicas = pods.total;
-      if (!this.scaleRequestPending) {
-        this.desiredReplicas = this.replicas;
+
+    this.subscriptions.push(
+      this.spaceId.subscribe((spaceId: string) => {
+        this.spaceIdRef = spaceId;
+        this.pods = this.deploymentsService.getPods(spaceId, this.applicationId, this.environment.name);
```
andrewazores: same here, Subject backing on pods

src/app/space/create/deployments/apps/deployment-details.component.ts
```diff
@@ -93,44 +93,46 @@ export class DeploymentDetailsComponent {
...
-      this.cpuData.yData.push(stat.used);
-      this.cpuData.xData.push(this.cpuTime++);
-      this.trimSparklineData(this.cpuData);
+    this.subscriptions.push(this.spaceId.subscribe((spaceId: string) => {
+
+      this.cpuStat =
+        this.deploymentsService.getDeploymentCpuStat(spaceId, this.applicationId, this.environment.name);
+
+      this.memStat =
+        this.deploymentsService.getDeploymentMemoryStat(spaceId, this.applicationId, this.environment.name);
```
andrewazores: and these two

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2455-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2455-fabric8-ui.badger.fabric8.io

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2455-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2455-fabric8-ui.badger.fabric8.io

dgutride:
Why did you choose to move more observables into the components instead of simplifying the component logic?  Adding more subscriptions is going to cause problems long term - we'd like to start moving toward a presentational component architecture and if you are loading the same data multiple times in the component tree, this could be an excellent example:  https://blog.angular-university.io/angular-component-design-how-to-avoid-custom-event-bubbling-and-extraneous-properties-in-the-local-component-tree/

jiekang:
@dgutride 

The impetus for this was the proposed change that makes the spaceId scheduler async. This patch is a step towards protecting us from the difference in behavior that comes from that. We weren't sure if that was going to be merged but figured making it safe in case that ever happens would be a good idea.

The async scheduler causes two main behavior changes:

1. Components do not wait for observable inputs before initializing. This meant that many components ended up initializing with null values. The use of Subjects and moving subscriptions to the point that the data is needed mitigates this.

2. Change detection on the observables doesn't work. This one sounds wrong, but I haven't found a solution or a different explanation for what occurs with the async scheduler. One solution is manual change detection. This PR doesn't solve this issue, but it would need to be looked at if the scheduler becomes async. It's not hard to do but does require lots of code changes.

Because all of our data relies on the spaceId to be acquired, my current guess is the async scheduler moves it out of the zone, and thus everything relying on the spaceId in our components is also out of the zone. However, I could be completely wrong and maybe there is some critical flaw that I'm missing to cause all of this.

You can see some of this behavior in this branch I made for testing this out: https://github.com/jiekang/fabric8-ui/commits/deployments-observables-refactor


Regarding the linked article and 'adding more subscriptions'. I'm a little confused as the article suggests accessing the service directly for subscriptions from the component using the data, rather than doing so in a parent 'container' component, and bubbling that down. That would inevitably lead to more subscriptions which seems to contradict your request to minimize the number of subscribe calls. The article describes state A and then how it's improved' as state B. Do you mean we want to be in state A?

This patch is very much like state B in the linked article, where it is subscribing for the data where it's needed; the only difference is we're subscribing off of Observable inputs that are bubbled down, while they're subscribing directly from the Service. It would be easier after this patch to then convert these subscriptions to be directly from the Service, instead of from the inputs, if that's desired.

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2455-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2455-fabric8-ui.badger.fabric8.io

joshuawilson:
@dgutride do you approve?

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2455-6 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2455-fabric8-ui.badger.fabric8.io

jiekang:
@dgutride When you have time could you comment more on how you'd like to see changes in Observable usage moving forward?

jiekang:
@dgutride Hi; could you please take a look at this again?

joshuawilson:
@jiekang please rebase.
@dgutride please review.

jiekang:
I am not planning to rebase and rework this and the changes are not feature related. Issue has been closed and I am closing this PR.

PR #2474, opened 2018-02-01
chore(env): use proper bash default values

kwk:
Replace more verbose if-then syntax to set variables to default values with shorted and easier to maintain `${foo:-bar}` syntax.


joshuawilson:
Please explain why you are making these changes. 
What is the value of them?
Why should it be merged?

kwk:
@joshuawilson 

> Please explain why you are making these changes.

I want to understand this code a bit and found that something looks just wrong or outdated and no longer used.

> What is the value of them?

No more unused files gives the sense of a cleaned up system and less to worry about for somebody like me who's new to the code.

> Why should it be merged?

Do you want to have dead files with over 5000 LOC in your code, really?

michaelkleinhenz:
Just an observation: given we painstakingly make sure every whitespace and every semicolon is checked, we have a lot of dead/unused/outdated code and configurations in the projects. This does not only look bad to anyone looking at it from the community, it makes it exceptionally hard to understand what's going on on the project and it's build. And this is an imminent risk to the stability.

joshuawilson:
@kwk it is harder to validate a PR that has multiple changes. You already have 2 things in this that are not related. one I could merge now and the other will take some effort to verify it doesn't break things. Please don't make this a WIP but make multiple PRs that can stand on there on.

joshuawilson:
The reason for the questions is that this is PR that was not requested, not in the plan, and has no issue or description. Just looking for some clarity.

environments/devshift-cluster.deploy.sh
```diff
@@ -18,24 +18,12 @@ export PROXIED_K8S_API_SERVER="${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_P
...
-if [ -z "${K8S_API_SERVER_PROTOCOL}" ]; then
-  export K8S_API_SERVER_PROTOCOL="http"
-fi
-if [ -z "${K8S_API_SERVER_BASE_PATH}" ]; then
-  export K8S_API_SERVER_BASE_PATH="/_p/oso"
-fi
-if [ -z "${WS_K8S_API_SERVER}" ]; then
-  export WS_K8S_API_SERVER=${PROXIED_K8S_API_SERVER}
-fi
+export OAUTH_ISSUER=${OAUTH_ISSUER:-"https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}"}
```
joshuawilson: Why remove the IF check? What does that buy you? Isn't it there so that you only export things you have?

kwk: @joshuawilson 

>  Isn't it there so that you only export things you have?

No, the `-z` checks if a value is not set (is zero). If something is zero, then we set it to a default. And for this there's a shortcut in bash that goes like this:

```sh
foo=${blabla:-"bar"}
```

This sets the variable `foo` to the value `"bar"` but only if `blabla` wasn't set above (e.g. through an environment variable). I find this much shorter, readable, hence maintainable than the `if` construct. That's why I've made the change.

joshuawilson:
@kwk I can't resolve this here, please rebase.

kwk:
> @kwk I can't resolve this here, please rebase.

@joshuawilson done

PR #2475, opened 2018-02-01
fix(deployments): use null check instead for desired replicas

jiekang:
This is a tiny fix. The intention was to not allow null. This now allows 0, which is a valid value for desired replicas. `==` is used to also check for undefined without having to write another or statement.

Sorry for the mistake;


fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2475-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2475-fabric8-ui.badger.fabric8.io

PR #2479, opened 2018-02-02
WIP fix(profile): fetch openshift account status from auth-native api

sbose78:
related to openshiftio/openshift.io#1685

### DIFF
--- a/src/app/profile/update/update.component.ts
+++ b/src/app/profile/update/update.component.ts
@@ -85,8 +85,8 @@ export class UpdateComponent implements AfterViewInit, OnInit {
     this.subscriptions.push(auth.gitHubToken.subscribe(token => {
       this.gitHubLinked = (token !== undefined && token.length !== 0);
     }));
-    this.subscriptions.push(auth.openShiftToken.subscribe(token => {
-      this.openShiftLinked = (token !== undefined && token.length !== 0);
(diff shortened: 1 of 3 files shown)


src/app/profile/update/update.component.ts
```diff
@@ -85,8 +85,8 @@ export class UpdateComponent implements AfterViewInit, OnInit {
     this.subscriptions.push(auth.gitHubToken.subscribe(token => {
       this.gitHubLinked = (token !== undefined && token.length !== 0);
     }));
-    this.subscriptions.push(auth.openShiftToken.subscribe(token => {
-      this.openShiftLinked = (token !== undefined && token.length !== 0);
+    this.subscriptions.push(auth.isOpenShiftConnected(this.loggedInUser.attributes.cluster).subscribe(isConnected => {
```
dgutride: This is risky to do here - loggedInUser.attributes might not be populated (based on line 82) - it really needs to only be created based on loggedInUser being available.  My recommendation would be we make loggedInUser a variable available on the user service instead of an observable - it can only change with a full window refresh anyway, so watching it is nonsensical and we shouldn't have ever built it this way in the first palce.

sbose78: Hi @dgutride would you able to take up the design refactoring please ( in this branch/PR or a separate one ). Not competent enough in angular to do quick-enough design revamps :)

PR #2494, opened 2018-02-05
fix(feature-flag): add missing provider to app.module

corinnekrych:
Part of openshiftio/openshift.io#2025

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -59,7 +59,7 @@ import { AppState, InternalStateType } from './app.service';
 // Footer & Header
 import { FeatureBannerComponent } from './feature-flag/banner/feature-banner.component';
 import { FeatureFlagResolver } from './feature-flag/resolver/feature-flag.resolver';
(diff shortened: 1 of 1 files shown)


src/app/app.module.ts
```diff
@@ -59,7 +59,7 @@ import { AppState, InternalStateType } from './app.service';
 // Footer & Header
 import { FeatureBannerComponent } from './feature-flag/banner/feature-banner.component';
 import { FeatureFlagResolver } from './feature-flag/resolver/feature-flag.resolver';
-import { FeatureTogglesService } from './feature-flag/service/feature-toggles.service';
+import { FABRIC8_FEATURE_TOGGLES_API_URL, featureTogglesApiUrlProvider, FeatureTogglesService } from './feature-flag/service/feature-toggles.service';
```
jiekang: As far as I can tell FABRIC8_FEATURE_TOGGLES_API_URL isn't used anywhere in this file so it doesn't need to be imported. Am I missing something?

gastaldi:
I can confirm that this makes the fabric8-ui usable again locally

PR #2504, opened 2018-02-06
feat(deployments): deployments charts display last fifteen minutes of…

aptmac:
… data on initialization

This PR addresses [openshift.io issue 2038](https://github.com/openshiftio/openshift.io/issues/2038)[0], and introduces functionality to retrieve the last 15 minutes worth of deployments chart data when the charts are initialized. The data is retrieved by using the statseries API, and specifying a start and end time. 

In deployments.service.ts, getTimeConstrainedTimeseriesData() is a private function that gets the response from the API using a start and end time, and getDeploymentTimeConstrainedStats() is the public function that cleans and returns information to the component. The timestamps corresponding to the time constrained data are retained in response to [todo comments](https://github.com/aptmac/fabric8-ui/blob/master/src/app/space/create/deployments/services/deployments.service.ts#L236), and are propagated to the component (but not currently used). 

In deployment-details.component.ts, the initial stats are requested and processed prior to the polling of cpu/memory/network data in ngOnInit().

There are included unit tests that check the functionality of:
- the component asking for the last 15 minutes worth of data
- the component including the cpu, memory, and network data prior to the later polled data
- the service handling an empty statseries response
- the service correctly combining quota and statseries responses

Here's an image of what the charts look like on creation currently:
before

And here's what they look like with this patch applied:
charts-after

[0] https://github.com/openshiftio/openshift.io/issues/2038
[1] https://github.com/aptmac/fabric8-ui/blob/master/src/app/space/create/deployments/services/deployments.service.ts#L236

### DIFF
--- a/src/app/space/create/deployments/apps/deployment-details.component.ts
+++ b/src/app/space/create/deployments/apps/deployment-details.component.ts
@@ -17,7 +17,7 @@ import { Environment } from '../models/environment';
 import { MemoryStat } from '../models/memory-stat';
 import { Pods } from '../models/pods';
 import { ScaledNetworkStat } from '../models/scaled-network-stat';
-import { DeploymentsService } from '../services/deployments.service';
+import { DeploymentsService, TimeConstrainedStats } from '../services/deployments.service';
 
 import { DeploymentsLinechartConfig } from '../deployments-linechart/deployments-linechart-config';
 import { DeploymentsLinechartData } from '../deployments-linechart/deployments-linechart-data';
@@ -78,6 +78,7 @@ export class DeploymentDetailsComponent {
   hasPods: Subject<boolean> = new ReplaySubject<boolean>(1);
   cpuStat: Observable<CpuStat>;
   memStat: Observable<MemoryStat>;
+  initialChartStats: Observable<TimeConstrainedStats>;
   cpuTime: number;
   memTime: number;
   cpuVal: number;
@@ -107,11 +108,16 @@ export class DeploymentDetailsComponent {
     this.cpuTime = 1;
     this.memTime = 1;
 
+    this.initialChartStats =
+      this.deploymentsService.getDeploymentTimeConstrainedStats(this.spaceId, this.applicationId, this.environment.name, DeploymentDetailsComponent.DEFAULT_SPARKLINE_DATA_DURATION);
+
     this.cpuStat =
       this.deploymentsService.getDeploymentCpuStat(this.spaceId, this.applicationId, this.environment.name);
 
     this.memStat =
       this.deploymentsService.getDeploymentMemoryStat(this.spaceId, this.applicationId, this.environment.name);
+
+    this.processTimeConstrainedStats(this.initialChartStats);
 
     this.subscriptions.push(this.cpuStat.subscribe(stat => {
       this.cpuVal = stat.used;
@@ -127,7 +133,7 @@ export class DeploymentDetailsComponent {
       this.memMax = stat.quota;
       this.memData.total = stat.quota;
       this.memData.yData.push(stat.used);
-      this.memData.xData.push(this.cpuTime++);
+      this.memData.xData.push(this.memTime++);
       this.memUnits = stat.units;
       this.trimSparklineData(this.memData);
     }));
@@ -166,6 +172,33 @@ export class DeploymentDetailsComponent {
         yData.splice(1, elementsToRemoveCount);
       });
     }
+  }
+
+  private processTimeConstrainedStats(stats: Observable<TimeConstrainedStats>): void {
+    this.subscriptions.push(stats.subscribe(s => {
+      s.cpu.forEach(e => {
+        this.cpuVal = e.data.used;
+        this.cpuMax = e.data.quota;
+        this.cpuData.total = e.data.quota;
+        this.cpuData.yData.push(e.data.used);
+        this.cpuData.xData.push(this.cpuTime++);
+      });
+      s.memory.forEach(e => {
+        this.memVal = e.data.used;
+        this.memMax = e.data.quota;
+        this.memData.total = e.data.quota;
+        this.memData.yData.push(e.data.used);
+        this.memData.xData.push(this.memTime++);
+      });
+      s.network.forEach(e => {
+        const netTotal: ScaledNetworkStat = new ScaledNetworkStat(e.data.received.raw + e.data.sent.raw);
+        this.netVal = round(netTotal.used, 1);
+        this.netUnits = netTotal.units;
+        this.netData.xData.push(e.timestamp);
+        this.netData.yData[0].push(e.data.sent.raw);
+        this.netData.yData[1].push(e.data.received.raw);
+      });
+    }));
   }
 
   public getChartMaxElements(): number {
--- a/src/app/space/create/deployments/services/deployments.service.ts
+++ b/src/app/space/create/deployments/services/deployments.service.ts
@@ -71,6 +71,10 @@ export interface TimeseriesResponse {
   data: DeploymentStats;
 }
 
+export interface MultiTimeseriesResponse {
+  data: MultiTimeseriesData;
+}
+
 export interface Application {
   attributes: ApplicationAttributes;
   id: string;
@@ -142,6 +146,36 @@ export interface TimeseriesData {
   memory: MemorySeries;
   net_tx: NetworkSentSeries;
   net_rx: NetworkReceivedSeries;
+}
+
+export interface MultiTimeseriesData {
+  cores: CoresSeries[];
+  memory: MemorySeries[];
+  net_tx: NetworkSentSeries[];
+  net_rx: NetworkReceivedSeries[];
+  start: number;
+  end: number;
+}
+
+export interface TimestampedCpuStats {
+  data: CpuStat;
+  timestamp: number;
+}
+
+export interface TimestampedMemoryStats {
+  data: ScaledMemoryStat;
+  timestamp: number;
+}
+
+export interface TimestampedNetworkStats {
+  data: { sent: ScaledNetworkStat, received: ScaledNetworkStat };
+  timestamp: number;
+}
+
+export interface TimeConstrainedStats {
+  cpu: TimestampedCpuStats[];
+  memory: TimestampedMemoryStats[];
+  network: TimestampedNetworkStats[];
 }
 
 export interface CoresSeries extends SeriesData { }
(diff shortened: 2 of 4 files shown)


fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2504-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2504-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/apps/deployment-details.component.ts
```diff
@@ -126,8 +132,8 @@ export class DeploymentDetailsComponent {
           this.netVal = round(netTotal.used, 1);
           this.netUnits = netTotal.units;
           this.netData.xData.push(+new Date());
-          this.netData.yData[0].push(stat.sent.raw);
-          this.netData.yData[1].push(stat.received.raw);
+          this.netData.yData[0].push(stat.sent.used);
+          this.netData.yData[1].push(stat.received.used);
```
andrewazores: IIRC I used `.raw` here rather than `used` to deal with the frequent changes in units and scaling that can occur with network traffic, ex. traffic on the order of bytes in one sample and on the order of kilo or megabytes in the next. Using `.used`, 800 bytes looks hundreds of times larger than 1KB.

src/app/space/create/deployments/apps/deployment-details.component.ts
```diff
@@ -155,6 +161,33 @@ export class DeploymentDetailsComponent {
     }
   }
 
+  private processTimeConstrainedStats(stats: Observable<TimeConstrainedStats>): void {
+    let subscription = stats.subscribe(s => {
```
andrewazores: This subscription needs to be unsubscribed on component destroy.

src/app/space/create/deployments/apps/deployment-details.component.ts
```diff
@@ -155,6 +161,33 @@ export class DeploymentDetailsComponent {
...
+        this.memData.yData.push(e.data.used);
+        this.memData.xData.push(this.memTime++);
+      });
+      s.network.forEach(e => {
+        const netTotal: ScaledNetworkStat = new ScaledNetworkStat(e.data.received.raw + e.data.sent.raw);
+        this.netVal = round(netTotal.used, 1);
+        this.netUnits = netTotal.units;
+        this.netData.xData.push(e.timestamp);
+        this.netData.yData[0].push(e.data.sent.used);
+        this.netData.yData[1].push(e.data.received.used);
```
andrewazores: Same here re. raw vs used.

src/app/space/create/deployments/services/deployments.service.ts
```diff
@@ -99,6 +103,21 @@ export interface TimeseriesData {
...
+  cores: CoresSeries[];
+  memory: MemorySeries[];
+  net_tx: NetworkSentSeries[];
+  net_rx: NetworkReceivedSeries[];
+  start: number;
+  end: number;
+}
+
+export interface TimeConstrainedStats {
+  cpu: { data: { used: number, quota: number }, timestamp: number }[];
```
andrewazores: `{ used: number, quota: number }` is a duplicate of the Stat interface, so this could be written `cpu: { data: CpuStat, timestamp: number }[]`.

andrewazores: The whole array type here could also be extracted out into another interface so it's written as `cpu: TimeConstrainedCpuStats` for example, which would save some typing later on too.

src/app/space/create/deployments/services/deployments.service.ts
```diff
@@ -259,6 +278,47 @@ export class DeploymentsService implements OnDestroy {
...
+      for (let i = 0; i < n[0].length; i++) {
+        arr.push({ data: { sent: new ScaledNetworkStat(n[0][i].value), received: new ScaledNetworkStat(n[1][i].value) }, timestamp: n[0][i].time });
+      }
+      return arr;
+    });
+    return Observable.combineLatest(cpuStats, memStats, networkStats, (
+        cpuStats: { data: { used: number, quota: number }, timestamp: number }[],
+        memStats: { data: ScaledMemoryStat, timestamp: number }[],
+        networkStats: { data: { sent: ScaledNetworkStat, received: ScaledNetworkStat }, timestamp: number }[]
+      ) => ({ cpu: cpuStats, memory: memStats, network: networkStats })
```
andrewazores: Here's where `TimeConstrainedCpuStats` would be a nice type to reference for these lambda params.

src/app/space/create/deployments/services/deployments.service.ts
```diff
@@ -259,6 +278,47 @@ export class DeploymentsService implements OnDestroy {
       );
   }
 
+  getDeploymentTimeConstrainedStats(spaceId: string, applicationId: string, environmentName: string, elapsedTime: number, endTime: number = Date.now()): Observable<TimeConstrainedStats> {
+    const startTime = endTime - elapsedTime;
+    const stats = this.getTimeConstrainedTimeseriesData(spaceId, applicationId, environmentName, startTime, endTime);
+    // Process CPU Stats
+    let cpuStats: Observable<CoresSeries[]> = stats.map((t: MultiTimeseriesData) => t.cores);
+    const cpuQuota = this.getEnvironmentCpuStat(spaceId, environmentName).map((stat: CpuStat) => stat.quota);
+    cpuStats = Observable.combineLatest(cpuStats, cpuQuota, (cpuStats: CoresSeries[], cpuQuota: number) => {
```
andrewazores: I'd prefer to avoid reassigning cpuStats and memStats like this. Better to have the initial definition be a different name and make them all const - just easier to follow IMO.

src/app/space/create/deployments/services/deployments.service.ts
```diff
@@ -389,6 +449,28 @@ export class DeploymentsService implements OnDestroy {
...
+      start: startTime,
+      end: endTime
+    };
+    return this.isApplicationDeployedInEnvironment(spaceId, applicationId, environmentName)
+    .flatMap((deployed: boolean) => {
+      if (deployed) {
+        const url = `${this.apiUrl}${spaceId}/applications/${applicationId}/deployments/${environmentName}/statseries?start=${startTime}&end=${endTime}`;
+        return this.http.get(url, { headers: this.headers })
+          .map((response: Response) => (response.json() as MultiTimeseriesResponse).data)
+          .catch(() => Observable.of(emptyResult));
```
andrewazores: Is there a particular reason for this emptyResult behaviour for this request, or is it just because it's been rebased since before handleHttpError was introduced? The other HTTP requests in this service now handle the case of a failed request by emitting an error event and logging the error, rather than returning an empty default value.

andrewazores: This is a relatively larger change so following up in a separate commit to bring the behaviour in line is fine.

andrewazores:
Also, please amend the commit message to contain reference to the OSIO issue that this targets.

src/app/space/create/deployments/services/deployments.service.ts
aptmac: Yeah I was working on a branch that was before handleHttpError was introduced, so I updated it in a second commit.

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2504-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2504-fabric8-ui.badger.fabric8.io

joshuawilson:
please rebase

src/app/space/create/deployments/services/deployments.service.ts
```diff
@@ -259,6 +293,47 @@ export class DeploymentsService implements OnDestroy {
...
 
+  getDeploymentTimeConstrainedStats(spaceId: string, applicationId: string, environmentName: string, elapsedTime: number, endTime: number = Date.now()): Observable<TimeConstrainedStats> {
+    const startTime = endTime - elapsedTime;
+    const series = this.getTimeConstrainedTimeseriesData(spaceId, applicationId, environmentName, startTime, endTime);
+    // Process CPU Stats
+    const coresSeries: Observable<CoresSeries[]> = series.map((t: MultiTimeseriesData) => t.cores);
+    const cpuQuota = this.getEnvironmentCpuStat(spaceId, environmentName).map((stat: CpuStat) => stat.quota);
+    let cpuStats = Observable.combineLatest(coresSeries, cpuQuota, (coresSeries: CoresSeries[], cpuQuota: number) => {
+      let arr = [];
+      for (let i = 0; i < coresSeries.length; i++) {
```
joshuawilson: Why are you looping over an Observable? Usually you Subscribe to them.

andrewazores: He's looping over an array emitted by an Observable, but it is a little confusing because the name is shadowed.

aptmac:
Hm, I'll be working on getting the tests to pass again, also realized I somehow duplicated my tests when I rebased, so I'll take care of that too.

src/app/space/create/deployments/services/deployments.service.ts
joshuawilson: it appears to a variable with the type of Observable.

aptmac: I'm trying to combine the cores series observable and quota observable similar to how it is done in [getDeploymentCpuStat](https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/space/create/deployments/services/deployments.service.ts#L282) [2]. In that instance there is only one cores series and one quota, so it's a quick combination and return. With the response that /statseries returns we get an array of coreseries, and combining each index with the quota isn't as straightforward? The loop accomplishes what [2] does for however many data points the api response returns (in this case, potentially 15 points). Since this cpuStats observable gets combined at the end of the function with the memory and network observable, does it make sense to manipulate the data while it's still an observable? Or should it be subscribed to for formatting, and then combined as an observable for the function return value?

[2] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/space/create/deployments/services/deployments.service.ts#L282

andrewazores: @joshuawilson it's not - there are two variables and the array one shadows the Observable one.

`const coresSeries: Observable<CoresSeries[]> = series.map((t: MultiTimeseriesData) => t.cores);` is the first, outer declaration, but the second usage that is iterated over as an array comes from the lambda parameter: `let cpuStats = Observable.combineLatest(coresSeries, cpuQuota, (*coresSeries*: CoresSeries[], cpuQuota: number) => {`

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2504-7 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2504-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/services/deployments.service.ts
joshuawilson: thanks for pointing it out

PR #2510, opened 2018-02-07
fix(header): fix analyze tab not highlighting when selected

aptmac:
This PR addresses https://github.com/openshiftio/openshift.io/issues/1708

This issue was caused by the analyze tab not being given an active state then selected. The analyze tab has no children, and should be activated (when selected) by [line 235 of the header component](https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/layout/header/header.component.ts#L235)[0]. However, the condition for checking if sub-menu children exist incorrectly passes because subMenus is given the value of [] at [line 193](https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/layout/header/header.component.ts#L193)[1].

[0] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/layout/header/header.component.ts#L235
[1] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/layout/header/header.component.ts#L193

### DIFF
--- a/src/app/layout/header/header.component.ts
+++ b/src/app/layout/header/header.component.ts
@@ -177,6 +177,13 @@ export class HeaderComponent implements OnInit, OnDestroy {
     return (this.router.url.indexOf('_gettingstarted') !== -1);
   }
 
+  private stripQueryFromUrl(url: string) {
+    if (url.indexOf('?q=') !== -1) {
+      url = url.substring(0, url.indexOf('?q='));
+    }
+    return url;
+  }
+
   private updateMenus() {
     if (this.context && this.context.type && this.context.type.hasOwnProperty('menus')) {
       let foundPath = false;
@@ -191,7 +198,7 @@ export class HeaderComponent implements OnInit, OnDestroy {
         // /namespace/space/create instead of /namespace/space/create/pipelines
         // as the 'Create' page matches to the 'Codebases' page
         let subMenus = (n.menus || []).slice().reverse();
-        if (subMenus) {
+        if (subMenus && subMenus.length > 0) {
           for (let o of subMenus) {
             // Clear the menu's active state
             o.active = false;
@@ -231,7 +238,7 @@ export class HeaderComponent implements OnInit, OnDestroy {
               }
             }
           }
-        } else if (!foundPath && n.fullPath === this.router.url) {
+        } else if (!foundPath && n.fullPath === this.stripQueryFromUrl(this.router.url)) {
           n.active = true;
           foundPath = true;
         }


aptmac:
Hmmm, I'm looking into why the tests fail with with patch applied.

src/app/layout/header/header.component.ts
```diff
@@ -191,7 +191,7 @@ export class HeaderComponent implements OnInit, OnDestroy {
         // /namespace/space/create instead of /namespace/space/create/pipelines
         // as the 'Create' page matches to the 'Codebases' page
         let subMenus = (n.menus || []).slice().reverse();
-        if (subMenus) {
+        if (subMenus.length > 0) {
```
jiekang: I'd prefer a null check beforehand as well, just to be safe.

E.g. `subMenus && subMenus.length > 0`

aptmac: Sure, I've added that in.

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2510-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2510-fabric8-ui.badger.fabric8.io

joshuawilson:
This fix highlights the tab when you first visit it but if you go to another tab and then back it is not highlighted.

aptmac:
It should work correctly if you go to the 'Create' tab and back to 'Analyze', but both the Plan and Settings tabs have (2) larger bugs that interfere with this.

1. When going to the Plan tab, a query gets added to the end of the URL, and what seems to be another router takes priority over the current one - so when going from 'Plan' to 'Analyze', the code used for highlighting the tab gets triggered twice. The first time, updateMenus() gets called with the correct URL, and it now correctly highlights the Analyze tab. Immediately afterwards, the updateMenus() code is called again with the queried URL, which will fail the condition [at this line](https://github.com/aptmac/fabric8-ui/blob/ea4488725974e135918a473c6117c5e3179c160c/src/app/layout/header/header.component.ts#L234)[2]. If you go from Plan to Analyze with this patch applied, it should appear that the Analyze tab highlights for about a second, and then fades because of this behaviour.

2. The settings tab: when I visit the settings tab I cannot visit any other tab, my page stays stuck on an empty '/settings' page.

[2] https://github.com/aptmac/fabric8-ui/blob/ea4488725974e135918a473c6117c5e3179c160c/src/app/layout/header/header.component.ts#L234

aptmac:
The condition isn't passing because the base url for Analyze may be compared with a url that has a query attached to the end (as a result of visiting the plan page). I've introduced a short function that strips the query for the purposes of url equivalence.

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2510-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2510-fabric8-ui.badger.fabric8.io

PR #2534, opened 2018-02-13
fix(workspaces): add support for old workspaces URL

joshuawilson:


joshuawilson:
@ibuziuk this is more along the lines of what we needed I think. I was not paying attention when I merged your PR.

fabric8cd:
@joshuawilson fabric8/fabric8-ui:SNAPSHOT-PR-2534-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2534-fabric8-ui.badger.fabric8.io

src/app/space/create/codebases/services/workspaces.service.ts
```diff
@@ -65,7 +67,24 @@ export class WorkspacesService {
...
       })
       .catch((error) => {
-        return this.handleError(error);
+        if (error.status === 404) {
+          return this.http
+            .get(old_url, {headers: this.headers})
+            .retryWhen(attempts => {
+              let count = 0;
+              return attempts.flatMap(error => {
+                return ++count >= 10 ? Observable.throw(error) : Observable.timer(count * 3000); // Wait for Che to start
```
corinnekrych: @joshuawilson We retry 10 times with a 3s pause. 
@ibuziuk is there an endpoint for `status`? so we wouldn't have to do those long retries.

ibuziuk: @corinnekrych yeah, here is the state endpoint[1]. Also, imo there is no need to retry if the error is 404 (When che server is not ready status code would be 503 / 504 ). Basically, if we got 404 than we should try new endpoint without checking state


[1] https://github.com/fabric8-services/fabric8-wit/blob/master/design/codebases.go#L214

joshuawilson:
@ibuziuk can you take the change and make a new PR and put in the changes you want?

ibuziuk:
:-) ok, but let's than just revert the commit with changing links. I will add task for changing link for the next sprint

joshuawilson:
This PR handles both. We should either merge this or just fix it. I prefer to move forward over reverting.

ibuziuk:
> This PR handles both

Unfortunately, this PR handles both incorrectly. While testing it locally the list of workspaces is never returned, since old endpoint (`/edit`) is never get called. Basically, it is pretty weird to see this `retryWhen` (I guess it was back to the days with single tenant che-server which was idled and there were no possibility to get the state properly so UI retried again and again till get valid response from wit) 
However, those checks are not relevant anymore since `che/state` endpoint is called currently before executing any operations with workspaces [1] and  `retryWhen` while obtaining the list of workspaces are simply irrelevant.

[1] https://github.com/fabric8-ui/fabric8-ui/blob/c8ab9f4d976e9fd3acac3334a1e1e6c111886b94/src/app/space/create/codebases/codebases.component.ts#L97

joshuawilson:
defer to #2536

fabric8cd:
@joshuawilson fabric8/fabric8-ui:SNAPSHOT-PR-2534-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2534-fabric8-ui.badger.fabric8.io

PR #2539, opened 2018-02-14
fix(deployments): wait for preload request completion

andrewazores:
Allow 15 minute data preload request to complete before subscribing for
per-minute data polling updates

fixes https://github.com/openshiftio/openshift.io/issues/2267

The fix here uses an Observable that fires a completion event after the initial stats response is received, and only subscribing for the polling updates strictly after that completion event occurs. This guarantees the ordering so that the 15 minute preload data is received and processed before a polling update response can be received. The tests take advantage of this as well, using a plain Subject rather than a BehaviorSubject so that the event emission can be controlled and the polling updates stalled for the purposes of the preloading tests.

### DIFF
--- a/src/app/space/create/deployments/apps/deployment-details.component.spec.ts
+++ b/src/app/space/create/deployments/apps/deployment-details.component.spec.ts
@@ -9,7 +9,8 @@ import { CollapseModule } from 'ngx-bootstrap/collapse';
 import 'patternfly/dist/js/patternfly-settings.js';
 import {
   BehaviorSubject,
-  Observable
+  Observable,
+  Subject
 } from 'rxjs';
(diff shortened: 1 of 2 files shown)


fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2539-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2539-fabric8-ui.badger.fabric8.io

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2539-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2539-fabric8-ui.badger.fabric8.io

PR #2545, opened 2018-02-15
fix(overview): add word-wrap functionality to bio text

aptmac:
This PR addresses https://github.com/openshiftio/openshift.io/issues/2100 [0].

This PR adds some CSS properties to the bio section of the profile page. The word-wrap is used to make text word-wrap, and the hyphens property introduces hyphens when word-wrapping occurs. However, the hyphens property is not currently supported by all browsers [1], and only the 'none' option is supported by Chrome [2]. This results in word-wrapping and hyphenating in many browsers (IE, Firefox, Edge, Safari, etc.), but Chrome will only have word-wrapping. Additionally, for hyphenation to work properly, the html component it applies to has to have the lang option set. I've included some screenshots below to demo the functionality.

Here is what the bio looks like currently:
old

Here is what the bio looks like with these fixes in Firefox:
new-firefox

Here is what the bio looks like with these fixes in Chrome:
new-chrome

[0] https://github.com/openshiftio/openshift.io/issues/2100
[1] https://caniuse.com/#search=hyphens
[2] https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens

### DIFF
--- a/src/app/profile/overview/overview.component.html
+++ b/src/app/profile/overview/overview.component.html
@@ -52,7 +52,7 @@
         </ng-template>
         <div *ngIf="(context.user.attributes.bio !== undefined && context.user.attributes.bio.length !== 0); then showBio else addBio"></div>
         <ng-template #showBio>
-          <p class="col-sm-12 profile-wrapper-bio">{{context.user.attributes.bio}}</p>
+          <p class="col-sm-12 profile-wrapper-bio" lang="en">{{context.user.attributes.bio}}</p>
         </ng-template>
         <ng-template #addBio>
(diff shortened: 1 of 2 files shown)


fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2545-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2545-fabric8-ui.badger.fabric8.io

mindreeper2420:
LGTM - tested using Firefox and Chrome on Mac & Fedora.

PR #2546, opened 2018-02-15
fix(docs): Update method for visiting documentation

mindreeper2420:
Update how users access docs.openshift.io. Documentation link was removed from the user menu into the main navigation as an 'infotip' dropdown. This now includes a direct link to docs.openshift.io, as well as the About modal.

related to https://github.com/openshiftio/openshift.io/issues/2215

screen shot 2018-02-15 at 3 44 59 pm
screen shot 2018-02-15 at 3 45 05 pm


catrobson:
@mindreeper2420 the icon actually shouldn't get a "select" state with the gray background on hover or click.  Only the status gets a background if there is a status of error or warning.  There is mixed direction in PatternFly on this, so I'm not sure what you're working from based on the PatternFly CSS, so if this doesn't make sense based on PatternFly CSS let's chat quickly about it with Kyle.

mindreeper2420:
@catrobson Ok I've updated to remove the background color when the the icon is hovered over or active. This is a custom style that is applied to the dropdown menus, to meet the visual design requirements.

mindreeper2420:
screen shot 2018-02-15 at 4 53 09 pm

catrobson:
Ok, thanks.  We'll resolve in a separate thread at the PatternFly level and then can clean up if needed at that point.

PR #2548, opened 2018-02-16
fix(deployments): fix resource usage collapsed icon

aptmac:
This short PR adjusts the fa-angle icon to change dynamically based on the "resource usage" section of the page being expanded or collapsed.

Currently, the icon is always set to fa-angle-down, regardless of the resources section being collapsed or expanded. Here, an `ngClass` can be introduced to change the icon based on the status of the `resourcesCollapsed` variable that gets toggled when the user expands/collapses the section.

What it looks like currently:
old-collapsed

What it looks like after this patch:
now-collapsed


fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2548-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2548-fabric8-ui.badger.fabric8.io

PR #2551, opened 2018-02-16
Relock create area's create button and move event handler code

chrislessard:
This PR accomplishes two goals : It ensures that when the Create Area modal is re-opened (i.e., when creating a second area) that the create button is disabled by default until user input is provided [0]. It also moves all event handler code to the create-area-dialog component, rather than having some there and the rest in the areas component. 

Solves: 
[0] https://github.com/openshiftio/openshift.io/issues/2252

There is discussion in [0] about the desired behaviour regarding which area should be selected on the modal open event. This implementation follows what @catrobson describes in that issue.

### DIFF
--- a/src/app/space/settings/areas/areas.component.html
+++ b/src/app/space/settings/areas/areas.component.html
@@ -38,7 +38,7 @@
   </div>
 </div>
 
-<div class="modal fade" bsModal #modal="bs-modal" tabindex="-1" role="dialog" aria-hidden="true" (onShown)="onShowHandler()" (onHide)="onHideHandler()">
+<div class="modal fade" bsModal #modal="bs-modal" tabindex="-1" role="dialog" aria-hidden="true" (onShown)="createAreaDialog.onOpen()" (onHide)="createAreaDialog.onClose()">
   <div class="modal-dialog">
     <div class="modal-content">
       <create-area-dialog #createAreaDialog [host]="modal" [parentId]="selectedAreaId" [areas]="areas" (onAdded)="addArea($event)"></create-area-dialog>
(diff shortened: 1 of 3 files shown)


fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2551-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2551-fabric8-ui.badger.fabric8.io

andrewazores:
The name field and submit button reset seems to work as expected, and the code looks fine.

For the parent space dropdown however, I don't think the behaviour with this changeset matches what Cat described (or maybe I'm just misinterpreting). If I use the kebab menu to add a child space then the dropdown correctly pre-selects the space, so that it matches with the entry whose kebab I used. If I use the top right "Add Areas" button/link, it initially defaults to the root space name, but will reflect the most recently selected space name afterward. Repro steps:

1. Visit space settings for ex. test-space
2. Add a space test-space/foo
3. Click the "Add Areas" button - modal dropdown has parent area "test-space"
4. Click the kebab/"Add Child Area" for test-space - modal dropdown has parent area test-space
5. Click the kebab/"Add Child Area" for test-space/foo - modal dropdown has parent area test-space/foo
6. Click the "Add Areas" button - modal dropdown has parent area test-space/foo (should be test-space)

chrislessard:
@andrewazores good catch - looks like a line was missing in the createArea function. I've pushed in an additional commit to deal with this.

fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2551-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2551-fabric8-ui.badger.fabric8.io

fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2551-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2551-fabric8-ui.badger.fabric8.io

PR #2553, opened 2018-02-16
feat(deployments): add time and color indicators to deployments chart…

aptmac:
… tooltips

This PR adds the newly propagated timestamp values from the DeploymentsService to the chart tooltips. This is an attempt to make the charts look more closely related to the OSO counter-parts. The CSS values were taken from inspecting the OSO console. The tooltips for the CPU and Memory charts have also been updated in a similar fashion.

Here's an example of what the openshift console network chart tooltips look like:
osio-network

Here's what the deployments network chart tooltips look like currently:
old-network

Here's what the deployments network chart tooltips look like with this PR:
new-network

### DIFF
--- a/src/app/space/create/deployments/apps/deployment-details.component.ts
+++ b/src/app/space/create/deployments/apps/deployment-details.component.ts
@@ -53,13 +53,13 @@ export class DeploymentDetailsComponent {
   cpuData: SparklineData = {
     dataAvailable: true,
     xData: ['time'],
-    yData: ['used']
+    yData: ['CPU']
   };
 
   memData: SparklineData = {
     dataAvailable: true,
     xData: ['time'],
-    yData: ['used']
+    yData: ['Memory']
   };
 
   netData: DeploymentsLinechartData = {
@@ -75,7 +75,9 @@ export class DeploymentDetailsComponent {
     chartId: uniqueId('cpu-chart'),
     axis: {
       type: 'timeseries'
-    }
+    },
+    tooltip: this.getTooltipContents(),
+    units: 'Cores'
   };
 
   memConfig: SparklineConfig = {
@@ -83,7 +85,8 @@ export class DeploymentDetailsComponent {
     chartId: uniqueId('mem-chart'),
     axis: {
       type: 'timeseries'
-    }
+    },
+    tooltip: this.getTooltipContents()
   };
 
   netConfig: DeploymentsLinechartConfig = {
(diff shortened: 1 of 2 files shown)


fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2553-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2553-fabric8-ui.badger.fabric8.io

mindreeper2420:
instead of replicating this bit
```
.c3-tooltip td>span {
  display: inline-block;
  width: 10px;
  height: 10px;
  margin-right: 6px;
}

.c3-tooltip td.value {
  text-align: left;
}
```
3 times, would it be cleaner to move it to a shared stylesheet that targets the c3 items?

aptmac:
@mindreeper2420 Yes that would have been much cleaner. After some digging and looking at the OSO Console tooltip CSS, I found that the required CSS for the tooltip is already present, but I had a typo in the `<td class="` I wrote in the component files, where I duplicated the 'value' class instead of having a 'name'. Now there are no modifications to the stylesheets.

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2553-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2553-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/apps/deployment-details.component.ts
```diff
@@ -241,4 +244,40 @@ export class DeploymentDetailsComponent {
...
+        let color = '#0088ce'; // pf-blue-400
+        let units: string = '';
+        if (d[0].name === 'CPU') {
+          units = this.cpuConfig.units;
+        } else if (d[0].name === 'Memory') {
+          units = this.memUnits;
+        }
+        for (let i = 0; i < d.length; i++) {
+          if (i === 0) {
+            tipRows += '<tr><th colspan="2">' + d[i].x.toLocaleString() + '</th></tr>';
```
jiekang: Can you add a comment explaining this special case for the first element? Is there an assumption on what `d` contains?

aptmac: From a bit of digging, `d` is the tooltip data [0]. I had modified this function from the default sparkline tooltip function in pf-ng [1], and fwiw, it looks like it was based on how c3 refers to it in their code as well [2]. 

For an example, I put a println into `getTooltipContents()` and got this for one of the CPU chart `d`: [{x: Wed Feb 21 2018 09:31:24 GMT-0500 (EST), value: 0, id: "CPU", index: 3, name: "CPU"}]. 

[0] http://c3js.org/reference.html#tooltip-contents
[1] https://github.com/patternfly/patternfly-ng/blob/master/src/app/chart/sparkline/sparkline.component.ts#L157
[2] https://github.com/c3js/c3/blob/master/c3.js#L9016

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2553-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2553-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/apps/deployment-details.component.ts
```diff
@@ -241,4 +244,39 @@ export class DeploymentDetailsComponent {
...
+        '  <td class="name"><span style="background-color:' + color + '"></span>' + d[0].name + '</td>' +
+        '  <td class="value text-nowrap">' + d[0].value + ' ' + units + '</td>' +
+        '</tr>';
+        return this.getTooltipTableHTML(tipRows);
+      }
+    };
+  }
+
+  private getTooltipTableHTML(tipRows: string): string {
+    return '<div class="module-triangle-bottom">' +
```
joshuawilson: why not use the back tick for this, it would be much cleaner and easier to maintain

aptmac: Sure that sounds good, I've updated this to use back ticks instead.

aptmac:
I've updated the strings to use back ticks as per @joshuawilson's comment, let me know if the indentation is okay.

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2553-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2553-fabric8-ui.badger.fabric8.io

PR #2560, opened 2018-02-20
Add error highlighting functionality

chrislessard:
In order, the accompanying commits to this PR achieve three things in the create area modal:

1) Introduce a custom validator directive to the modal, which checks if a duplicate name has been entered. This saves us from the current wasteful implementation whereby a request must be sent, receiving an 409, before an error can be displayed informing the user that they must enter a unique name. 

2) Adds correct error highlighting (in red) as well as a helpful message underneath any infringing fields, as by the [patternfly specification](http://www.patternfly.org/pattern-library/forms-and-controls/errors-and-validation).

3) Adds additional error messages for the situation where a user types a name into the input, and then removes all inputted text, leaving the field blank. 

Solves : https://github.com/openshiftio/openshift.io/issues/2255

Screenshots: 

2018-02-20-152428_1903x932_scrot

2018-02-20-152411_1917x926_scrot

### DIFF
--- a/src/app/space/settings/areas/create-area-dialog/create-area-dialog.component.html
+++ b/src/app/space/settings/areas/create-area-dialog/create-area-dialog.component.html
@@ -7,15 +7,16 @@
   <form role="form" #areaForm="ngForm" (ngSubmit)="createArea()" novalidate>
     <div class="form">
       <fieldset class="create-fieldset">
-        <div *ngIf="errors" class="alert alert-danger">
-          <div [hidden]="!errors.uniqueValidationFailure">
+        <div [hidden]="!errors.uniqueValidationFailure" class="alert alert-danger">
+          <div>
             <span class="pficon pficon-error-circle-o"></span>
             <b>Unique Name</b> A unique name is required to add a new area.
           </div>
         </div>
-        <div class="form-group">
+        <div class="form-group" [ngClass]="{'has-error': errors.uniqueValidationFailure}">
           <label for="name" class="control-label">Name</label>
-          <input #nameInput id="name" name="name" type="text" class="form-control" placeholder="Enter a name for your area" [(ngModel)]="name" required />
+          <input #rawInputField id="name" name="name" type="text" class="form-control" placeholder="Enter a name for your area" [(ngModel)]="name" required />
+          <small [hidden]="!errors.uniqueValidationFailure">Please choose a unique name.</small>
         </div>
         <div class="form-group">
           <label class="control-label">Parent area</label>
--- a/src/app/space/settings/areas/create-area-dialog/create-area-dialog.component.html
+++ b/src/app/space/settings/areas/create-area-dialog/create-area-dialog.component.html
@@ -8,15 +8,18 @@
     <div class="form">
       <fieldset class="create-fieldset">
         <div [hidden]="!errors.uniqueValidationFailure" class="alert alert-danger">
-          <div>
-            <span class="pficon pficon-error-circle-o"></span>
-            <b>Unique Name</b> A unique name is required to add a new area.
-          </div>
+          <span class="pficon pficon-error-circle-o"></span>
+          <b>Unique Name</b> A unique name is required to add a new area.
         </div>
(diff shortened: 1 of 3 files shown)


jiekang:
@catrobson 

The screenshots show duplicate error information in the top banner and the message underneath the field. Are both required? I feel like having either or is sufficient. Both feels wrong to me. I prefer the message underneath as it's very clear what field I need to fix.

fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2560-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2560-fabric8-ui.badger.fabric8.io

catrobson:
@jiekang Yeah, I hear you.  What is implemented follows the PatternFly standard.  Unfortunately the PatternFly standard doesn't really get down to the details of the difference of very small, simple forms like this vs. long, complex forms.  Long, scrolling forms have real value for the top error message as well as the field level messages so you can show there is an overall problem and then the user can scroll through each of the red errors on the page.  In this cause, because you can see them both and its so obvious, it feels like overkill.

For consistency sake with other products and forms we do across the product, I would suggest we keep both even though it does feel like too much in this situation.

jiekang:
Okay, that makes sense. Thanks!

chrislessard:
@jiekang 

I agree with your point that the current implementation isn't wasteful. I left it in the first commit mainly because someone might be able to bypass the client side validation and force the request. 

With that said, I don't think client side validation through a directive is a bad idea. Its the commonly accepted way of achieving validation in angular and, because of that, makes for a more consistent user experience. If we drop the first commit and leave the later ones, the UX would get a bit funny : the "Create" button would disable in the case of a "no name provided" error, and would immediately show an error message. The button would not disable in the event of a "duplicate name" error, and an error message would only appear after the request was made. In addition to making for an inconsistent experience, this approach becomes more difficult to work with (in terms of setting and clearing the errors) in the html.

What do you think ?

fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2560-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2560-fabric8-ui.badger.fabric8.io

jiekang:
@chrislessard 

Client side validation is valid for situations where you don't need backend information to make the decision. In this case, checking whether an area exists with the same name does require server side validation. Your client side validation is also checking against `areas` which actually comes from the backend anyways.

I think it's common for forms to have client side validation as you type for static validators like: minimum/maximum length of input and invalid characters, followed by server side validation on submit for things like, unique account name. Creating an account on Gmail or Facebook, or most services with an account system will not notify you that the name you selected already exists until you either submit the form, or press some button to validate it for existence. This should be fine for this dialog as well.

I don't see where there is any change in setting and clearing errors either way. Can you elaborate?

chrislessard:
@jiekang 

If we use the current implementation, we rely on the errors array in the component code to determine whether an error message should be displayed, etc. The underlying code also assumes that 409s are unique validation errors, I'm not sure whether than matters or not. This means we need to manually manipulate the underline ngForm's valid attribute, which makes it much more convenient to disable the button and make sure that state doesn't persist when the model is reopened. The current impl relies on areaForm.form.valid to disable and enable the create button, using directives has angular do all that toggling for us. IMO its cleaner, but this isn't really an issue worth fussing over. If you'd prefer it to be changed then I'll just change it, although I still think it would be weird if some errors disable the create button and other's don't, so we should make a decision regarding that.

fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2560-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2560-fabric8-ui.badger.fabric8.io

fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2560-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2560-fabric8-ui.badger.fabric8.io

PR #2561, opened 2018-02-22
Deployments service refactor

andrewazores:
This patch set relates to https://github.com/openshiftio/openshift.io/issues/2253 . The goal of these patches is to simplify the DeploymentsService interface by removing the distinction between historical and streaming series data - consumers should be able to simply ask for series data, and it's the service's responsibility to decide what that data is. In this case, the service will decide to include 15 minutes' worth of stats immediately on the stream, and then continue publishing data on the same stream according to the service's own polling cycle to update the consumer with live results. This "front loading" of 15 minutes' worth of data is done on-demand per <space, app, env> tuple so that multiple consumers do not trigger multiple front-loading requests, and the last 15 data points are also cached by the service so that late subscribers are unable to distinguish between results that were front loaded vs results that were streamed.

The next step after this to close issue 2253 is to move the data trimming logic out of the components and into the service, which effectively means that the service should simply be re-emitting all 15 of its cached data points on every poll cycle, rather than only the latest new data point. Consumers would then be able to simply replace their own local state with that provided by the service, rather than needing to maintain their own buffer of historical data and trim it to size.

### DIFF
--- a/src/app/dashboard-widgets/environment-widget/environment-widget.module.ts
+++ b/src/app/dashboard-widgets/environment-widget/environment-widget.module.ts
@@ -5,11 +5,11 @@ import { RouterModule } from '@angular/router';
 
 import { MomentModule } from 'angular2-moment';
 
+import { DeploymentsModule } from '../../space/create/deployments/deployments.module';
 import { EnvironmentWidgetComponent } from './environment-widget.component';
 
-
 @NgModule({
-  imports: [CommonModule, FormsModule, RouterModule, MomentModule ],
+  imports: [CommonModule, DeploymentsModule, FormsModule, RouterModule, MomentModule],
   declarations: [EnvironmentWidgetComponent],
   exports: [EnvironmentWidgetComponent]
 })
--- a/src/app/space/create/deployments/apps/deployment-details.component.spec.ts
+++ b/src/app/space/create/deployments/apps/deployment-details.component.spec.ts
@@ -25,8 +25,7 @@ import { Pods } from '../models/pods';
 import { ScaledNetworkStat } from '../models/scaled-network-stat';
 import {
   DeploymentsService,
-  NetworkStat,
-  TimeConstrainedStats
+  NetworkStat
 } from '../services/deployments.service';
 import { DeploymentDetailsComponent } from './deployment-details.component';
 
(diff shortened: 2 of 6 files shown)


fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2561-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2561-fabric8-ui.badger.fabric8.io

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2561-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2561-fabric8-ui.badger.fabric8.io

joshuawilson:
@dgutride please review

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2561-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2561-fabric8-ui.badger.fabric8.io

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2561-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2561-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/deployments.component.ts
```diff
@@ -4,14 +4,25 @@ import { Spaces } from 'ngx-fabric8-wit';
...
 @Component({
   encapsulation: ViewEncapsulation.None,
   selector: 'alm-apps',
   templateUrl: 'deployments.component.html',
   styleUrls: ['./deployments.component.less'],
-  providers: [DeploymentsService]
+  providers: [
+    DeploymentsService,
+    {
+      provide: TIMER_TOKEN,
```
dgutride: Why does this need to be provided outside of the DeploymentsService?  It seems a bit odd to inject a timer here instead of just creating it inside the deployments service - is there a best practice or some example you followed that you could link to explaining this possibly?

andrewazores: The timer was just created inside the service previously, but I have a commit here making it injectable instead so that the unit test suite can inject a different implementation. Namely, a plain Subject so that the unit tests can easily control the update cycle of the service, rather than an actual timer. This could probably also have been done with fakeAsync, but replacing the timer altogether seemed like it would lead to more robust and maintainable tests.

dgutride: Ok - makes sense - not sure how this pattern will scale as we go, it's a little confusing to expose this as a shared observable but let's give it a shot.

andrewazores: The other solution I had in mind was to make the observable protected and then have the tests subclass the service and override it that way, which also means the tests get to drop DI and just construct the service directly. That might be more clear to readers but was a more invasive change for the DeploymentsService, which is why I opted against it for this PR.

Alternatively the timer could just be a public assignable property and the tests simply reassign it, but that felt pretty wrong.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2561-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2561-fabric8-ui.badger.fabric8.io

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2561-6 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2561-fabric8-ui.badger.fabric8.io

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2561-7 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2561-fabric8-ui.badger.fabric8.io

joshuawilson:
please rebase

andrewazores:
@dgutride I needed to change the DI approach when rebasing here. Since the dashboard environment-widget also now depends on the DeploymentsService, the poll timer provider was moved from the DeploymentsComponent to DeploymentsModule. A DeploymentsModule import dependency also needed to be added to the EnvironmentWidgetModule. Otherwise nothing really changed with the rebase.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2561-8 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2561-fabric8-ui.badger.fabric8.io

PR #2579, opened 2018-02-27
fix(deployments): fix deployment card dropdown tests

jiekang:
This fixes the dropdown tests to provide actual Event objects. The event from the handler is now being used by the component and would result in TypeErrors when running the tests. The TypeErrors are no longer seen with this fix. Interestingly, this did not affect the tests ability to run and pass/fail.

This resolves https://github.com/openshiftio/openshift.io/issues/2386


fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2579-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2579-fabric8-ui.badger.fabric8.io

PR #2581, opened 2018-02-27
fix(space-wizard): make request to create space once

jiekang:
This fixes the observable used to create a space in the space-wizard. It now is used once and cleaned up via `first()`. As well, the code is updated to use the latest API from `SpaceServices` as the previous API was deprecated.

This also addresses [1], where logging out would trigger another attempt to create the space because of the long-living observable receiving an event emission when logging out (user changes).

[1] https://github.com/openshiftio/openshift.io/issues/2206

### DIFF
--- a/src/app/space/wizard/space-wizard.component.ts
+++ b/src/app/space/wizard/space-wizard.component.ts
@@ -1,7 +1,7 @@
 import { Component, EventEmitter, OnDestroy, OnInit, Output } from '@angular/core';
 import { Router } from '@angular/router';
 
-import { Notification, Notifications, NotificationType } from 'ngx-base';
+import { Logger, Notification, Notifications, NotificationType } from 'ngx-base';
 import { Context, SpaceNamePipe, SpaceService } from 'ngx-fabric8-wit';
 import { ProcessTemplate } from 'ngx-fabric8-wit';
 import { Space, SpaceAttributes } from 'ngx-fabric8-wit';
@@ -37,7 +37,8 @@ export class SpaceWizardComponent implements OnInit, OnDestroy {
     private spaceNamespaceService: SpaceNamespaceService,
     private spaceNamePipe: SpaceNamePipe,
     private spacesService: SpacesService,
-    private context: ContextService
+    private context: ContextService,
+    private logger: Logger
   ) {
     this.spaceTemplates = dummy.processTemplates;
     this.space = this.createTransientSpace();
@@ -54,7 +55,7 @@ export class SpaceWizardComponent implements OnInit, OnDestroy {
    */
   createSpace() {
     if (!this.userService.currentLoggedInUser && !this.userService.currentLoggedInUser.id) {
-      console.log('Error creating space, invalid user.', this.userService.currentLoggedInUser);
+      this.logger.error('Error creating space, invalid user.' + this.userService.currentLoggedInUser);
       this.notifications.message(<Notification> {
         message: `Failed to create "${this.space.name}". Invalid user: "${this.userService.currentLoggedInUser}"`,
         type: NotificationType.DANGER
@@ -86,7 +87,7 @@ export class SpaceWizardComponent implements OnInit, OnDestroy {
         this.finish();
       },
       err => {
-        console.log('Error creating space', err);
+        this.logger.error(err);
         this.notifications.message(<Notification> {
           message: `Failed to create "${this.space.name}"`,
           type: NotificationType.DANGER
--- a/src/app/space/wizard/space-wizard.component.ts
+++ b/src/app/space/wizard/space-wizard.component.ts
@@ -1,4 +1,11 @@
-import { Component, EventEmitter, OnDestroy, OnInit, Output } from '@angular/core';
+import {
+  Component,
+  ErrorHandler,
+  EventEmitter,
+  OnDestroy,
+  OnInit,
+  Output
+} from '@angular/core';
 import { Router } from '@angular/router';
 
 import { Logger, Notification, Notifications, NotificationType } from 'ngx-base';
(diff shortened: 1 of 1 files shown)


src/app/space/wizard/space-wizard.component.ts
```diff
@@ -53,39 +53,47 @@ export class SpaceWizardComponent implements OnInit, OnDestroy {
...
+        // Ignore any errors coming out here, we've logged and notified them earlier
+        .catch(err => Observable.of(createdSpace));
+    })
+    .subscribe(createdSpace => {
+        this.router.navigate([createdSpace.relationalData.creator.attributes.username,
+          createdSpace.attributes.name]);
+        this.finish();
+      },
+      err => {
+        console.log('Error creating space', err);
```
andrewazores: I realize this is just carried over, but since you're here perhaps `console.log` should be `console.error`. Or better yet, `logger.error`.

src/app/space/wizard/space-wizard.component.ts
```diff
@@ -53,39 +53,47 @@ export class SpaceWizardComponent implements OnInit, OnDestroy {
...
-          console.log('Error creating space', err);
-          this.notifications.message(<Notification> {
-            message: `Failed to create "${this.space.name}"`,
-            type: NotificationType.DANGER
-          });
-          this.finish();
+
+    this.space.relationships['owned-by'].data.id = this.userService.currentLoggedInUser.id;
+    this.spaceService.create(this.space)
+    .first()
```
andrewazores: Is it confirmed that the returned `Observable<Space>` from `spaceService.create` is a hot Observable? I don't know the implementation details but it seems like that's the kind of thing that would just emit once and complete - similar to what a `Promise` would've been used for. Our discussion about this before was on using `.first()` or a non-empty filter on the userService so that the logged-out empty user wouldn't trigger this code path a second time.

You can verify it easily enough by adding a `.do(noop, noop, () => console.log('subscription completed'))` to the `spaceService.create`, so that you get a browser console message when the Observable completes (if ever). If that Observable already completes after emitting a Space then there's no need for the `.first()` operator. If it doesn't complete, then that's pretty weird, but the `.first()` operator would be reasonable.

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2581-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2581-fabric8-ui.badger.fabric8.io

src/app/space/wizard/space-wizard.component.ts
jiekang: Yeah, looks like it does complete. Does this mean I don't need to unsubscribe?

src/app/space/wizard/space-wizard.component.ts
jiekang: Good point, fixed as well while I'm here.

src/app/space/wizard/space-wizard.component.ts
andrewazores: Yea, no need to explicitly unsubscribe here then. Nice.

src/app/space/wizard/space-wizard.component.ts
```diff
@@ -53,39 +53,47 @@ export class SpaceWizardComponent implements OnInit, OnDestroy {
    * by invoking the spaceService
    */
   createSpace() {
-    console.log('Creating space', this.space);
+    if (!this.userService.currentLoggedInUser && !this.userService.currentLoggedInUser.id) {
+      console.log('Error creating space, invalid user.', this.userService.currentLoggedInUser);
```
andrewazores: `logger.error` here too please.

jiekang: Right, my bad.

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2581-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2581-fabric8-ui.badger.fabric8.io

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2581-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2581-fabric8-ui.badger.fabric8.io

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2581-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2581-fabric8-ui.badger.fabric8.io

src/app/space/wizard/space-wizard.component.ts
```diff
@@ -53,39 +54,46 @@ export class SpaceWizardComponent implements OnInit, OnDestroy {
    * by invoking the spaceService
    */
   createSpace() {
-    console.log('Creating space', this.space);
+    if (!this.userService.currentLoggedInUser && !this.userService.currentLoggedInUser.id) {
+      this.logger.error('Error creating space, invalid user.' + this.userService.currentLoggedInUser);
```
joshuawilson: If you are creating an error you should also send it to the global error handler. That way it gets picked up by Sentry.

src/app/space/wizard/space-wizard.component.ts
```diff
@@ -53,39 +54,46 @@ export class SpaceWizardComponent implements OnInit, OnDestroy {
...
+        .catch(err => Observable.of(createdSpace));
+    })
+    .subscribe(createdSpace => {
+        this.router.navigate([createdSpace.relationalData.creator.attributes.username,
+          createdSpace.attributes.name]);
+        this.finish();
+      },
+      err => {
+        this.logger.error(err);
+        this.notifications.message(<Notification> {
```
joshuawilson: this should call the error handler

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2581-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2581-fabric8-ui.badger.fabric8.io

PR #2600, opened 2018-03-02
Deployments loading

chrislessard:
These patches introduce a loading display for the deployment page, as described by the design mock [here](https://redhat.invisionapp.com/share/YSDL0KRKW#/screens/273548936). 

Screenshot (to see this, I had to apply a set timeout to the call to updateResources in deployments.component.ts) 

2018-03-02-141634_1912x953_scrot

Solves : https://github.com/openshiftio/openshift.io/issues/1865

### DIFF
--- a/src/app/space/create/deployments/apps/deployments-apps.component.html
+++ b/src/app/space/create/deployments/apps/deployments-apps.component.html
@@ -1,32 +1,66 @@
 <deployments-toolbar (onFilterChange)="filterChange($event)" (onSortChange)="sortChange($event)" [resultsCount]="resultsCount">
 </deployments-toolbar>
 
-<div class="cards-pf">
-  <div class="row row-cards-pf">
-    <div class="col-xs-12 col-sm-4 col-md-2">
-      <h3>
-        <b>Applications</b>
-      </h3>
+<div *ngIf="(hasLoaded | async); else loading">
+  <div class="cards-pf">
+    <div class="row row-cards-pf">
+      <div class="col-xs-12 col-sm-4 col-md-2">
+        <h3>
+          <b>Applications</b>
+        </h3>
+      </div>
+      <div class="col-xs-12 col-sm-4 col-md-2">
+        <h3>
+          <b>Environments</b>
+        </h3>
+      </div>
     </div>
-    <div class="col-xs-12 col-sm-4 col-md-2">
-      <h3>
-        <b>Environments</b>
-      </h3>
+  </div>
+
+  <div class="cards-pf">
+    <div class="row row-cards-pf">
+      <div class="col-xs-12 col-sm-4 col-md-2"></div>
+      <div class="col-xs-12 col-sm-4 col-md-3" *ngFor="let environment of environments | async">
+        <h4 class="env-card-title">
+          {{environment.name}}
+        </h4>
+      </div>
     </div>
+  </div>
+
+  <div *ngFor="let application of filteredApplicationsList">
+    <deployment-card-container [spaceId]="spaceId | async" [application]="application" [environments]="environments"></deployment-card-container>
   </div>
 </div>
 
-<div class="cards-pf">
-  <div class="row row-cards-pf">
-    <div class="col-xs-12 col-sm-4 col-md-2"></div>
-    <div class="col-xs-12 col-sm-4 col-md-3" *ngFor="let environment of environments | async">
-      <h4 class="env-card-title">
-        {{environment.name}}
-      </h4>
+<ng-template #loading>
+  <div>
+    <div class="cards-pf">
+        <div class="row row-cards-pf">
+          <div>
+            <div class="spinner"></div>
+            <h3 class="loading-msg">Loading Deployments...</h3>
+          </div>
+        </div>
+      </div>
+
+    <div class="cards-pf">
+      <div class="row row-cards-pf">
+        <div class="spacer col-xs-12 col-sm-4 col-md-2">&nbsp;</div>
+        <div class="filler col-xs-12 col-sm-4 col-md-3">&nbsp;</div>
+        <div class="filler col-xs-12 col-sm-4 col-md-3">&nbsp;</div>
+      </div>
+    </div>
+
+    <div class="cards-pf">
+      <div *ngFor="let number of [0,1,2,3]">
+        <div class="row row-cards-pf">
+          <div class="filler col-xs-12 col-sm-4 col-md-2">&nbsp;</div>
+          <div class="filler col-xs-12 col-sm-4 col-md-3"><br><br><br></div>
+          <div class="filler col-xs-12 col-sm-4 col-md-3"><br><br><br></div>
+        </div>
+      </div>
+      <br>
     </div>
   </div>
-</div>
-
-<div *ngFor="let application of filteredApplicationsList">
-  <deployment-card-container [spaceId]="spaceId | async" [application]="application" [environments]="environments"></deployment-card-container>
-</div>
+</ng-template>
--- a/src/app/space/create/deployments/apps/deployments-apps.component.less
+++ b/src/app/space/create/deployments/apps/deployments-apps.component.less
@@ -1,0 +1,18 @@
+.filler {
+  background-color: white;
+  margin: 0px 10px 10px 10px;
+}
+
+.spacer {
+  margin: 0px 10px 10px 10px;
+}
+
+.loading-msg {
+  display: inline-block;
+  vertical-align: 5px;
+}
+
+.spinner {
+  margin: 0px 0px 0px 0px;
+  display: inline-block;
+}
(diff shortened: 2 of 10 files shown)


fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2600-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2600-fabric8-ui.badger.fabric8.io

joshuawilson:
@chrislessard please reword the commit message. https://github.com/fabric8-ui/fabric8-ui#type

fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2600-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2600-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/apps/deployments-apps.component.html
```diff
@@ -1,32 +1,82 @@
...
+        </div>
+      </div>
+
+    <div class="cards-pf">
+      <div class="row row-cards-pf">
+        <div class="spacer col-xs-12 col-sm-4 col-md-2">&nbsp;</div>
+        <div class="filler col-xs-12 col-sm-4 col-md-3">&nbsp;</div>
+        <div class="filler col-xs-12 col-sm-4 col-md-3">&nbsp;</div>
+      </div>
+    </div>
```
andrewazores: It'd be nice to use `*ngFor` here rather than copy-pasting the placeholder cards.

src/app/space/create/deployments/apps/deployments-apps.component.less
```diff
@@ -0,0 +1,18 @@
+.filler {
+  background-color: white;
+  margin: 0px 10px 10px 10px;
+}
+
+.spacer {
+  margin: 0px 10px 10px 10px;
+}
+
+#loading-msg {
```
andrewazores: Could you do this with a class rather than an ID?

src/app/space/create/deployments/apps/deployments-apps.component.html
```diff
@@ -1,32 +1,82 @@
...
-    <div class="col-xs-12 col-sm-4 col-md-2">
-      <h3>
-        <b>Applications</b>
-      </h3>
-    </div>
-    <div class="col-xs-12 col-sm-4 col-md-2">
-      <h3>
-        <b>Environments</b>
-      </h3>
+<div *ngIf="(applications.first() | async); else loading">
```
andrewazores: Much better, although it's still possible for the request for applications to resolve very quickly and the request for environments to resolve very slowly, in which case the loading screen is again prematurely hidden. Perhaps rather than displaying the loading screen conditionally on just `applications` this should be a merged observable of both `applications` and `environments`' first results?

src/app/space/create/deployments/resource-usage/loading-utilization-bar.component.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  private currentState: State;
+
+  ngOnInit(): void {
+    this.currentState = State.Okay;
+  }
+
+  getUtilizationClass(): State {
+    return this.currentState;
+  }
```
andrewazores: This component is completely static, so there's a bunch of machinery happening here to return the hardcoded 'utilization-okay' string to the [ngClass] in the template. Why not just inline it and get rid of all this component code?

src/app/space/create/deployments/resource-usage/loading-utilization-bar.component.html
```diff
@@ -0,0 +1,12 @@
+<div class="progress-container progress-description-left progress-label-right">
+    <div class="progress-description">
+      {{ resourceTitle }} ({{ resourceUnit }})
+    </div>
+    <div class="progress">
+      <div class="progress-bar" role="progressbar" [attr.aria-valuenow]="0" aria-valuemin="0" aria-valuemax="100"
+        [ngClass]="getUtilizationClass()"
+        [style.width]="0 + '%'" data-toggle="tooltip" title="Loading">
```
andrewazores: Does the style.width really need to be computed from an expression as a number concatenated with a string? Why not just give it a string literal?

src/app/space/create/deployments/resource-usage/deployments-resource-usage.component.html
```diff
@@ -11,11 +11,25 @@ <h3>
...
+            <div [collapse]="resourcesCollapsed">
+              <div class="col-xs-12 col-sm-4 col-md-3">
+                  <loading-resource-card></loading-resource-card>
+                </div>
+                <div class="col-xs-12 col-sm-4 col-md-3">
+                  <loading-resource-card></loading-resource-card>
+                </div>
+            </div>
+          </div>
+        </ng-template>
```
andrewazores: I think this part can be cleaned up and streamlined a bit. The `*ngIf` being around the `[collapse]` seems funny since the collapsible section is there whether or not placeholders are being displayed. The columns containing the cards are also the same size whether they are real cards or placeholders.

Also, since you are actually using the result of the environments observable in this template (as opposed to the template in the other commit), there's no need to use `.first()` on it here. You can use the `environments | async as environmentsList; else loading` syntax for example, and then change the `*ngFor` to iterate over `environmentsList` rather than `environments | async`.

src/app/space/create/deployments/resource-usage/deployments-resource-usage.component.html
```diff
@@ -11,9 +11,18 @@ <h3>
             </b>
           </h3>
         </div>
-        <div [collapse]="resourcesCollapsed">
-          <div class="col-xs-12 col-sm-4 col-md-3" *ngFor="let environment of environments | async">
-            <resource-card [spaceId]="spaceId | async" [environment]="environment"></resource-card>
+        <div>
```
andrewazores: Does this outermost `<div>` with no attributes do something meaningful?

src/app/space/create/deployments/resource-usage/deployments-resource-usage.component.html
```diff
@@ -11,9 +11,18 @@ <h3>
...
+          <div [collapse]="resourcesCollapsed">
+            <div *ngIf="environments | async as environmentsList; else loading">
+              <div class="col-xs-12 col-sm-4 col-md-3" *ngFor="let environment of environmentsList">
+                <resource-card [spaceId]="spaceId | async" [environment]="environment"></resource-card>
+              </div>
+            </div>
+            <ng-template #loading>
+              <div class="col-xs-12 col-sm-4 col-md-3" *ngFor="let num of [0, 1]">
+                <loading-resource-card></loading-resource-card>
```
jiekang: Rather than add a `loading-resource-card` template and having it be replaced by the `resource-card` when loaded, I would prefer the `resource-card` to have the initial loading state, where the side text displays 'loading' and the percent is 0. When data comes in, it will update appropriately. The loading state as a separate template should not be necessary.

andrewazores: +1

chrislessard: @jiekang While I agree this would be nicer, I don't quite understand how it can be done. The resource card takes two inputs - an environment and a string - and in the time where the environments observable is waiting for a value, at least one of these won't be available. With that as the case, resource-card's ngOnInit will fail. 

I can see that the change would be to wait for the environment to become available, then run ngOnInit, but is that possible in js ? Moreover, in deployments-resource-usage.component.html, I would still need to hard code two resource cards to the loading template to fit our design spec. Wouldn't these just be replaced anyways when the ngIf recognised that the observable had received a value ?

src/app/space/create/deployments/resource-usage/resource-card.component.ts
```diff
@@ -1,6 +1,7 @@
 import {
   Component,
   Input,
+  OnChanges,
```
andrewazores: This import appears to be unused

src/app/space/create/deployments/resource-usage/resource-card.component.html
```diff
@@ -6,3 +6,11 @@
     </div>
   </div>
 </ng-container>
+<ng-container *ngIf="!(environment)">
```
andrewazores: `!spaceId || !environment", right?

src/app/space/create/deployments/resource-usage/loading-utilization-bar.component.ts
```diff
@@ -0,0 +1,16 @@
+import {
+  Component,
+  Input,
+  OnDestroy,
+  OnInit
```
andrewazores: OnDestroy and OnInit imports are unused

andrewazores:
Nice, code looks good and testing this (locally) it does actually feel quite nice.

andrewazores:
Chris, please rebase and take a look at unit tests failures.

fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2600-8 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2600-fabric8-ui.badger.fabric8.io

PR #2594, opened 2018-03-01
fix(pipelines): fix pipeline filtering and sorting

andrewazores:
BuildConfigs do not contain expected "app" and "codebase" labels, which the Pipelines component
expects to be present and uses for filtering and sorting pipelines by their Application and
Codebase. The "id" and "gitUrl" properties of the BuildConfig are used instead.

fixes https://github.com/openshiftio/openshift.io/issues/2405

Also included are some code cleanup commits for the component.

Working on tests now, but since this is a small wrapper around a runtime-console component, it's taking a lot of time to mock up all of the runtime-console dependencies. I'll probably end up replacing the actual runtime-console child component(s) with mocks, so the only meaningful parts behind the test will have to do with this component's filtering and sorting.

PR #2601, opened 2018-03-02
feat(pipelines): support filtering Type Ahead

andrewazores:
fixes https://github.com/openshiftio/openshift.io/issues/2421

Implementation based on the Type Ahead example here: https://rawgit.com/patternfly/patternfly-ng/master-dist/dist-demo/#/filters

### DIFF
--- a/src/app/space/create/pipelines/pipelines.component.html
+++ b/src/app/space/create/pipelines/pipelines.component.html
@@ -3,7 +3,8 @@
     <pfng-toolbar [config]="toolbarConfig"
                   [viewTemplate]="addOpenShiftConsoleLink"
                   (onFilterChange)="filterChange($event)"
-                  (onSortChange)="sortChange($event)">
+                  (onSortChange)="sortChange($event)"
+                  (onFilterTypeAhead)="filterQueries($event)">
       <ng-template #addOpenShiftConsoleLink>
         <div class="openshift-links" placement="bottom">
           <div class='pull-right dropdown-kebab-pf dropdown' dropdown>
--- a/src/app/space/create/pipelines/pipelines.component.spec.ts
+++ b/src/app/space/create/pipelines/pipelines.component.spec.ts
@@ -305,6 +305,91 @@ describe('PipelinesComponent', () => {
       ]);
       expect(this.testedDirective.toolbarConfig.filterConfig.resultsCount).toEqual(2);
     });
+
+    describe('Type Ahead', () => {
+      it('should narrow available filters on user input', function(this: TestingContext) {
+        this.testedDirective.filterQueries({
+          field: {
+            id: 'application',
+            title: 'Application',
+            placeholder: 'Filter by Application...',
+            type: 'typeahead',
+            queries: [
+              {
+                id: 'app',
+                value: 'app'
+              },
+              {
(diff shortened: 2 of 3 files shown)


jiekang:
Hmm, while testing I noticed this behaves a little differently from the codebases and deployments page toolbars. On those, you can type and enter it as text to filter against. This one seems to allow you to type, but still only select an existing filter. As well, the selection can only be done via mouse, the 'enter' key doesn't do anything.

The codebases/deployments toolbars allow you to type and enter something to filter with, but don't have a predetermined list to select from. It would be nice if both features could exist together.

In the interim, I'd prefer this to follow the functionality on the codebases and deployments pages.

andrewazores:
Yes, the behaviour there is as intended - that's the "Type Ahead" pattern. The mouse-only selection is perhaps unfortunate but that does also follow with how the Type Ahead example works from the patternfly-ng examples.

IMO, Type Ahead is nicer than the basic filtering, since it by default displays all available filter queries (which is what the page initially tried to do before https://github.com/fabric8-ui/fabric8-ui/pull/2594), but allows filtering of the queries themselves. The "predetermined list" is actually all of the available results in the view which might be matched by a filter. Removing this list in favour of basic filtering just means that users become able to enter filters which are guaranteed not to match anything, and have no hints as to what a valid query term might be. If this is the preferred design I can change the implementation, but I don't see the advantage in going that direction personally. @catrobson ?

jiekang:
I agree the design isn't that great, but as a user I prefer being able to enter my own text, letting me filter against projects that are labelled a certain way. E.g.

projects:

osio-blah
osio-blob
prod-z
prod-y

I can use osio to see all 'osio' related and 'prod' to see all prod related. The type ahead does not allow this kind of filtering; in general it's very limited.

It would be best if the behaviors could be combined, but this would be patternfly work. I suggest the basic filtering just because we already have two pages doing that and this would be consistent with that. We could make those two pages use type ahead too. Either way would work for me.

jiekang:
Also note that the behavior of type ahead means that this filter is essentially a 'show one application' button when "filtering" by application name . Application names are unique and unless someone creates an application that has a name that is part of the name of another application name, you can never select something that would show more than one application.

catrobson:
+1 that basic filtering is desired rather than type-ahead right now so users can put in partial matches. @dlabrecq - can you also verify that this is not currently supported with the type-ahead? If not, could either of you log a bug against NG around supporting a type ahead that allows for partial matching?

This means that we do need to support a "no results found" scenario.  You can see a null result set example here:  http://www.patternfly.org/pattern-library/forms-and-controls/filter/#design

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2601-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2601-fabric8-ui.badger.fabric8.io

dlabrecq:
@catrobson it is up to the developer to filter the list shown by the patternfly-ng type-ahead feature. The user would then select from the list matching the partial text entered in the input field. The type ahead feature was a contribution, so keyboard navigation needs some work (i.e., it could be made easier for the user to select the partial match).

PR #2607, opened 2018-03-06
fix(deployments): guarantee pod status key ordering

andrewazores:
fixes https://github.com/openshiftio/openshift.io/issues/2473


src/app/space/create/deployments/services/deployments.service.spec.ts
```diff
@@ -1290,9 +1292,11 @@ describe('DeploymentsService', () => {
       const expectedResponse = {
         total: 6,
```
jiekang: I guess pod_total and the actual total aren't connected in any way since the backend supplies the values. Would be nice to have it correct in the tests; though it's a trivial nit.

andrewazores: Whoops - yea, good catch. I'll update it so the test acts as a better reference for what a real response is expected to look like.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2607-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2607-fabric8-ui.badger.fabric8.io

joshuawilson:
please add tests

jiekang:
@joshuawilson the code already has tests exercising this. The patch also modifies the relevant test data to assert this sorting occurs.

joshuawilson:
I must have been sleeping when I looked at. Approved.

PR #2610, opened 2018-03-06
app-launcher: build slow due to ngx-forge.js bundle size

dlabrecq:
This fixes an issue with the fabric8-ui build running too slow.

The ngx-forge package now uses tree shaking to reduce the ngx-forge.js bundle size by half. Thus, the build no longer appears to hang while Angular resolves duplicates.


invincibleJai:
@dlabrecq thanks :)

PR #2612, opened 2018-03-06
fix(pipelines): use "text" filtering rather than "select"

andrewazores:
fixes https://github.com/openshiftio/openshift.io/issues/2421


andrewazores:
@jiekang thanks, fixed. Had to modify one existing test and added another to explicitly test this case.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2612-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2612-fabric8-ui.badger.fabric8.io

PR #2614, opened 2018-03-08
fix(profile): show email on profile overview page if email is public

rohitkrai03:
Fixes a bug in [PR](https://github.com/fabric8-ui/fabric8-ui/pull/2489).

- Found a bug in the logic of showing email based on private/public status of user.
- Since we're using conditional ng-template already, *ngIf on ng-template doesn't work. [Refer](https://stackoverflow.com/questions/44837756/why-ngif-doesntwork-with-ng-template)
- So, moved *ngIf to p tag inside of ng-template instead.
- Also, we don't need the condition for addEmail element since that will never be shown if user has an email.

### DIFF
--- a/src/app/profile/overview/overview.component.html
+++ b/src/app/profile/overview/overview.component.html
@@ -20,13 +20,13 @@
         </p>
         <hr>
         <div *ngIf="(context.user.attributes.email !== undefined && context.user.attributes.email.length !== 0); then showEmail else addEmail"></div>
-        <ng-template #showEmail *ngIf="context.user.attributes.emailPrivate">
-          <p class="col-sm-12 profile-wrapper-email truncate">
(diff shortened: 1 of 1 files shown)


fabric8cd:
@rohitkrai03 fabric8/fabric8-ui:SNAPSHOT-PR-2614-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2614-fabric8-ui.badger.fabric8.io

sbose78:
[test]

fabric8cd:
@rohitkrai03 fabric8/fabric8-ui:SNAPSHOT-PR-2614-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2614-fabric8-ui.badger.fabric8.io

PR #2617, opened 2018-03-08
Memory usage warning

andrewazores:
targeting https://github.com/openshiftio/openshift.io/issues/2519

This PR adds a new service, `DeploymentStatusService`, which provides information about the status of a deployment. Currently this status only reflects resource usage vs quota for CPU and Memory metrics, but the intent is for it to evolve and provide status information for all deployment failure modes (see https://github.com/openshiftio/openshift.io/issues/2086). This is implemented as a separate service building on top of the existing `DeploymentsService` since it's at a higher conceptual abstraction level, and to avoid growing the already very large DeploymentsService any further.

This new DeploymentStatus service incorporates the pre-existing logic that was contained in the DeploymentCardComponent to set the card icon state, which was determined from CPU usage vs quota only. It extends this by also including Memory usage, and providing an aggregate status as well which combines both. This aggregate status is used by the DeploymentCardComponent to set its icon state.

Additionally, the DeploymentStatusService is intended to be useful for the implementation for this separate but related issue: https://github.com/openshiftio/openshift.io/issues/2520

I have an opinion question for the reviewer(s): should the DeploymentCardComponent remain responsible for setting the icon state, or should this responsibility be rolled into the DeploymentStatusIconComponent itself? Currently the icon component is quite "dumb" and simply takes two inputs and uses them in its template directly, with no internal logic of its own.

### DIFF
--- a/src/app/space/create/deployments/apps/deployment-card.component.spec.ts
+++ b/src/app/space/create/deployments/apps/deployment-card.component.spec.ts
@@ -251,43 +251,6 @@ describe('DeploymentCardComponent', () => {
     expect(detailsComponent.active).toBeTruthy();
   });
 
-  describe('iconStatusLogic', () => {
-    it("should set the icon's initial value to ok", function(this: Context) {
-      expect(this.testedDirective.iconClass).toBe(DeploymentStatusIconComponent.CLASSES.ICON_OK);
-      expect(this.testedDirective.toolTip).toBe('Everything is OK.');
-    });
-
-    it("should change the icon's value to warning if capacity changes", function(this: Context) {
-      mockCpuData.next([{ used: 4, quota: 5, timestamp: 2 }] as CpuStat[]);
-      this.detectChanges();
-      expect(this.testedDirective.iconClass).toBe(DeploymentStatusIconComponent.CLASSES.ICON_WARN);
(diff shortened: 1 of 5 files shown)


fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2617-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2617-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/apps/deployment-card.component.spec.ts
```diff
@@ -130,7 +131,8 @@ describe('DeploymentCardComponent async tests', () => {
         providers: [
           BsDropdownConfig,
           { provide: NotificationsService, useValue: notifications },
-          { provide: DeploymentsService, useValue: mockSvc }
+          { provide: DeploymentsService, useValue: mockSvc },
+          DeploymentStatusService
```
jiekang: A bit surprised this isn't mocked;

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2617-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2617-fabric8-ui.badger.fabric8.io

PR #2622, opened 2018-03-12
Deployments warning indicators

andrewazores:
https://github.com/openshiftio/openshift.io/issues/2520

These patches modify the Deployments page deployment and resource usage cards to show more warning/error indicators when resources are nearing quotas, as shown in the updated designs docs linked in the above issue.

Before, without warning:
screenshot from 2018-03-12 12-51-14

Before, with warning (forced by setting warning threshold to 10% in local build):
screenshot from 2018-03-12 13-07-57

After (forced warning in local build):
screenshot from 2018-03-12 12-52-03

(After without warning looks identical to Before without warning)

One difference in the implemented "After" screenshot vs the design is the size of the warning "ribbon" on a deployment card's left edge. In the design mock this ribbon is the full height of the card component. In the actual implementation, it is only the height of the inner card content component. This is because the card component grows in height when expanded to display its charts, so implementing the ribbon to take the entire left edge results in the ribbon expanding down the length of the card when expanded as well. This should be fixable but will require some refactoring of the card component to support, so I have left it out for now. If it looks good as-is then great, otherwise I will add a commit on top of this changeset to refactor the card component so that it's possible to highlight the entire collapsed height without highlighting the expanded height.

### DIFF
--- a/src/app/space/create/deployments/apps/deployment-card.component.less
+++ b/src/app/space/create/deployments/apps/deployment-card.component.less
@@ -12,16 +12,17 @@
   justify-content: center;
 }
 
-.status-ribbon-warn {
-  border-left-style: solid;
-  border-left-width: 4px;
-  border-left-color: @color-pf-orange-300;
-}
-
-.status-ribbon-err {
-  border-left-style: solid;
-  border-left-width: 4px;
-  border-left-color: @color-pf-red-100;
+.status-ribbon {
+  &-warn {
+    border-left-style: solid;
+    border-left-width: 4px;
+    border-left-color: @color-pf-orange-300;
+  }
+  &-err {
+    border-left-style: solid;
+    border-left-width: 4px;
+    border-left-color: @color-pf-red-100;
+  }
 }
 
 .deployment-dropdown {
--- a/src/app/space/create/deployments/apps/deployment-card.component.less
+++ b/src/app/space/create/deployments/apps/deployment-card.component.less
@@ -12,6 +12,18 @@
   justify-content: center;
 }
 
+.status-ribbon-warn {
+  border-left-style: solid;
+  border-left-width: 4px;
+  border-left-color: @color-pf-orange-300;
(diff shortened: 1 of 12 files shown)


andrewazores:
@catrobson please take a look at the "After" screenshot vs the design, thanks!

catrobson:
@andrewazores Looks great to me, I think the implementation makes sense to capture the right information in the warning, I'm ok with the warning vertical bar not extending to full height of the bar in the header.

src/app/space/create/deployments/apps/deployment-card.component.less
```diff
@@ -12,6 +12,18 @@
   justify-content: center;
 }
 
+.status-ribbon-warn {
```
jiekang:
```
.status-ribbon {
  &-warn { ... },
  &-err { ... }
}
```

src/app/space/create/deployments/apps/deployment-details.component.html
```diff
@@ -4,23 +4,29 @@
     <deployments-donut [applicationId]="applicationId" [environment]="environment" [spaceId]="spaceId" [mini]="false"></deployments-donut>
 
     <ng-container *ngIf="hasPods | async">
-      <p class="resource-usage-header">Resource Usage</p>
+      <p class="resource-usage-header">
+        Resource Usage
+        <span class="badge resource-badge" *ngIf="usageMessage">
+          <span class="pficon-warning-triangle-o"></span>
+          {{usageMessage}}
```
jiekang: I think style for this is to have spaces like `{{ usageMessage }}`

src/app/space/create/deployments/apps/deployment-card.component.less
jiekang: This can also be done to other css changes.

andrewazores:
@jiekang two last commits address your review concerns

PR #2623, opened 2018-03-12
fix(deployments): truncate graph labels for width constraint

andrewazores:
fixes https://github.com/openshiftio/openshift.io/issues/2541

This uses the Angular "number" pipe (`DecimalPipe`: https://angular.io/api/common/DecimalPipe) to display a maximum number of fraction digits so that numbers like 2.0000001 don't force the text too wide for the card width.


andrewazores:
It's done within the HTML template, so no unit test is included. It's doable but seemed not really worthwhile for such a trivial change.

joshuawilson:
@andrewazores why did you merge this with only ci.centos passing?

andrewazores:
@joshuawilson shoot, I didn't realize :(

PR #2626, opened 2018-03-12
Deployment delete

chrislessard:
These patches connect our front end to the backend's new deletion API. I'll provide screenshots as soon as I can, I'm waiting on a few builds to finish so that I can have deployments to delete. 

Solves: https://github.com/openshiftio/openshift.io/issues/2474

### DIFF
--- a/src/app/space/create/deployments/apps/delete-deployment-modal.component.ts
+++ b/src/app/space/create/deployments/apps/delete-deployment-modal.component.ts
@@ -1,0 +1,33 @@
+import { Component, EventEmitter, Input, Output, ViewEncapsulation } from '@angular/core';
+import { ModalDirective } from 'ngx-bootstrap/modal';
+
+@Component({
+  host: {
+    class: 'create-dialog'
+  },
+  encapsulation: ViewEncapsulation.None,
+  selector: 'delete-deployment-modal',
+  templateUrl: './delete-deployment-modal.component.html'
+})
+export class DeleteDeploymentModal {
+
+  @Input() host: ModalDirective;
+  @Input() applicationId: string = '';
+  @Input() environmentName: string = '';
+  @Output() deleteEvent = new EventEmitter();
+
+  constructor() {}
+
+  public openModal(): void {
+    this.host.show();
+  }
+
+  public closeModal(): void {
+    this.host.hide();
+  }
+
+  public confirmDeletion() {
+    this.deleteEvent.emit();
+    this.host.hide();
+  }
+}
--- a/src/app/space/create/deployments/apps/deployment-card-container.component.ts
+++ b/src/app/space/create/deployments/apps/deployment-card-container.component.ts
@@ -7,6 +7,8 @@ import { Observable } from 'rxjs';
 
 import { Environment } from '../models/environment';
 
+import { DeleteDeploymentModal } from './delete-deployment-modal.component';
+
 @Component({
   selector: 'deployment-card-container',
   templateUrl: 'deployment-card-container.component.html'
--- a/src/app/space/create/deployments/apps/deployment-card.component.less
+++ b/src/app/space/create/deployments/apps/deployment-card.component.less
@@ -28,3 +28,11 @@
 .deployment-dropdown {
   transform: translateY(2em);
 }
+
+.delete-spinner {
+  margin: 0;
+}
+
+.delete-msg, .delete-spinner {
+  display: inline-block;
+}
--- a/src/app/space/create/deployments/apps/deployment-card.component.spec.ts
+++ b/src/app/space/create/deployments/apps/deployment-card.component.spec.ts
@@ -1,7 +1,9 @@
 import {
   Component,
   DebugElement,
-  Input
+  EventEmitter,
+  Input,
+  Output
 } from '@angular/core';
 import {
   ComponentFixture,
@@ -22,6 +24,8 @@ import {
   BsDropdownModule,
   BsDropdownToggleDirective
 } from 'ngx-bootstrap/dropdown';
+import { ModalDirective, ModalModule } from 'ngx-bootstrap/modal';
+
 import { ChartModule } from 'patternfly-ng';
 import 'patternfly/dist/js/patternfly-settings.js';
 import {
@@ -59,6 +63,18 @@ class FakeDeploymentsDonutComponent {
   @Input() applicationId: string;
   @Input() environment: Environment;
 }
+
+@Component({
+  selector: 'delete-deployment-modal',
+  template: ''
+})
+class FakeDeleteDeploymentModal {
+  @Input() host: ModalDirective;
+  @Input() applicationId: string;
+  @Input() environmentName: string;
+  @Output() deleteEvent = new EventEmitter();
+}
+
 @Component({
   selector: 'deployment-graph-label',
   template: ''
@@ -128,12 +144,14 @@ describe('DeploymentCardComponent async tests', () => {
         FakeDeploymentsDonutComponent,
         FakeDeploymentGraphLabelComponent,
         FakeDeploymentDetailsComponent,
-        FakeDeploymentStatusIconComponent
+        FakeDeploymentStatusIconComponent,
+        FakeDeleteDeploymentModal
       ],
       imports: [
         BsDropdownModule.forRoot(),
         CollapseModule.forRoot(),
-        ChartModule
+        ChartModule,
+        ModalModule.forRoot()
       ],
       providers: [
         BsDropdownConfig,
@@ -150,6 +168,7 @@ describe('DeploymentCardComponent async tests', () => {
     component.applicationId = 'mockAppId';
     component.environment = { name: 'mockEnvironment' } as Environment;
 
+    spyOn(component, 'openModal');
     fixture.detectChanges();
     flush();
     flushMicrotasks();
@@ -184,17 +203,22 @@ describe('DeploymentCardComponent async tests', () => {
       expect(item).toBeFalsy();
     }));
 
-    it('should invoke service \'delete\' function on Delete item click', fakeAsync(() => {
+    it('should call the delete modal open method', fakeAsync(()  => {
       const item: DebugElement = getItemByLabel('Delete');
       expect(item).toBeTruthy();
-      expect(mockSvc.deleteDeployment).not.toHaveBeenCalled();
       item.query(By.css('a')).triggerEventHandler('click', new CustomEvent('click'));
-
       fixture.detectChanges();
 
-      expect(mockSvc.deleteDeployment).toHaveBeenCalled();
-      expect(notifications.message).toHaveBeenCalled();
+      expect(component.openModal).toHaveBeenCalled();
     }));
+
+    it('should call the delete service method when the modal event fires', fakeAsync(() => {
+      const de: DebugElement = fixture.debugElement.query(By.directive(FakeDeleteDeploymentModal));
+      expect(mockSvc.deleteDeployment).not.toHaveBeenCalled();
+      de.componentInstance.deleteEvent.emit();
+      expect(mockSvc.deleteDeployment).toHaveBeenCalled();
+    }));
+
   });
 
   it('should not display inactive environments', fakeAsync(() => {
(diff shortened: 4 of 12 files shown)


src/app/space/create/deployments/apps/delete-deployment-modal.component.ts
```diff
@@ -0,0 +1,56 @@
...
+import { ModalDirective } from 'ngx-bootstrap/modal';
+import { Subscription } from 'rxjs';
+import { Environment } from '../models/environment';
+
+import { DeploymentsService } from '../services/deployments.service';
+import { DeploymentCardComponent } from './deployment-card.component';
+
+@Component({
+  host: {
+    'class': 'create-dialog'
```
andrewazores: Quotation marks aren't necessary on the "class" key here

src/app/space/create/deployments/apps/delete-deployment-modal.component.ts
```diff
@@ -0,0 +1,56 @@
...
+    this.applicationId = '';
+    this.environmentName = '';
+    this.host.hide();
+  }
+
+  public confirmDeletion() {
+    let success = this.deploymentCardComponent.delete();
+
+    if (success) {
+      this.applicationList.filter(id => id !== this.applicationId);
```
andrewazores: The modal is the one that removes the deleted application from its parent's applications list? This seems weird. I think the modal should just present the confirmation and report what selection was made to its parent. Then the parent can decide to call its own deletion function. I also think it's the wrong path to head down to have the component manually filter out its own changes this way - this should be handled by the DeploymentsService. After a deletion request completes the service should fire off a new poll cycle, which will automatically push state changes to all subscribers.

src/app/space/create/deployments/apps/deployment-card.component.ts
```diff
@@ -129,23 +131,35 @@ export class DeploymentCardComponent implements OnDestroy, OnInit {
...
+            console.log(error);
             this.notifications.message({
               type: NotificationType.WARNING,
               message: error
             });
           }
         )
     );
+
+    return succeeded;
```
andrewazores: Does this really work? You're returning synchronously and setting the flag within an async handler.

src/app/space/create/deployments/apps/deployment-card.component.ts
```diff
@@ -129,23 +131,35 @@ export class DeploymentCardComponent implements OnDestroy, OnInit {
...
+        ).subscribe(
           (success: string) => {
             this.notifications.message({
               type: NotificationType.SUCCESS,
               message: success
             });
+            succeeded = true;
           },
           (error: any) => {
+            console.log(error);
```
andrewazores: Please either remove this console.log or replace it with a call to Logger

chrislessard:
[build]

chrislessard:
[test]

src/app/space/create/deployments/apps/delete-deployment-modal.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  encapsulation: ViewEncapsulation.None,
+  selector: 'delete-deployment-modal',
+  templateUrl: './delete-deployment-modal.component.html',
+  styleUrls: ['./delete-deployment-modal.component.less']
+})
+export class DeleteDeploymentModal {
+
+  @Input() host: ModalDirective;
+
+  deploymentCardComponent: DeploymentCardComponent;
```
andrewazores: Rather than holding a reference to its parent card component, this dialog component should have an `@Output` emitting events to signify when the user has confirmed they do (or don't) want to delete an application. The parent component would then listen for these events and handle them accordingly.

src/app/space/create/deployments/apps/deployment-card.component.less
```diff
@@ -28,3 +28,11 @@
 .deployment-dropdown {
   transform: translateY(2em);
 }
+
+.deleteSpinner {
```
andrewazores: Please use `class-name` naming scheme rather than `className`

src/app/space/create/deployments/apps/deployment-card.component.ts
```diff
@@ -43,10 +44,12 @@ export class DeploymentCardComponent implements OnDestroy, OnInit {
   @Input() spaceId: string;
   @Input() applicationId: string;
   @Input() environment: Environment;
+  @Input() modal: DeleteDeploymentModal;
 
   active: boolean = false;
   detailsActive: boolean = false;
   collapsed: boolean = true;
+  deletingDeployment: boolean = false;
```
andrewazores: I don't think the "deployment" needs to be part of this property's name - it already belongs to a component representing a deployment.

src/app/space/create/deployments/apps/deployment-card.component.ts
```diff
@@ -140,10 +145,22 @@ export class DeploymentCardComponent implements OnDestroy, OnInit {
...
 
+  openModal(): void {
+    this.modal.openModal(this);
+  }
+
+  lockAndDelete(): void {
+    this.collapsed = true;
+    this.deletingDeployment = true;
+  }
+
```
andrewazores: If these functions aren't meant to be called from within the template (ie only from elsewhere within the component class), please make them private

src/app/space/create/deployments/apps/deployments-apps.component.html
```diff
@@ -29,8 +29,23 @@ <h4 class="env-card-title">
...
+
+
+  <div class="modal fade" bsModal #modal="bs-modal" tabindex="-1" role="dialog" aria-hidden="true">
+    <div class="modal-dialog">
+      <div class="modal-content">
+        <delete-deployment-modal #deleteDeploymentModal [host]="modal"></delete-deployment-modal>
+      </div>
+    </div>
+  </div>
+
```
andrewazores: I find it a little confusing that this modal is located here in this template, but then a reference to it is passed down through the container and into each card. Is this required to have the modal appear as expected within the entire view?

src/app/space/create/deployments/apps/deployments-apps.component.ts
```diff
@@ -31,6 +34,9 @@ export class DeploymentsAppsComponent implements OnInit, OnDestroy {
   @Input() environments: Observable<Environment[]>;
   @Input() spaceId: Observable<string>;
 
+  @ViewChild(DeleteDeploymentModal) createAreaDialog: DeleteDeploymentModal;
```
andrewazores: Is this property used for anything? It also needs renaming if it is.

aptmac:
Not sure if it's within the scope of this PR or should be addressed later, but after I have successfully deleted a deployment I get an error notification telling me that it no longer exists.

2018-03-13-145739_657x118_scrot

chrislessard:
@aptmac When does this notification show up ? Immediately after you have delete the deployment ? I don't think the delete api has made its way into production yet, I had to point my copy of the UI at a local copy of WIT running with Elliot's changes to successfully delete deployments.

andrewazores:
I wonder if that can happen if the service itself has pending requests for the deployment you've just deleted. Either requests in flight at the time, or if we need to clear some cached state within the service.

aptmac:
@chrislessard right, I may have jumped the gun, disregard my image then (it's fun to delete things though!). I tried out your branch locally running the prod-cluster.sh

src/app/space/create/deployments/apps/deployment-card.component.spec.ts
```diff
@@ -184,17 +202,22 @@ describe('DeploymentCardComponent async tests', () => {
...
 
-      expect(mockSvc.deleteApplication).toHaveBeenCalled();
-      expect(notifications.message).toHaveBeenCalled();
+      expect(component.openModal).toHaveBeenCalled();
     }));
+
+    it('should call the delete service method when the modal event fires', fakeAsync(() => {
+      const de: DebugElement = fixture.debugElement.query(By.directive(FakeDeleteDeploymentModal));
+      de.componentInstance.deleteEvent.emit();
+      expect(component.delete).toHaveBeenCalled();
```
andrewazores: Add a `expect(component.delete).not.toHaveBeenCalled()` at the top of this test as well so that it's clear that the event emission triggered the call. Rather than testing component.delete, would it make sense to test `mockSvc.deleteDeployment` being called instead? This promotes a little bit looser coupling IMO.

src/app/space/create/deployments/apps/delete-deployment-modal.component.ts
```diff
@@ -0,0 +1,48 @@
...
+  applicationId: string = '';
+  environmentName: string = '';
+
+  constructor(
+    private deploymentsService: DeploymentsService) {
+  }
+
+  public openModal(applicationId: string, environmentName: string): void {
+    this.applicationId = applicationId;
+    this.environmentName = environmentName;
```
andrewazores: Could `applicationId` and `environmentName` be converted to `@Input` properties of the modal, rather than being passed in as params to this function call?

src/app/space/create/deployments/apps/delete-deployment-modal.component.ts
```diff
@@ -0,0 +1,48 @@
...
+    this.host.show();
+  }
+
+  public closeModal(): void {
+    this.applicationId = '';
+    this.environmentName = '';
```
andrewazores: Since the modal now belongs to only one deployment card, rather than being reused, do we need to worry about resetting these properties here?

fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2626-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2626-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/apps/delete-deployment-modal.component.ts
chrislessard: Good call.

src/app/space/create/deployments/apps/delete-deployment-modal.component.ts
```diff
@@ -0,0 +1,43 @@
...
+})
+export class DeleteDeploymentModal {
+
+  @Input() host: ModalDirective;
+  @Input() applicationId: string = '';
+  @Input() environmentName: string = '';
+  @Output() deleteEvent = new EventEmitter();
+
+  constructor(
+    private deploymentsService: DeploymentsService) {
```
andrewazores: Sorry, missed this on the last round. Now that we're emitting events rather than directly issuing the deletion call from the modal we don't need this service injected anymore. There are also several imports to be removed.

src/app/space/create/deployments/apps/deployment-card.component.spec.ts
```diff
@@ -184,17 +203,22 @@ describe('DeploymentCardComponent async tests', () => {
...
       expect(item).toBeTruthy();
-      expect(mockSvc.deleteDeployment).not.toHaveBeenCalled();
       item.query(By.css('a')).triggerEventHandler('click', new CustomEvent('click'));
-
       fixture.detectChanges();
 
+      expect(component.openModal).toHaveBeenCalled();
+    }));
+
+    fit('should call the delete service method when the modal event fires', fakeAsync(() => {
```
andrewazores: s/fit/it/

src/app/space/create/deployments/apps/delete-deployment-modal.component.ts
chrislessard: Good catch

fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-2626-6 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2626-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/apps/deployment-card-container.component.ts
```diff
@@ -7,6 +7,8 @@ import { Observable } from 'rxjs';
 
 import { Environment } from '../models/environment';
 
+import { DeleteDeploymentModal } from './delete-deployment-modal.component';
```
joshuawilson: Is this used? You don't call it in the file.

src/app/space/create/deployments/apps/deployments-apps.component.ts
```diff
@@ -19,6 +19,7 @@ import {
 
 import { DeploymentsToolbarComponent } from '../deployments-toolbar/deployments-toolbar.component';
 import { Environment } from '../models/environment';
+import { DeleteDeploymentModal } from './delete-deployment-modal.component';
```
joshuawilson: doesn't look like you call this after it is imported

joshuawilson:
@chrislessard you can address my questions in another PR if needed, they aren't holding this up.

PR #2627, opened 2018-03-13
Added ids to pod status entires.

ljelinkova:
I need to add some ids to the Deployments page for E2E testing. I've tried to do it myself but I am not sure if it works correctly and/or it is done right. So, any inputs would be appreciated.

### DIFF
--- a/src/app/space/create/deployments/deployments-donut/deployments-donut-chart/deployments-donut-chart.component.html
+++ b/src/app/space/create/deployments/deployments-donut/deployments-donut-chart/deployments-donut-chart.component.html
@@ -14,7 +14,7 @@
     <ng-template #elsePods>
       <div>
         Pod status:
-        <li *ngFor="let column of pods.pods">
+        <li *ngFor="let column of pods.pods"  [attr.id]="'pod_status_' + column[0]">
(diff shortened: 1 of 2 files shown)


joshuawilson:
[test]

src/app/space/create/deployments/deployments-donut/deployments-donut-chart/deployments-donut-chart.component.html
```diff
@@ -14,7 +14,7 @@
     <ng-template #elsePods>
       <div>
         Pod status:
-        <li *ngFor="let column of pods.pods">
+        <li *ngFor="let column of pods.pods"  [attr.id]="'pod_status_' + column[0]">
```
joshuawilson: What does `[attr.id]` do? Why not just use `id`?

ljelinkova: I've used this answer in StackOverflow 

https://stackoverflow.com/questions/34568497/how-to-set-dynamic-idngfor-in-ionic2-angular2

I've just tried id = "'pod_status_' + column[0]" and it did not work.

joshuawilson: good to know, thanks

PR #2630, opened 2018-03-13
Deployments: "0 used / 0 quota" should not cause a resource usage quota error after scaling down to 0 pods

andrewazores:
https://github.com/openshiftio/openshift.io/issues/2542

The fix applied here is to have the DeploymentStatusService also account for the Pods belonging to a deployment when determining its resource usage/quota status. If the deployment has 0 pods then the status for CPU and Memory usages are defined to be OK. The previous behaviour erroneously compared 0 usage to 0 quota and determined that there was a problem since the usage had reached the quota.


andrewazores:
[test]

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2630-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2630-fabric8-ui.badger.fabric8.io

PR #2635, opened 2018-03-14
fix(codebases): handle missing GitHub repos

andrewazores:
Handle case where a GitHub repo has been deleted without being removed from a Space by sending a
notification and continuing to process other codebases in the space

fixes https://github.com/openshiftio/openshift.io/issues/2548

The root of the problem is that a Codebase's GitHub repository may disappear, but the existing logic in this component assumes that all Codebases will have a valid GitHub link.

When a repo is deleted the GitHubService `getRepoDetailsByUrl` requests for the repository's URL will fail with a 404. This becomes an error emission on the returned Observable. The component tries to fetch details for all of its known Codebases, which is done sequentially. If any of these requests fail and emits an error then the `Observable.forkJoin` errors out early as well, emitting no `Codebase[]` value, leaving the template in its empty state. The solution here sends a notification to alert the user that some Codebases details are not shown because they could not be retrieved from GitHub and maps the result to `null` instead (rather than an error emission). This allows the `forkJoin` to continue working as expected. The result emitted from the forkJoin is then remapped to filter out the null results so that all that remains are the valid Codebases where the repository details could be fetched from GitHub.

### DIFF
--- a/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.html
+++ b/src/app/space/create/codebases/codebases-item-actions/codebases-item-actions.component.html
@@ -3,14 +3,18 @@
     <span class="fa fa-ellipsis-v"></span>
   </button>
   <ul class="dropdown-menu-right dropdown-menu" aria-labelledby="dropdownKebab" *dropdownMenu>
-    <li [ngClass]="{'disabled': workspaceBusy}"
-        [attr.role]="menuitem">
-      <a href="javascript:void(0)" class="dropdown-item secondary-action" *ngIf="!workspaceBusy"
+    <li [ngClass]="{'disabled': workspaceBusy || !codebase.gitHubRepo}"
+        [attr.role]="'menuitem'">
+      <a href="javascript:void(0)" class="dropdown-item secondary-action" *ngIf="!workspaceBusy && codebase.gitHubRepo; else createDisabled"
          (click)="createAndOpenWorkspace()">Create workspace</a>
-      <a href="javascript:void(0)" class="dropdown-item secondary-action" *ngIf="workspaceBusy"
-         tooltip="Waiting for Che..." disabled>Create workspace</a>
+      <ng-template #createDisabled>
+        <a href="javascript:void(0)" class="dropdown-item secondary-action" *ngIf="workspaceBusy"
+          tooltip="Waiting for Che..." disabled>Create workspace</a>
+        <a href="javascript:void(0)" class="dropdown-item secondary-action" *ngIf="!codebase.gitHubRepo"
+          tooltip="GitHub repository is not accessible" disabled>Create workspace</a>
+      </ng-template>
     </li>
-    <li [attr.role]="menuitem">
+    <li [attr.role]="'menuitem'">
       <a href="javascript:void(0)" class="dropdown-item secondary-action"
          (click)="confirmDeleteCodebase($event)">Remove codebase</a>
     </li>
--- a/src/app/space/create/codebases/codebases-item-details/codebases-item-details.component.html
+++ b/src/app/space/create/codebases/codebases-item-details/codebases-item-details.component.html
@@ -6,13 +6,14 @@
     </dl>
     <dl class="dl-horizontal" *ngIf="!isHtmlUrlInvalid()">
       <dt>Git URL</dt>
-      <dd><a [href]="htmlUrl" target="_blank">{{gitUrl}}</a></dd>
+      <dd><a [href]="codebase.gitHubRepo?.htmlUrl || codebase.name" target="_blank">{{gitUrl || codebase.name}}</a></dd>
       <dt>Latest Commit SHA1</dt>
-      <dd>{{lastCommit}}</dd>
+      <dd>{{lastCommit || 'Unavailable'}}</dd>
       <dt>Status</dt>
-      <dd>{{filesChanged}} total files changed</dd>
+      <dd *ngIf="codebase.gitHubRepo && lastCommit">{{filesChanged}} total files changed</dd>
+      <dd *ngIf="!codebase.gitHubRepo">GitHub repository is not accessible</dd>
       <dt>License</dt>
-      <dd>{{license || 'None'}}</dd>
+      <dd>{{license || 'Unavailable'}}</dd>
     </dl>
   </div>
 </div>
--- a/src/app/space/create/codebases/codebases-item-details/codebases-item-details.component.spec.ts
+++ b/src/app/space/create/codebases/codebases-item-details/codebases-item-details.component.spec.ts
@@ -3,7 +3,6 @@ import { async, TestBed } from '@angular/core/testing';
 import { FormsModule } from '@angular/forms';
 import { HttpModule } from '@angular/http';
 
-import { Logger } from 'ngx-base';
 import { Contexts } from 'ngx-fabric8-wit';
 import { Observable } from 'rxjs';
 
@@ -18,7 +17,6 @@ describe('Codebases Item Details Component', () => {
 
   beforeEach(() => {
     gitHubServiceMock = jasmine.createSpyObj('GitHubService', ['getRepoDetailsByUrl', 'getRepoLastCommitByUrl', 'getRepoCommitStatusByUrl', 'getRepoLicenseByUrl']);
-    loggerMock = jasmine.createSpy('Logger');
 
     TestBed.configureTestingModule({
       imports: [FormsModule, HttpModule],
@@ -29,9 +27,6 @@ describe('Codebases Item Details Component', () => {
         },
         {
           provide: GitHubService, useValue: gitHubServiceMock
-        },
-        {
-          provide: Logger, useValue: loggerMock
         }
       ],
       // Tells the compiler not to error on unknown elements and attributes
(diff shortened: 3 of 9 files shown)


andrewazores:
@joshuawilson not sure who should review this one, any suggestion?

andrewazores:
[test]

andrewazores:
@catrobson any guidance here?

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2635-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2635-fabric8-ui.badger.fabric8.io

andrewazores:
screenshot from 2018-03-14 14-47-28

I've changed (locally, not pushed) the CodebasesComponent logic to simply pass the Codebase object leaving the gitHubRepo property undefined on error, and updated a few of the other child components to handle this case and include a very basic "invalid repo" state if the gitHubRepo property is undefined. The above screenshot illustrates what this proof of concept currently looks like. I've also removed the error notification. Is this along the lines of what you had in mind @dlabrecq ?

dlabrecq:
Yes, I like we're that's headed. Although, a repo could also have been made private.

Do we still have the codebase name and Git URL available? I ask because if the user selected "remove codebase", the name would help clarify what is being removed.

@mindreeper2420 and @catrobson Could we show something like this below? This also disables the "create workspace" control in the kebab.

"Unavailable" snapshot
screen shot 2018-03-14 at 4 13 05 pm

mindreeper2420:
@dlabrecq I like the direction of that screen - it maintains the same look/feel of all of the other codebases in the list, while clearly informing the user that it is no longer available to use.

If we can still reference the URL that is now missing, that'd be helpful - it may have been removed on accident. Otherwise, I think a simple message is fine.

joshuawilson:
I like where you are going with this, nice work.

joshuawilson:
When I tried it out I saw 2 things.
1) I added a codebase and then deleted it on github. Nothing changed on the screen (I didn't expect it to). I clicked the expand arrow and got a bunch of 404s
2) I then refreshed the page and the codebase was gone but I did get a Notification pop-up and a 404 in the console

We do not want any 404's in the console. The code needs to catch the error and deal with it. Don't pass it up the chain in this case. 

I like the idea of leaving the URL in place like your image but that didn't happen for me.

andrewazores:
@joshua yea, while looking through this I noticed at least one spot where a child component duplicated the GitHubService calls made by its parent rather than using the already available Codebase properties passed into the child. So either each child needs to have its own handling for the error case (IMO, yuck), or the child components should only present the data passed to them by the parent (IMO, yay). I'll work on that later with this changeset and that'll resolve the problem of clicking the expand arrow and getting new errors. I'm not sure where the 404 in the console is coming from, but I've seen it happen with my patch applied too.

@dlabrecq the "patternfly/patternfly" or "andrewazores/space2-2" links actually come from the GitHubRepoDetails object, which is what we're unable to get anymore after the repo disappears. What we have left is the original upstream URL, which is also what the Codebase's name is set to. I can make the missing fields display "Unavailable" rather than blank, set the Status, and disable the "Create workspaces" menu item rather than remove it, though.

andrewazores:
@joshuawilson I've found the source of the 404 in the console. It comes from GitHubService line 139 - it is actually catching the error emission from the HTTP request observable, and then it passes that off to an error handler function which logs the error and then rethrows it. We then re-catch that in my modification to the CodebasesComponent fetchCodebases function. So it is actually fully handled. If you comment out/remove the `this.logger.error(error)` in GitHubService.handleError, you'll no longer see it in the console.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2635-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2635-fabric8-ui.badger.fabric8.io

joshuawilson:
Each of those calls may get handled but it is still network traffic.

andrewazores:
Yes. In my patch I have it set up so that the child component will not make those requests if the parent component's first request already failed. In the case you tested where you're already on the page and then delete the repo afterward those requests will still get made and then fail, since there was never a prior opportunity to discover the repo's deletion. I'm not sure what else can be done about that.

joshuawilson:
I expect that to fail. But you could still handle the errors and just change the display. 
e.g. I click the expand. It gets 404. Close the expanded view. Don't report the error. Change the URL line to be like the image above with the X on it.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2635-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2635-fabric8-ui.badger.fabric8.io

andrewazores:
The most recent prototype behaves how you describe other than re-collapsing the view @joshuawilson . This is how it looks:

screenshot from 2018-03-15 16-33-47

If this is what we want to proceed with, I'll clean up the commits and add tests.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2635-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2635-fabric8-ui.badger.fabric8.io

dlabrecq:
I like the "Github repository is not accessible" message -- that could also be used for the status.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2635-7 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2635-fabric8-ui.badger.fabric8.io

andrewazores:
Cleaned it up. The first two commits are for the original approach that simply filtered out the codebases with missing repos. The second two are for the error state. Should I squash them? The testing looks a little anemic but most of the work was actually in templates rather than component code, so I'm not sure how much we want to add testing for this.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2635-9 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2635-fabric8-ui.badger.fabric8.io

joshuawilson:
@dlabrecq do you still have an issue with this PR?

andrewazores:
I seem to have introduced a regression while rebasing. The behaviour when you delete the repo while on the Codebases view and then click the expansion arrow is wrong - it no longer correctly updates the item to the error state. Sorry, looking into this.

andrewazores:
Okay, interesting. There's actually no regression from the previous behaviour in the patch as far as I can tell. The issue I saw was that the CodebasesItemDetailsComponent.ngOnInit call to GitHubService.getRepoDetailsByUrl will actually emit twice each time the component is initialized, whether the repo is there or not. If the repo is not there then you get one (apparently valid) value emission and one error emission. I'm not sure why the Observable emits twice, and why it still emits a valid value in the case that the repo is actually gone. I think this counts as a separate issue to investigate though. Opinions?

dlabrecq:
The changes look good. Although, I'm not able to test prod-preview, I don't want to hold up this PR.

joshuawilson:
This still has a 404 in the console for the deleted repo. It does display correctly in the UI, but since we are handling the 404 internally, it shouldn't appear in the console.

andrewazores:
I don't see that @joshuawilson . What can I do to reproduce it? I've tried both loading the view when the repo is already gone, as well as loading the view, deleting the repo, and then clicking the expansion arrow.

joshuawilson:
I used the badger link above to test it. I had previously added a repo to a space in OSIO, then deleted it in github. I went to the space and codebases page and the 404 was in the console.

andrewazores:
Very strange. My testing path is the exact same and I don't see any console 404. I grepped the codebases directory and there are no calls to console.log/console.error/etc. that would produce this. Similarly, there are no logger calls that would produce it either - all of the logger calls are from services other than the GitHubService. I also read through all component code in the directory and verified that all subscriptions to HTTP requesting services include error handlers, and none of them would log the error to the console as far as I could see. So I'm not sure how this console 404 can be produced... :/

joshuawilson:
I did a hard refresh and I still get it. Can someone else please try.

jiekang:
@joshuawilson can you provide a screenshot of the 404 being logged in the console, preferably expanded, as well as a screenshot of the corresponding network request in the network tab?

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2635-10 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2635-fabric8-ui.badger.fabric8.io

andrewazores:
There are problems with either preview or my account I guess, so I'm unable to test properly with that badger link above. Testing locally I still can't reproduce the 404 being printed in the console. It does appear in the browser dev tools network tab, but this is expected - it's upon this error that we discover the repo has disappeared. This is handled by the CodebasesComponent#fetchCodebases error handler, which simply returns the unmodified Codebase object without the additional GitHub repo details inserted.

joshuawilson:
create-openshift-io

with this error in the console `GET https://api.github.com/repos/joshuawilson/ocean-short 404 (Not Found)`

andrewazores:
@joshuawilson ah, I think I've discovered the discrepancy:

https://bugs.chromium.org/p/chromium/issues/detail?id=124534

I have the "Hide Network" option toggled on in my Chrome developer tab, and also happen to do most of my development work in Firefox which seems to mimic that same behaviour. The default Chrome setting will display 404s in the console regardless of there being an error handler in the JS catching it. With that option toggled off there is nothing I can do to prevent the 404 from appearing in your browser console, as far as I can tell. There doesn't appear to be any request I can make to determine if the repo exists, just the current approach the GitHubService already takes and checking 200 vs 404 status code.

https://developer.github.com/v3/repos/

andrewazores:
[build]

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2635-11 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2635-fabric8-ui.badger.fabric8.io

joshuawilson:
I would like to avoid empty/useless errors showing up in Sentry.

andrewazores:
Will the error still show up in Sentry? There is an error handler in the code and it does not rethrow the error, so as far as I understand it this will not be propagated to Sentry. Google Chrome simply displays the 404 regardless of this fact, unless you toggle off the option I linked above. I do not see any way in the GitHub API that we can detect that the repository is missing other than by issuing this GET and catching the 404. This is not a regression or change in behaviour from before this patch is applied.

joshuawilson:
OK, I'll merge it and if it shows up we can deal with it.

PR #2636, opened 2018-03-14
fix(webpack): resolve webpack sourcemap and dependency warnings

aptmac:
This PR addresses the warnings in the console when running a local version of f8-ui. This is resolved in two parts: the `ContextReplacementPlugin` regex needed to be updated [0], and there were 5 dependencies that needed to be excluded in the webpack source-map-loader.

Additionally, there was a an extra comma at the end of the name tag in `index.ejs` which was causing the 

> "The key "" is not recognized and ignored."

 warning, and this persists in production so this fix should clean up the console. 

Here are a couple of screen-shots to show what my console looked like before and after this PR:

Before (there were 12 of these warning messages):
before

After:
after

[0] https://github.com/angular/angular/issues/11580#issuecomment-367296829

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -414,7 +414,7 @@ module.exports = function (options) {
        */
       new ContextReplacementPlugin(
         // The (\\|\/) piece accounts for path separators in *nix and Windows
-        /angular(\\|\/)core(\\|\/)(esm(\\|\/)src|src)(\\|\/)linker/,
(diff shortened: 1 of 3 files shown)


fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2636-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2636-fabric8-ui.badger.fabric8.io

PR #2637, opened 2018-03-14
fix(cleanup): fix errors being logged by the cleanup component unit tests

aptmac:
This short PR addresses errors that were being displayed when running the cleanup component unit tests. The cause of these errors are with the `component.contextSubscription` not being mocked, so when `ngOnDestroy()` is called it is trying to unsubscribe from an undefined object (related StackOverflow discussion [here](https://stackoverflow.com/questions/43350115/testing-angular-component-with-unsubscribe-error-during-cleanup-of-component/43471130) [0]). 

This fix creates a spy object for `component.contextSubscription` that spies on the unsubscribe function, which will not cause errors when `ngOnDestroy()` is called.

Before:
before

After:
after

[0] https://stackoverflow.com/questions/43350115/testing-angular-component-with-unsubscribe-error-during-cleanup-of-component/43471130

### DIFF
--- a/src/app/profile/cleanup/cleanup.component.spec.ts
+++ b/src/app/profile/cleanup/cleanup.component.spec.ts
@@ -19,7 +19,7 @@ describe('CleanupComponent', () => {
   let fixture: ComponentFixture<CleanupComponent>;
   let component: DebugNode['componentInstance'];
   let mockContexts: any = jasmine.createSpy('Contexts');
-  let mockSpaceService: any = jasmine.createSpy('SpaceService');
+  let mockSpaceService: any = jasmine.createSpyObj('SpaceService', ['deleteSpace', 'getSpacesByUser']);
   let mockTenantService: any = jasmine.createSpy('TenantService');
   let mockEventService: any = jasmine.createSpy('EventService');
   let mockRouter: any = jasmine.createSpyObj('Router', ['navigate']);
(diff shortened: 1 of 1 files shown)


andrewazores:
Rather than mocking the `Subscription` object and assigning this directly to the component property, could you instead mock the `Context` object in the injected `Contexts`'s `.current` and have it return a real `Observable`? I prefer to only directly set component properties like this for ones annotated with `@Input`, since this most closely mimics how the component will actually be interacted with at runtime.

src/app/profile/cleanup/cleanup.component.spec.ts
```diff
@@ -51,7 +72,7 @@ describe('CleanupComponent', () => {
     fixture = TestBed.createComponent(CleanupComponent);
     component = fixture.debugElement.componentInstance;
     component.confirmCleanup = jasmine.createSpyObj('IModalHost', ['open', 'close']);
-    mockSpace = jasmine.createSpy('Space');
+    component.ngOnInit();
```
andrewazores: `fixture.detectChanges()` should trigger the component lifecycle hooks, and IMO is preferable to manually calling the hooks like this.

andrewazores:
Looks pretty good now, I think it's OK after the `detectChanges` tweak.

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2637-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2637-fabric8-ui.badger.fabric8.io

src/app/profile/cleanup/cleanup.component.spec.ts
aptmac: Ah that's neat, I'll make this and re-push in a sec

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2637-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2637-fabric8-ui.badger.fabric8.io

PR #2638, opened 2018-03-14
Hierarchical areas

dlabrecq:
Replaced the existing list with pfng-tree-list to better show hierarchical areas. Also added a toolbar with filter/sort/add controls and spec.

Note: This updates patternfly-ng to 3.1.2 since the current version is missing CSS styles.

Fixes: https://github.com/fabric8-ui/fabric8-ux/issues/835

### DIFF
--- a/src/app/space/settings/areas/areas.component.html
+++ b/src/app/space/settings/areas/areas.component.html
@@ -7,7 +7,7 @@
 <div class="container-fluid">
   <div class="row">
     <div class="col-sm-12">
-      <pfng-tree-list *ngIf="showTree === true"
+      <pfng-tree-list
           [actionTemplate]="actionTemplate"
           [config]="treeListConfig"
           [itemTemplate]="itemTemplate"
--- a/src/app/space/settings/areas/areas.component.html
+++ b/src/app/space/settings/areas/areas.component.html
@@ -1,47 +1,39 @@
+<areas-toolbar
+    (onAddArea)="addChildArea()"
+    (onFilterChange)="filterChange($event)"
+    (onSortChange)="sortChange($event)"
+    [resultsCount]="resultsCount">
+</areas-toolbar>
 <div class="container-fluid">
   <div class="row">
-    <div class="col-xs-8 col-sm-8 col-md-9">
-      <h3 class="areas-header">Areas that make up the {{context.space.attributes.name | spaceName}} Space:</h3>
-    </div>
-    <div class="col-xs-4 col-sm-4 col-md-3">
-      <div class="table-action-heading" (click)="addChildArea()">
-        <a><i class="pficon pficon-add-circle-o"></i> Add Areas</a>
-      </div>
-    </div>
-  </div>
-  <div class="row">
-    <div class="col-md-12">
-      <pfng-list
-        [config]="listConfig"
-        [itemTemplate]="itemTemplate"
-        [actionTemplate]="actionTemplate"
-        [items]="areas" >
-        <ng-template #itemTemplate let-item="item" let-index="index">
+    <div class="col-sm-12">
+      <pfng-tree-list *ngIf="showTree === true"
+          [actionTemplate]="actionTemplate"
+          [config]="treeListConfig"
+          [itemTemplate]="itemTemplate"
+          [items]="treeAreas"
+          (onActionSelect)="handleAction($event)">
+        <ng-template #itemTemplate let-node="node" let-index="index">
           <div class="list-pf-content-wrapper">
             <div class="list-pf-main-content">
-              <div class="list-pf-title">{{itemPath(item.attributes)}}</div>
+              <div class="list-pf-title">{{node.data?.attributes?.name}}</div>
             </div>
           </div>
         </ng-template>
-        <ng-template #actionTemplate let-item="item" let-index="index">
-          <span class="dropdown-kebab-pf areas-display dropdown" dropdown>
-            <button class="btn btn-link dropdown-toggle" type="button" dropdownToggle>
-              <span class="fa fa-ellipsis-v"></span>
-            </button>
-            <ul class="dropdown-menu-right dropdown-menu" aria-labelledby="dropdownKebab" *dropdownMenu>
-              <li role="menuitem"><a href="javascript:void(0)" class="secondary-action" (click)="addChildArea(item.id)">Add Child Area</a></li>
-            </ul>
-          </span>
+        <ng-template #actionTemplate let-node="node" let-index="index">
+          <pfng-action class="list-pf-actions"
+                       [config]="actionConfig"
+                       (onActionSelect)="handleAction($event, node)">
+          </pfng-action>
         </ng-template>
-      </pfng-list>
+      </pfng-tree-list>
     </div>
   </div>
 </div>
-
 <div class="modal fade" bsModal #modal="bs-modal" tabindex="-1" role="dialog" aria-hidden="true" (onShown)="createAreaDialog.onOpen()" (onHide)="createAreaDialog.onClose()">
   <div class="modal-dialog">
     <div class="modal-content">
-      <create-area-dialog #createAreaDialog [host]="modal" [parentId]="selectedAreaId" [areas]="areas" (onAdded)="addArea($event)"></create-area-dialog>
+      <create-area-dialog #createAreaDialog [host]="modal" [parentId]="selectedAreaId" [areas]="allAreas" (onAdded)="addArea($event)"></create-area-dialog>
     </div>
   </div>
 </div>
--- a/src/app/space/settings/areas/areas.component.ts
+++ b/src/app/space/settings/areas/areas.component.ts
@@ -57,7 +57,6 @@ export class AreasComponent implements OnInit, OnDestroy {
   selectedAreaId: string;
   subscriptions: Subscription[] = [];
   resultsCount: number = 0;
-  showTree: boolean = true;
   treeListConfig: TreeListConfig;
 
   constructor(private contexts: ContextService,
@@ -132,15 +131,10 @@ export class AreasComponent implements OnInit, OnDestroy {
 
   addArea(area: ExtArea): void {
     this.allAreas.push(area);
-    this.showTree = false;
 
-    // Hide the tree list to rerender existing toggles in expanded state
-    setTimeout(() => {
-      // Reapply sort and filter, if any
-      this.allAreas.sort((area1: Area, area2: Area) => this.compare(area1, area2));
-      this.applyFilters(this.appliedFilters);
-      this.showTree = true;
-    }, 10);
+    // Reapply sort and filter, if any
+    this.allAreas.sort((area1: Area, area2: Area) => this.compare(area1, area2));
+    this.applyFilters(this.appliedFilters);
   }
 
   // Filter
@@ -214,7 +208,6 @@ export class AreasComponent implements OnInit, OnDestroy {
         let children = this.getNestedChildren(elements, element);
         if (children.length > 0) {
           element.children = children;
-          element.expanded = true;
         }
         tree.push(element);
       }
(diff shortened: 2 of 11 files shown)


fabric8cd:
@dlabrecq fabric8/fabric8-ui:SNAPSHOT-PR-2638-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2638-fabric8-ui.badger.fabric8.io

mindreeper2420:
@dlabrecq it appears that the PatternFly-ng CSS is having a clash of cursors on hovering of the child expansion carets. When I hover over the caret, I still have the standard cursor and not the pointer to signify a clickable element.

This bit from patternfly-ng.min.css `.pfng-tree-list .tree` is adding `cursor: default` and overwriting the behavior of `.tree`, which has `cursor: pointer`.

dlabrecq:
@mindreeper2420 The cursor is something we can fix in patternfly-ng. The cursor appears to have the same behavior as the demo: http://www.patternfly.org/patternfly-ng/#/treelist

dlabrecq:
@mindreeper2420 Fixed the cursor on patternfly-ng. If you want to merge, I'll update the version number here.

https://github.com/patternfly/patternfly-ng/pull/326

mindreeper2420:
@dlabrecq all set - approved and merged the patternfly-ng PR.

dlabrecq:
@mindreeper2420 Thanks! Updated the version number here

fabric8cd:
@dlabrecq fabric8/fabric8-ui:SNAPSHOT-PR-2638-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2638-fabric8-ui.badger.fabric8.io

src/app/space/settings/areas/areas.component.ts
```diff
@@ -18,69 +30,251 @@ export class AreasComponent implements OnInit, OnDestroy {
   @ViewChild(CreateAreaDialogComponent) createAreaDialog: CreateAreaDialogComponent;
   @ViewChild(ModalDirective) modal: ModalDirective;
 
+  private allAreas: Area[]; // flat array obtained directly from API
+  private filteredAreas: Area[]; // flat array filtered and sorted
+  private treeAreas: Area[]; // transformed flat array into tree list
```
joshuawilson: Looks like `treeAreas` is updated but it doesn't get read. Did you forget to use it or just not need it?

src/app/space/settings/areas/areas.component.ts
```diff
@@ -18,69 +30,251 @@ export class AreasComponent implements OnInit, OnDestroy {
...
+  sortChange($event: SortEvent): void {
+    this.currentSortField = $event.field;
+    this.isAscendingSort = $event.isAscending;
+    this.allAreas.sort((area1: Area, area2: Area) => this.compare(area1, area2));
+    this.applyFilters(this.appliedFilters); // Reapply filters, if any
+  }
+
+  // Transform flat array into tree list
+
+  private buildTree(elements: any[], tree = []): any[] {
```
jiekang: I don't agree with the use of `any` in order to add properties that don't exist to Typed objects. I would rather have the component define a new type that contains an Area along with it's extra tree information. It then becomes obvious to any developer unfamiliar with the code what the object contains and what things are used for.

src/app/space/settings/areas/areas.component.ts
```diff
@@ -18,69 +30,251 @@ export class AreasComponent implements OnInit, OnDestroy {
...
+      if (this.allAreas[i].id === id) {
+        area = this.allAreas[i];
+      }
+    }
+    return area;
+  }
+
+  // Actions
+
+  handleAction($event: Action, node: any): void {
```
jiekang: Is there a type for the `node` parameter? The function's implementation assumes it has specific properties while the signature suggests it can take any kind of object.

src/app/space/settings/areas/areas.component.html
```diff
@@ -1,47 +1,39 @@
+<areas-toolbar
+    (onAddArea)="addChildArea()"
+    (onFilterChange)="filterChange($event)"
+    (onSortChange)="sortChange($event)"
+    [resultsCount]="resultsCount">
```
jiekang: Angular emits warnings like `[Angular] Identifier 'resultsCount' refers to a private member of the component`. This notification exists because AoT will not work if there are any occurrences of html identifiers referring to a private member of the component [1][2]. I think there has been some discussion on seeing if AoT could improve the web application performance [3]. Could you fix these to be public now while you're touching this file?

[1]
https://stackoverflow.com/questions/45520838/identifier-refers-to-a-private-member-of-the-component
[2]
https://github.com/angular/angular/issues/11978
[3] 
https://github.com/openshiftio/openshift.io/issues/2498
https://github.com/openshiftio/openshift.io/issues/938

fabric8cd:
@dlabrecq fabric8/fabric8-ui:SNAPSHOT-PR-2638-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2638-fabric8-ui.badger.fabric8.io

src/app/space/settings/areas/areas.component.ts
dlabrecq: It's used in the HTML template by the pfng-tree-list component

src/app/space/settings/areas/areas.component.html
dlabrecq: Sure, I'll take a stab at it

src/app/space/settings/areas/areas.component.ts
dlabrecq: I'll extend the interface

src/app/space/settings/areas/areas.component.ts
dlabrecq: Unfortunately, this object is from the underlying angular-tree-component, which is type any.

src/app/space/settings/areas/areas.component.html
dlabrecq: My IDE currently doesn't show AOT warnings, but I've changed the variables to public.

dlabrecq:
Addressed the review comments and rebased.

fabric8cd:
@dlabrecq fabric8/fabric8-ui:SNAPSHOT-PR-2638-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2638-fabric8-ui.badger.fabric8.io

src/app/space/settings/areas/areas.component.html
jiekang: Ah, my bad. s|Angular|Angular Language Service in vscode|

I'm not sure if any other IDEs have similar support.

src/app/space/settings/areas/areas.component.ts
```diff
@@ -18,69 +43,251 @@ export class AreasComponent implements OnInit, OnDestroy {
...
+    this.allAreas.push(area);
+    this.showTree = false;
+
+    // Hide the tree list to rerender existing toggles in expanded state
+    setTimeout(() => {
+      // Reapply sort and filter, if any
+      this.allAreas.sort((area1: Area, area2: Area) => this.compare(area1, area2));
+      this.applyFilters(this.appliedFilters);
+      this.showTree = true;
+    }, 10);
```
jiekang: What's the reason for a 10 ms timeout for these lines?

dlabrecq: When an area is added, the tree list is briefly hidden in order to re-render existing toggles in expanded state. This is related to angular-tree-component and Angular's change detection strategy.

Alternatively, if we don't show the toggles expanded, the timeout can be removed. Or, if we want to fetch new data and rebuild the tree, that may force a change detection? However, that is probably less performant than adding the area directly to the existing list.

src/app/space/settings/areas/areas.component.html
dlabrecq: I couldn't get Webstorm to show AOT warnings; although, Joshua can somehow? I read that in Angular 5, "ng test" and "ng build -prod" will report errors when an html template accesses private members.

Other than making private members public, I wasn't able to make any further changes here. Running the build:prod-aot and build:aot targets didn't produce warnings for me.

src/app/space/settings/areas/areas.component.html
```diff
@@ -1,47 +1,39 @@
...
-  <div class="row">
-    <div class="col-md-12">
-      <pfng-list
-        [config]="listConfig"
-        [itemTemplate]="itemTemplate"
-        [actionTemplate]="actionTemplate"
-        [items]="areas" >
-        <ng-template #itemTemplate let-item="item" let-index="index">
+    <div class="col-sm-12">
+      <pfng-tree-list *ngIf="showTree === true"
```
jiekang: Use `*ngIf="showTree"` rather than a comparison evaluation

src/app/space/settings/areas/areas.component.ts
jiekang: With the above fix to the `ngIf` usage, this function can be shortened to:

```
    this.allAreas.push(area);
    this.allAreas.sort((area1: Area, area2: Area) => this.compare(area1, area2));
    this.applyFilters(this.appliedFilters);
```

This will then function as you've described without having to force Angular CD to run via a timeout.

src/app/space/settings/areas/areas.component.html
jiekang: Actually, the `showTree` can be removed entirely as it isn't used after the change below.

src/app/space/settings/areas/areas.component.ts
jiekang: Actually I think I misunderstood your intention.

The current behavior will expand 'all' unexpanded areas in the tree whenever an area is added anywhere. Is that the intention?

jiekang: Well, regardless, I don't think expanding all areas in the tree when I add an area is good behavior. It should at most expand the parent areas that my new area is added under to guarantee my new area is visible, or not expand at all. The former will require quote some changes to the code, the latter can be done with the suggested code above.

src/app/space/settings/areas/areas.component.html
dlabrecq: Sorry, but we cannot remove the timeout and achieve the desired functionality. 

This forces a change detection so that the existing expansion toggles can be rendered properly. If you remove the timeout, the area is still added, but the toggles don't function the way I described.

jiekang: Can you elaborate on 'existing expansion toggles can be rendered properly'?

src/app/space/settings/areas/areas.component.ts
dlabrecq: Removed the functionality

dlabrecq:
Removed expanded toggles and rebased.

jiekang:
Approved in the context of Hierarchical areas PR. @dlabrecq do you know if there are plans for further work on this component? Some issues come to mind.

1. This probably applies to other pages, but it's unfortunate that state isn't saved. It's more obvious now that it's a hierarchical list, but if I've gone to the trouble of hiding/showing various elements, revisiting the page resets it to the default. A fix for this doesn't immediately come to mind.
2. This definitely applies to other data elements: it's unfortunate that the components do not poll for updates; whenever there are third party changes to backend data that is being shown, the component will not update unless a user manually forces a refresh. The most obvious which has caused user confusion before is the codebase widget and adding an application via the space wizard. The codebase widget doesn't show the new codebase, unless the user refreshes by refreshing the page or visiting another page and then going back. This applies here too, the component makes a single http request in onInit for the data, then it's in the dark from there.

fabric8cd:
@dlabrecq fabric8/fabric8-ui:SNAPSHOT-PR-2638-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2638-fabric8-ui.badger.fabric8.io

dlabrecq:
As far as future work for the areas component, the back-end APIs do not provide descriptions, nor any way to delete an area, or even change its name -- that would be a good place to start. Currently, the only way to remove an area is to delete the space.

For the issues regarding application state, I believe getting Redux in place will go a long way to help manage that.

joshuawilson:
@jiekang I think that once we have ngrx setup, we can store state there.

PR #2646, opened 2018-03-18
fix(profile): render correct user data when accessing another profile while logged in

rohitkrai03:
Related Issues -  [OSIO](https://github.com/openshiftio/openshift.io/issues/2613)   [fabric8-auth](https://github.com/fabric8-services/fabric8-auth/issues/389) 

- Fixes rendering of user data when accessing another user's profile.
- Fixes change in context when navigating to another user's profile from within the application.
- Fixes change in spaces section data when switching user context.
- Hides `Update Profile` button when viewing another user's profile.
- Hides `Add Email` option when viewing another user's profile.
- Toggles visibility of connected accounts section when switching user context.
- Adds a new var `_currentContextUser` in context.service.ts to track username in current context because extracted username from router was not working when switching context within the application.


fabric8cd:
@rohitkrai03 fabric8/fabric8-ui:SNAPSHOT-PR-2646-1 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2646-fabric8-ui.badger.fabric8.io

src/app/profile/overview/overview.component.html
```diff
@@ -27,7 +27,7 @@ <h3 class="col-sm-12 profile-wrapper-username">{{context?.user?.attributes?.full
           </p>
         </ng-template>
         <ng-template #addEmail>
-          <p class="col-sm-12 profile-wrapper-email">
+          <p class="col-sm-12 profile-wrapper-email" *ngIf="!context.user.attributes.emailPrivate">
```
dgutride: This is problematic because there are no elvis operators on the properties - try logging out on this page and watch the console to see what I mean.  The context user pushes an empty object so this will cause failures,

rohitkrai03: @dgutride I could reproduce any console errors on my local machine or the cd snapshot. Can you please explain what problems or errors could happen because of this change?

rohitkrai03: @dgutride Since context is always set with user data in overview.component.ts even if viewing a user page without login, do we need elvis operators on properties? On the same page i can see a few places that we've used elvis operators and on the others not. Is there a reason for that? Do let me know if I am missing something or some corner cases where context is not defined which will throw console errors.

dgutride: @rohitkrai03  - looks like you are correct in this instance, the context.user is now using the new property I added last month called currentLoggedInUser.  If it was pointing at loggedInUser from the subscription above, you'd need to protect that value or it would fail.  I'll approve this one.

fabric8cd:
@rohitkrai03 fabric8/fabric8-ui:SNAPSHOT-PR-2646-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2646-fabric8-ui.badger.fabric8.io

Issue #270, opened 2017-03-12
ContextCurrentUserAuthGuard is hanging because the context is not loaded yet


pmuir:
Reproduce by:

1) Logging out
2) visiting `/<username>/_settings`
3) Observe spinning wheel

This is because we have a race between the context resolver and this guard

PR #2650, opened 2018-03-20
patternfly-ng: import individual modules 

dlabrecq:
Taking advantage of patternfly-ng's individual modules so fabric8-ui imports only what’s used in the UI.

Fixes: https://github.com/openshiftio/openshift.io/issues/270
Fixes: https://github.com/openshiftio/openshift.io/issues/2499

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -9,7 +9,8 @@ import { Broadcaster } from 'ngx-base';
 import { BsModalService } from 'ngx-bootstrap/modal';
 import { Spaces } from 'ngx-fabric8-wit';
 import { AuthenticationService } from 'ngx-login-client';
-import { ActionConfig, EmptyStateConfig } from 'patternfly-ng';
+import { ActionConfig } from 'patternfly-ng/action';
+import { EmptyStateConfig } from 'patternfly-ng/empty-state';
 
 import { OnLogin } from '../a-runtime-console/index';
 import { FeatureFlagConfig } from './models/feature-flag-config';
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -30,14 +30,11 @@ import {
   UserService
 } from 'ngx-login-client';
 import { WidgetsModule } from 'ngx-widgets';
-import {
-  ActionModule,
-  ChartModule,
-  EmptyStateModule,
-  ListModule,
-  PatternFlyNgModule,
-  TreeListComponent
-} from 'patternfly-ng';
+
+import { ActionModule } from 'patternfly-ng/action';
+import { EmptyStateModule } from 'patternfly-ng/empty-state';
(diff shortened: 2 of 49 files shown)


dlabrecq:
@joshuawilson This build has been running 8+ hours. Can we kill/restart it?

fabric8cd:
@dlabrecq fabric8/fabric8-ui:SNAPSHOT-PR-2650-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2650-fabric8-ui.badger.fabric8.io

jiekang:
This change confuses me. Would anyone be so inclined to explain the beneficial effects behind this? I'd like to learn.

When I look at other projects that export js/ts modules, it looks common to have a single top-level index.js/index.d.ts that contains the exports. And after reading [1], [2] and [3], if anything, this change requires more files to be read during compile time and extra disk i/o is never great. It also means the caching effect described in [3] is less beneficial. The notes are pretty vague though and I couldn't find any resources on best-practices around this. The statement 'Taking advantage of patternfly-ng's individual modules so fabric8-ui imports only what’s used in the UI.' doesn't seem to follow after reading the documentation listed.

For what it's worth, the listing of issue #270 seems to be a typo

[1] https://www.typescriptlang.org/docs/handbook/modules.html
[2] https://www.typescriptlang.org/docs/handbook/module-resolution.html
[3] https://webpack.js.org/concepts/module-resolution/

christianvogt:
If tree shaking is failing then this change would help with bundle sizing by importing specifically only the modules you use.

I would suggest doing this with a simple webpack plugin. So we don't have to monitor how devs write their imports. I've done it before with babel. Not sure how much different it would be with typescript.

dlabrecq:
This allows fabric8-ui to import only the modules it uses. That is, instead of pulling in an unused dependency like ngx-datatable, angular-tree-component, etc. You can still pull in modules via the top level index.js, but you'll need include unused patternfly-ng dependencies.

jiekang:
My understanding of typescript module resolution along with webpack after reading 1-3 is if a Typescript import is declared like:

```
import { A } from "x"
```

Module resolution will find file "x" to search for "A". If "x" exports "A" from some other file, it will then go 'down' the tree to find the file that actually contains the definition of "A", and once it reaches that, use that definition for compilation. This is all compile time and there is no concept of 'importing unused dependencies' in this step. Doing otherwise sounds like a terrible flaw in a compilation system.

Is webpack actually going to pull in extra when building it's bundles? Is it really leaving such a thing for the developer to solve? Can you clarify with an example?

jiekang:
So after some more reading about webpack, Christian's explanation makes sense. If tree shaking is not enabled or does not work, then this will improve bundle size. I understand now thanks!

PR #2651, opened 2018-03-20
fix(deployments): cancel timeseries queries after deletion

andrewazores:
Cancel future queries for timeseries data for a deployment after the deployment is observed to have
disappeared (ie. via deletion)

fixes https://github.com/openshiftio/openshift.io/issues/2663

The meat of the fix is the chunk that switches `getStreamingTimeseriesData` from using `pollTimer` as its trigger to using `getApplicationsResponse` and adds the `.takeUntil()` operator. These two small changes are what ensure that the timeseries requests happen strictly after the applications requests within each poll cycle, and that the timeseries request is canceled if the deployment has disappeared. The rest of the removals remove now-redundant handling for the "not deployed" case. The previous logic only considered the possibility of the deployment not existing at the time of the initial request, but this check is no longer required if the deployment's existence is checked (using already acquired information) before each individual request. Removing these redundant checks for deployment presence also made testing a bit easier.

### DIFF
--- a/src/app/space/create/deployments/services/deployments.service.spec.ts
+++ b/src/app/space/create/deployments/services/deployments.service.spec.ts
@@ -991,12 +991,14 @@ describe('DeploymentsService', () => {
       });
 
       svc.getDeploymentCpuStat('foo-space', 'foo-app', 'foo-env', 3)
+        .first()
         .subscribe((stats: CpuStat[]) => {
           expect(stats).toEqual([
             { used: 1, quota: 3, timestamp: 1 },
             { used: 2, quota: 3, timestamp: 2 },
             { used: 9, quota: 3, timestamp: 9 }
           ]);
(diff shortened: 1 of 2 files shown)


jiekang:
I'm curious now what the behavior of the backend/frontend is like when a user deletes a deployment for an application and then recreates it again. Hopefully everything still works without requiring a page refresh. I'll look into it when I get a few cycles.

andrewazores:
On the frontend, once the deployment has been deleted and disappears from the applications response, the service effectively forgets about it. When it is recreated it'll be treated as a new instance and a new Observable will be created with a fresh state, and it will go through the same cycle as before. In manual testing this my workflow included two different applications, and I repeatedly deleted deployments from an application and recreated them by restarting the pipeline on the Create > Pipelines page. Once the deployments came back they continued working as expected until deleted again.

PR #2660, opened 2018-03-22
fix(deployments): always show resource card

jiekang:
Fixes https://github.com/openshiftio/openshift.io/issues/2735

The resource cards are always shown. This was actually the initial behavior and then code was added (I think by me) to hide it if there were 0 deployments. The code is removed.


fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2660-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2660-fabric8-ui.badger.fabric8.io

PR #2663, opened 2018-03-23
refactor(deployments): cleanup: removed unused imports

c-koehler:
fixes https://github.com/openshiftio/openshift.io/issues/2753

### DIFF
--- a/src/app/space/create/deployments/apps/deployment-card.component.spec.ts
+++ b/src/app/space/create/deployments/apps/deployment-card.component.spec.ts
@@ -45,7 +45,6 @@ import {
 } from '../services/deployment-status.service';
 import { DeploymentsService } from '../services/deployments.service';
 import { DeploymentCardComponent } from './deployment-card.component';
-import { DeploymentStatusIconComponent } from './deployment-status-icon.component';
 
 @Component({
   template: '<deployment-card></deployment-card>'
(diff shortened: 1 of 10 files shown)


joshuawilson:
@c-koehler please rebase

src/app/space/create/deployments/models/scaled-memory-stat.ts
```diff
@@ -14,6 +14,7 @@ export class ScaledMemoryStat implements MemoryStat {
   constructor(
     public readonly used: number,
     public readonly quota: number,
+    // FIXME: Currently timestamp unused?
```
andrewazores: I don't think this is unused. `DeploymentsService#getDeploymentMemoryStat` returns `ScaledMemoryStat` objects, and there is at least one subscriber to this that uses the timestamp field in the `DeploymentDetailsComponent`. This component only expects the type to be the MemoryStat interface however, not the specific ScaledMemoryStat class, so this might make the timestamp field appear unreferenced at a glance.

fabric8cd:
@c-koehler fabric8/fabric8-ui:SNAPSHOT-PR-2663-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2663-fabric8-ui.badger.fabric8.io

PR #2666, opened 2018-03-26
Fix - broken fonts and images file names

nurali-techie:
This PR fix broken file names for images and fonts.
Before fix:<some_image_name>.<some_hash>png
After fix:<some_image_name><some_hash>.png

Note, the dot (.) moved before hash value to after hash value.  This issue was came into attention when working on https://github.com/fabric8-ui/fabric8-ui-openshift-nginx/pull/17 where browser cache setting are dependent on file extension in nginx.  Currently, browser cache is NOT dependent on file extension but it's better to fix this broken name issue.


joshuawilson:
@nurali-techie for all UI commits please use the commit syntax. 
https://github.com/fabric8-ui/fabric8-ui#continuous-delivery--semantic-releases

PR #2669, opened 2018-03-26
fix(cleanup): fixed typo on cleanup page (it's -> its)

csheremeta:
On Reset Environment page, changed:

"...reverting your OpenShift.io environment to it's default state."

to:

"...reverting your OpenShift.io environment to its default state."


joshuawilson:
[test]

PR #2673, opened 2018-03-28
Codebases dashboard widget updates immediately after forge wizard completion

andrewazores:
This PR adds a feature to the Codebases widget on the space "Analyze" dashboard so that upon completing the "Add to space" wizard, a newly imported or created codebase will be automatically displayed in the widget rather than requiring the page to be reloaded. This solves the problem of the widget becoming stale whenever the wizard is used and completed by a user within the same Analyze tab, but does not solve the problem of the widget becoming stale if the wizard is completed in a different tab, by a different user, etc., or if a Codebase is added to the Space via some other manner than the wizard.

This required a TODO for adding a Broadcast in the wizard to be completed as well (first commit in the series) - this can be easily removed from this PR so that an owner of the wizard components can implement this differently if that's desired. FWIW the CodebasesComponent also had an existing expectation for the codebaseAdded broadcast to possibly be emitted, which is why I thought this would already exist.

### DIFF
--- a/src/app/space/forge-wizard/abstract-wizard.component.ts
+++ b/src/app/space/forge-wizard/abstract-wizard.component.ts
@@ -1,7 +1,7 @@
 import { EventEmitter, OnInit, Output, ViewChild } from '@angular/core';
 import { FormControl, FormGroup, Validators } from '@angular/forms';
 
-import { Notification, Notifications, NotificationType } from 'ngx-base';
+import { Broadcaster, Notification, Notifications, NotificationType } from 'ngx-base';
 import { Context, Space } from 'ngx-fabric8-wit';
 import {
   ForgeService,
@@ -44,6 +44,7 @@ export abstract class AbstractWizard implements OnInit {
   validation: Promise<boolean>;
 
   constructor(public forgeService: ForgeService,
+              public broadcaster: Broadcaster,
               public codebasesService: CodebasesService,
               public context: ContextService,
               public notifications: Notifications) {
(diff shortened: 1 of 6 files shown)


andrewazores:
@joshuawilson could you recommend a reviewer for this?

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2673-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2673-fabric8-ui.badger.fabric8.io

joshuawilson:
@dlabrecq had worked on it a long time back. @jiekang would be good to review it too.

joshuawilson:
@dgutride @dlabrecq this touches the Context and the new wizard, can you please take a look.

src/app/space/forge-wizard/abstract-wizard.component.ts
```diff
@@ -87,8 +88,12 @@ export abstract class AbstractWizard implements OnInit {
...
     obs.subscribe(
       codebase => {
-        // todo broadcast
-        // this._broadcaster.broadcast('codebaseAdded', codebase);
+        if (codebase.attributes.type === 'git') {
+          codebase.name = codebase.attributes.url.replace('.git', '').replace('git@github.com:', '');
+        } else {
+          codebase.name = codebase.attributes.url;
+        }
+        this.broadcaster.broadcast('codebaseAdded', codebase);
```
dlabrecq: Although this works for the existing/old forge wizard, we also need a solution that works with the new app-launcher. 

In addition to broadcasting here, we also need to listen for the onComplete event generated by the new app-launcher component. Then, broadcast the same event from the create-app and import-app components of app-launcher.

jiekang:
In addition, can the codebases list be polled every minute to refresh it?

src/app/space/forge-wizard/abstract-wizard.component.ts
andrewazores: Okay, I see where this logic would be added, but I'm unsure why it's required. Since the new app-launcher is not just an overlay over the Analyze page but actually routes the user away to the applauncher and back again upon completion, will the widget ever actually receive this broadcast from the app launcher? It appears that the workflow would include the widget being destroyed and then recreated after the user has completed (or exited) the wizard, in which case the onInit logic should already restore the correct state. Is this incorrect?

And if that's how the new app launcher works, then perhaps we can just drop this patch, since the bug is only possible with the old wizard?

dlabrecq: If the widget is being recreated, then it should update properly without a broadcast. Please disregard my comments

PR #2674, opened 2018-03-28
fix(deployments): encode urls

jiekang:
Fixes https://github.com/openshiftio/openshift.io/issues/2814
This increases the defensiveness of the deployments service by encoding the urls for the requests.

Currently OSIO doesn't seem to have a path that allows users to insert special characters into their application names and I imagine the same would occur if/when environments are user created rather than static names (run/stage). Space id is also a UUID. However on the off-case someone manages to got a special character in, we will be protected.

GitHub also replaces all special characters with '-' character so I haven't been able to recreate a scenario where this affects the service.

### DIFF
--- a/src/app/space/create/deployments/services/deployments.service.spec.ts
+++ b/src/app/space/create/deployments/services/deployments.service.spec.ts
@@ -70,7 +70,7 @@ describe('DeploymentsService', () => {
     mockNotificationsService = jasmine.createSpyObj<NotificationsService>('NotificationsService', ['message']);
 
     TestBed.configureTestingModule({
-      imports: [ HttpModule ],
+      imports: [HttpModule],
       providers: [
         {
           provide: XHRBackend, useClass: MockBackend
@@ -923,9 +923,9 @@ describe('DeploymentsService', () => {
                         name: 'stage',
                         pod_total: 2,
                         pods: [
-                          [ 'Running', '1' ],
-                          [ 'Starting', '0' ],
-                          [ 'Stopping', '1' ]
+                          ['Running', '1'],
+                          ['Starting', '0'],
+                          ['Stopping', '1']
                         ]
                       }
                     }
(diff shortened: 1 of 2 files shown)


andrewazores:
Can you add some test case(s) to verify the encoding? The service test helper function `doMockHttpTest` should be able to handle this already if you just use pass arguments to a service function that will intentionally require encoding.

jiekang:
@andrewazores yes will do. Thanks for the tip! I wasn't sure how to go about it (why I marked it originally as WIP)

jiekang:
@andrewazores tests added.

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2674-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2674-fabric8-ui.badger.fabric8.io

src/app/space/create/deployments/services/deployments.service.spec.ts
```diff
@@ -1118,6 +1390,114 @@ describe('DeploymentsService', () => {
         expect(svc.getPodsQuota).toHaveBeenCalledWith('foo-space', 'foo-env', 'foo-app');
       });
     });
+
+    it('should combine timeseries and quota data', (done: DoneFn) => {
```
andrewazores: I think this test's name needs to be updated too

andrewazores: Looks good otherwise

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2674-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2674-fabric8-ui.badger.fabric8.io

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2674-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2674-fabric8-ui.badger.fabric8.io

PR #2676, opened 2018-03-28
chore(environments): Add openshift-prod-preview-cluster.sh

gastaldi:
I copied from `openshift-prod-cluster.sh` and changed the parameters to point to prod-preview


gastaldi:
@joshuawilson confirmed that running the ui without running one of the scripts will point to prod-preview. Closing this PR now

joshuawilson:
Looks like the defaults are already set in the code for using prod-preview.

jiekang:
@gastaldi @joshuawilson 

I actually think having this is still helpful. It's much less confusing for developers who don't know about the defaults.

I'd also prefer the defaults to actually load the contents of this script, if possible. That way there aren't two places that need changing if the prod-preview settings ever need to be updated. Alternatively, no defaults and force users to be aware that the env variables need to be set for things to run and the scripts in the environment folder are usable to do that quickly.

Thoughts?

gastaldi:
@jiekang I agree with you. Reopening it

corinnekrych:
agreed sth I've always wanted.
I think we should remove `https://github.com/fabric8-ui/fabric8-ui/blob/master/environments/openshift-free-int-cluster.sh`this is the one i use to use for prod-preview.

fabric8cd:
@gastaldi fabric8/fabric8-ui:SNAPSHOT-PR-2676-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2676-fabric8-ui.badger.fabric8.io

PR #2683, opened 2018-04-02
ngx-forge upgraded to 0.0.66-development

jyasveer:
This upgrade contains the following changes
1. Set default values for artifactid , groupId and version
2. Updates to the summary screen in the import flow as per UXD
3. Handled inconsistency in selection of mission and runtime after redirecting from github step
4. Updates Runtime and Mission screen as per UXD


jyasveer:
@joshuawilson can you please take care of this. Thanks

PR #2685, opened 2018-04-03
Fix pipelines console link to be acquired dynamically

jiekang:
Fixes https://github.com/openshiftio/openshift.io/issues/2229

Adds a service to acquire openshift console url from WIT /api/user/services and uses it. Tested with account in 2 and account in 2a.

### DIFF
--- a/src/app/space/create/pipelines/services/pipelines.service.spec.ts
+++ b/src/app/space/create/pipelines/services/pipelines.service.spec.ts
@@ -309,4 +309,70 @@ describe('Pipelines Service', () => {
       );
   });
 
+  describe('PipelinesService - OK calls', () => {
+    const userServiceResponse = {
+      'data': {
+        'attributes': {
+          'created-at': '2017-05-11T16:44:56.376777Z',
+          'namespaces': [{
+            'cluster-app-domain': '8a09.starter-us-east-2.openshiftapps.com',
+            'cluster-console-url': 'https://console.starter-us-east-2.openshift.com/console/',
+            'cluster-logging-url': 'https://console.starter-us-east-2.openshift.com/console/',
+            'cluster-metrics-url': 'https://metrics.starter-us-east-2.openshift.com/',
+            'cluster-url': 'https://api.starter-us-east-2.openshift.com/',
+            'created-at': '2017-05-11T16:44:56.561538Z',
+            'name': 'blob',
+            'state': 'created',
+            'type': 'user',
+            'updated-at': '2017-05-11T16:44:56.561538Z',
+            'version': '1.0.91'
+          }]
+        },
+        'id': 'eae1de87-f58f-4e67-977a-95024dd7c6aa',
+        'type': 'userservices'
+      }
+    };
+
+    var subscription: Subscription;
+
+    beforeEach(() => {
+      this.subscription = mockBackend.connections.subscribe((connection: MockConnection) => {
+        connection.mockRespond(new Response(
+          new ResponseOptions({
+            body: JSON.stringify(userServiceResponse),
+            status: 200
+          })
+        ));
+      });
+    });
+
+    it('should emit response with repeat calls', (done: DoneFn) => {
+      svc.getOpenshiftConsoleUrl()
+        .subscribe(
+          (msg: string) => {
+            expect(msg).toEqual('https://console.starter-us-east-2.openshift.com/console/project/blob/browse/pipelines');
+            done();
+          },
+          (err: string) => {
+            done.fail(err);
+          }
+        );
+
+      svc.getOpenshiftConsoleUrl()
+        .subscribe(
+          (msg: string) => {
+            expect(msg).toEqual('https://console.starter-us-east-2.openshift.com/console/project/blob/browse/pipelines');
+            done();
+          },
+          (err: string) => {
+            done.fail(err);
+          }
+        );
+    });
+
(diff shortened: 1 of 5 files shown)


jiekang:
@catrobson 
This change has a UX Concern: The dropdown button to go to the openshift console is now hidden if the  openshift console url is not available. Thoughts? Another option is to display a 'Openshift Console Unavailable' "button" that isn't clickable.

When available:

console-visible

When unavailable: 

console-invisible

andrewazores:
There was a similar problem with what to do when a menu item is a link that is known to be unavailable/invalid on the Codebases page recently. The solution we took there was to keep the menu item present and with the same label text as normal, but to disable the item so it is unclickable, and to add/modify the tooltip text when hovering over the item to say that the resource is unavailable. I don't know if there are other similar situations elsewhere in the UI, but we should find a common solution for this sort of problem and be consistent.

jiekang:
Oh, I wasn't aware of that. That sounds fine with me; consistency is good!

src/app/space/create/pipelines/services/pipelines.service.ts
```diff
@@ -0,0 +1,97 @@
...
+              return namespace['cluster-console-url'] + 'project/' + namespace.name + '/browse/pipelines';
+            }
+          }
+
+          return '';
+        });
+    }
+
+    return this.consoleObservable;
+  }
```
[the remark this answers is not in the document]
jiekang: The initial goal was to cache and I think you're right, I'd need to use my own ReplaySubject for this. However on further consideration, caching automatically for a request that isn't polling doesn't necessarily make sense. If a client does want another http request to go through, this would not be possible with the current service. My original reason was to help prevent cases where multiple users of the service result in multiple http requests for the same data. On the Angular side, with the way the service is currently being imported, would there be the same instance used throughout the app? Or separate instances? Do you know?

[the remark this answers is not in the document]
andrewazores: You've given it a component-level provide, so I believe there will be one service instance for each PipelinesComponent, and the service will share the same lifecycle as the component. If you provide the service at the module level instead then there will be one long-lived service instance that is injected into all consumers.

src/app/space/create/pipelines/services/pipelines.service.spec.ts
```diff
@@ -309,4 +309,62 @@ describe('Pipelines Service', () => {
...
+        })
+      ));
+    });
+
+    svc.getOpenshiftConsoleUrl()
+      .subscribe(
+        (msg: string) => {
+          expect(msg).toEqual('https://console.starter-us-east-2.openshift.com/console/project/blob/browse/pipelines');
+          subscription.unsubscribe();
+          done();
```
andrewazores: Is this test case working as intended? You're calling `done` and unsubscribing the mock connection handler in both subscription blocks, which seems wrong.

jiekang: Hmm good point. I'm not sure how I can get an unsubscribe in at the end of the test without encapsulating this in a describe with before/after Each/All. Any ideas?

jiekang: I've gone with that solution for now. If you have a better idea I'd be happy to incorporate it!

src/app/space/create/pipelines/services/pipelines.service.spec.ts
```diff
@@ -309,4 +309,70 @@ describe('Pipelines Service', () => {
...
+        },
+        'id': 'eae1de87-f58f-4e67-977a-95024dd7c6aa',
+        'type': 'userservices'
+      }
+    };
+
+    var subscription: Subscription;
+
+    beforeEach(() => {
+      this.subscription = mockBackend.connections.subscribe((connection: MockConnection) => {
```
andrewazores: `this.subscription` doesn't refer to the `subscription` variable you just declared above - you're assigning a new property on the test context object here.

src/app/space/create/pipelines/services/pipelines.service.spec.ts
```diff
@@ -309,4 +309,70 @@ describe('Pipelines Service', () => {
...
+          (err: string) => {
+            done.fail(err);
+          }
+        );
+
+      svc.getOpenshiftConsoleUrl()
+        .subscribe(
+          (msg: string) => {
+            expect(msg).toEqual('https://console.starter-us-east-2.openshift.com/console/project/blob/browse/pipelines');
+            done();
```
andrewazores: You're still calling `done()` in both of these service subscription handlers, and I think it should only be called once. Maybe you can get around this like so:

```
svc.getOpenshiftConsoleUrl()
  .subscribe(msg1 => {
    expect(msg1).toEqual('foo');
    svc.getOpenshiftConsoleUrl()
      .subscribe(msg2 => {
        expect(msg2).toEqual('bar');
        done();
      });
    });
```

src/app/space/create/pipelines/services/pipelines.service.spec.ts
jiekang: Should be addressed now along with the multiple done() calls. Thanks for your review!

joshuawilson:
[test]

joshuawilson:
@jiekang please rebase. I can't get it to update and build.

PR #2686, opened 2018-04-03
Preserve visited URL when routing to _error (404) state

andrewazores:
fixes https://github.com/openshiftio/openshift.io/issues/2731

This patch adds a small handler to replace the browser's location/state with the attempted URL rather than the path to the _error state, so that the user can more easily see or report what URL they tried to visit and which failed. The approach taken is to replace the browser's location state using the Angular Location service since all other approaches tried using various router functions still resulted in the browser location state being updated for the routed _error state, or resulted in the URL remaining what it was before an invalid link was clicked. Additionally the HeaderComponent is modified to check if the current router state is _error, and if so, to return a falsy (in this case, null) Context, so that the header component template is not displayed while in the error state.

### DIFF
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -1,0 +1,108 @@
+import { Location } from '@angular/common';
+import {
+  Component,
+  NO_ERRORS_SCHEMA
+} from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+import { Router } from '@angular/router';
+import { RouterTestingModule } from '@angular/router/testing';
+
+import { createMock } from 'testing/mock';
+import {
+  initContext,
+  TestContext
+} from 'testing/test-context';
+
+import { Broadcaster } from 'ngx-base';
+import { BsModalService } from 'ngx-bootstrap/modal';
+import { Spaces } from 'ngx-fabric8-wit';
+import { AuthenticationService } from 'ngx-login-client';
+import {
+  Observable,
+  Subject
+} from 'rxjs';
+
+import { OnLogin } from 'a-runtime-console';
+import { AboutService } from './shared/about.service';
+import { ProviderService } from './shared/account/provider.service';
+import { AnalyticService } from './shared/analytics.service';
+import { BrandingService } from './shared/branding.service';
+import { LoginService } from './shared/login.service';
+import { NotificationsService } from './shared/notifications.service';
+
+import { AppComponent } from './app.component';
+
+@Component({
+  template: '<f8-app></f8-app>'
+})
+class HostComponent { }
+
+@Component({
+  template: ''
+})
+class MockHomeComponent { }
+
+@Component({
+  template: ''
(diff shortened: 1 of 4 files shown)


andrewazores:
@joshuawilson recommend a reviewer?

src/app/app.component.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+import { Broadcaster } from 'ngx-base';
+import { BsModalService } from 'ngx-bootstrap/modal';
+import { Spaces } from 'ngx-fabric8-wit';
+import { AuthenticationService } from 'ngx-login-client';
+import {
+  Observable,
+  Subject
+} from 'rxjs';
+
+import { OnLogin } from '../a-runtime-console/index';
```
joshuawilson: this can be shortened to just `import { OnLogin } from './a-runtime-console/index';`

andrewazores:
@joshuawilson did you do both of these by entering those URLs directly in the browser's address bar?

1. is a result of how the routing is set up. `_homeA` is not a path listed in app-routing and so it will be interpreted as an `:entity` and passed to the `ContextResolver`. This will then fail because the context name contains the reserved character `_`. This behaviour is the same even before the patch in this PR  is applied. Since this flow does not include the 404 page being hit at all (although perhaps it should?) I intentionally did not touch it with this PR.

2. Similar to 1, this is because of routing. `josh-6/my-app` is interpreted with `josh-6` as the entity and `my-app` as the space. These are then passed again to the `ContextResolver`. If the entity is correct but the space does not exist, the guard will again fail, but differently - rather than returning `false` because of the reserved character, it will emit an error on the Observable result due to the HTTP error when trying to access the nonexistent space. This error results in a NavigationError event on the router, which falls into the errorHandler, and we're now looking at https://github.com/openshiftio/openshift.io/issues/2860 . Again, this behaviour is the same even before this patch is applied, and I intentionally did not touch it.

I should have documented this in the initial comment on the PR, sorry. This patch will only affect 404s generated by accessing totally invalid application paths - ones that the router has no match for at all. For example, `/nonexistent/app/path` as used in the unit tests. The router will be unable to match this path to any of its path matchers, so it will fall into the `**` matcher, which redirects to the `_error` state, and will trigger the behaviour added in this patch.

If we're looking at changing the behaviour of the `ContextResolver` to also redirect the user to the 404 page when an invalid user or space is provided then I'd be happy to write a patch for that as well. Do you think this is within the scope of the original issue or should a new one be opened?

joshuawilson:
I'm fine with doing the rest of the work in a new PR.

andrewazores:
@joshuawilson I think this one is ready to go. It still only handles cases where there is no routing match, eg. `/nonexistent/app/path`. The other WIP PR I have open (https://github.com/fabric8-ui/fabric8-ui/pull/2695) is the one where the behaviour is changed for other routing cases, and still needs UX verification.

PR #2689, opened 2018-04-04
fix(version): update ngx-fabric8-wit to 6.20.2

fabric8cd:


sanbornsen:
This should not go without the respective changes in the platform. Hence closing it. Will add this change with the respective PR.

PR #2692, opened 2018-04-04
Dummy commit to test PR builds

jaseemabid:


joshuawilson:
If you still need it, reopen. Since we are going to try to move away from fabric8 build, this is not needed.

alien-ike:
### Ike Plugins (test-keeper)

Thank you @jaseemabid for this contribution!



If you are an admin and you are sure that no test is needed then you can use the command `/ok-without-tests` as a comment to make the status green.

PR #2693, opened 2018-04-04
fix(launcher): updates overlay style as per ux, swaps Import card wit…

invincibleJai:
### Launcher 

- updates overlay style and texts as per ux
- Introduced radio selection with Continue button  
- swaps Import card with Quickstarts card in overlay screen

https://redhat.invisionapp.com/share/VGF1RF15J#/screens/286894927


invincibleJai:
Thanks @mindreeper2420, have updated pls take a look
- Scroll in overlay
- handles  content crashes into the other columns
- title is set to line up "Create a new codebase" and "import an existing codebase"

mindreeper2420:
Looks good! Thanks for the updates.

PR #2695, opened 2018-04-05
Consistently route to 404 page when visiting invalid links

andrewazores:
https://github.com/openshiftio/openshift.io/issues/2887
https://github.com/openshiftio/openshift.io/issues/2860

~~This is a work in progress for prototyping 404/_error page routing work. The homepage component gains several "error n" links in the banner under "Setup your profile", which exercise various types of failure modes with different sorts of invalid URLs. These can be clicked to manually test the behaviour of the prototype when visiting an invalid URL. Once on that page, the browser can be refreshed to also simulate the behaviour of entering that invalid URL directly into the address bar, as opposed to visiting it from within the running application.~~

This PR modifies the context resolution and routing behaviour so that the behaviour of an invalid application link, including links to invalid User or Space contexts, consistently routes the user to the 404 page (`_error` state), with the attempted URL preserved for the user in their browser's address bar. See the [demo](https://github.com/openshiftio/openshift.io/issues/2887#issuecomment-380915894) - the notification is no longer displayed, but the routing behaviour is the same as shown.

### DIFF
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -3,7 +3,11 @@ import {
   Component,
   NO_ERRORS_SCHEMA
 } from '@angular/core';
-import { TestBed } from '@angular/core/testing';
+import {
+  fakeAsync,
+  TestBed,
+  tick
+} from '@angular/core/testing';
 import { Router } from '@angular/router';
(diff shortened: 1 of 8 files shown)


joshuawilson:
I just tried it locally, this works very nicely. Well done.

PR #2696, opened 2018-04-05
fix(launcher): updated anchor text for launcher experience

invincibleJai:
### Launcher 

Updates anchor text to trigger new Launcher flow.

screen shot 2018-04-05 at 11 34 03 am

Issue : https://github.com/openshiftio/openshift.io/issues/2832

### DIFF
--- a/src/app/space/forge-wizard/components/flow-selector/flow-selector.component.html
+++ b/src/app/space/forge-wizard/components/flow-selector/flow-selector.component.html
@@ -47,7 +47,7 @@
         <button class="btn btn-link"
               type="button"
               (click)="showAddAppOverlay(); cancel();"
-              aria-label="Open New Import Experience">New Import Experience</button>
+              aria-label="Open New Launcher Experience">Try out the new Launcher experience</button>
       </div>
     </div>
   </div>


src/app/space/forge-wizard/components/flow-selector/flow-selector.component.html
```diff
@@ -47,7 +47,7 @@ <h3>Create a new <b>Quickstart</b> project</h3>
         <button class="btn btn-link"
               type="button"
               (click)="showAddAppOverlay(); cancel();"
-              aria-label="Open New Import Experience">New Import Experience</button>
+              aria-label="Open New Launcher Experience">Try out new Launcher experience</button>
```
joshuawilson: The wording does not feel natural. @catrobson can you please review

catrobson: @stacymcauliffe can you review?

stacymcauliffe: How about: 
We're introducing a new way to create applications. Try it out!

@catrobson @joshuawilson

invincibleJai: @joshuawilson @catrobson  had a discussion with @stacymcauliffe on launcher channel, we have two suggestions : 
- "We're introducing a new way to create applications. Try it out!" this would make sense to the general user
- "Try out new Launcher experience" being an internal feature it would be accessible to Redhat folks so they may relate better to "New Launcher" terminology.

joshuawilson: My issue is that "Try out new Launcher experience" is not natural. It would be better as:
 - "Try out **the** new Launcher experience"
 - "Try ou**r** new Launcher experience" 

The second point is that Launcher is an unknown term. So the generic wording is better to me.

invincibleJai: Thanks @joshuawilson , as suggested will go with "Try out the new Launcher experience" now and i'll get it verified with Stevan as well.

PR #2701, opened 2018-04-06
fix(launcher): removes isPageRedirect method as not needed

invincibleJai:
### Launcher 

- removes isPageRedirect method as not needed in app-launcher-gitprovider.service.ts


dlabrecq:
LGTM

PR #2709, opened 2018-04-09
fix(version): update fabric8-planner to 0.43.11

nimishamukherjee:
fix(version): update fabric8-planner to 0.43.11


joshuawilson:
Do not Push to prod until this is fixed: https://github.com/openshiftio/openshift.io/issues/2979

PR #2710, opened 2018-04-09
feat(deployment-api): introduce DeploymentApiService

andrewazores:
Add a new DeploymentApiService which simply wraps Deployments WIT endpoints, for non-Deployments
page UI consumers.

fixes https://github.com/openshiftio/openshift.io/issues/2924


jiekang:
What was the reasoning behind not including the changes here (possibly a separate commit) to make the DeploymentsService use this new DeploymentsApiService?

andrewazores:
@jiekang because @chrislessard has a PR that was just opened that should also be a consumer of the DeploymentApiService, so I wanted to get this service out and available for him to use as quickly as possible. I figured having multiple consumers will also help ensure that the service is generalized enough rather than remaining oriented toward the Deployments page, so getting it out early it will help guard against moving it in that direction.

PR #2712, opened 2018-04-09
feat(deployments): tests added for deleting deployment

c-koehler:
fixes https://github.com/openshiftio/openshift.io/issues/2818

### DIFF
--- a/src/app/space/create/deployments/services/deployments.service.spec.ts
+++ b/src/app/space/create/deployments/services/deployments.service.spec.ts
@@ -1842,6 +1842,59 @@ describe('DeploymentsService', () => {
     });
   });
 
+  describe('#deleteDeployment', () => {
+    it('should delete a deployment with the correct URL', (done: DoneFn) => {
+      const spaceId = 'someSpace-Id';
+      const environmentId = 'some Stage';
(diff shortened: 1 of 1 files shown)


src/app/space/create/deployments/services/deployments.service.spec.ts
```diff
@@ -1842,6 +1842,56 @@ describe('DeploymentsService', () => {
     });
   });
 
+  describe('#deleteDeployment', () => {
+    const spaceId = 'someSpaceId';
+    const environmentId = 'someStage';
+    const appId = 'someAppName';
```
andrewazores: Could you move these const declarations inside the `it` block as well? I think it's acceptable that this means both `it` blocks will have duplicate declarations. And perhaps you can modify them so that their values will be modified after being run through `encodeURIComponent`, by adding a space character for example.

c-koehler:
@andrewazores I changed it to some having spaces, another having a dash. Hopefully that covers it, if not let me know.

andrewazores:
@c-koehler okay, that works. FWIW the dash doesn't get altered by URI encoding though, but the spaces will.

PR #2713, opened 2018-04-10
fix(stylelint): fail prod and dev build on stylelint errors

mindreeper2420:
- Set the build and local development environments to fail when there are stylelint errors.
- Fix errors so build passes


mindreeper2420:
@joshuawilson here is the PR to set the production and development builds to fail on stylelint errors. I also fixed the 26 errors that made it into production code.

PR #2718, opened 2018-04-10
fix(webpack): add dev server with increased keepAliveTimeout

christianvogt:
Adding an alternative dev server which increases the keepAliveTimeout to prevent errors with serving very large dev JS files that take longer than 5 seconds resulting in the following error:
`http://localhost:3000/vendor.bundle.js net::ERR_CONTENT_LENGTH_MISMATCH`

Use the following command to run the dev server:
`npm run server:dev2`

The timeout is increased by default from 5 seconds to 10 seconds and can further controlled via argument in `package.json` script or in `devserver.js`.


alien-ike:
### Ike Plugins (test-keeper)

If you are an admin and you are sure that no test is needed then you can use the command `/ok-without-tests` as a comment to make the status green.


corinnekrych:
@christianvogt 👍 let me merge, I've also added to our how-to wiki page a section about it - copy/pasting your description ;)

PR #2725, opened 2018-04-11
test(feature-flag): add error check

corinnekrych:
fixes https://github.com/openshiftio/openshift.io/issues/2400  for error cases

### DIFF
--- a/src/app/feature-flag/notification-footer/feature-footer.component.spec.ts
+++ b/src/app/feature-flag/notification-footer/feature-footer.component.spec.ts
@@ -1,0 +1,90 @@
+import { Component } from '@angular/core';
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { ModalModule } from 'ngx-bootstrap';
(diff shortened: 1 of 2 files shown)


src/app/feature-flag/notification-footer/feature-footer.component.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+      imports: [FeatureFooterModule, TooltipModule, ModalModule.forRoot()],
+      declarations: [TestHostComponent],
+      providers: []
+    });
+
+    hostFixture = TestBed.createComponent(TestHostComponent);
+    hostComponent = hostFixture.componentInstance;
+  });
+
+  fit('should render internal icon when user level is internal', async(() => {
```
aptmac: This should probably be changed to 'it'? And again at line 82.

corinnekrych: oops yes indeed!

PR #2732, opened 2018-04-12
Email Verification - Fix image in build and link to dashboard

SMahil:
Fix image in build and link to dashboard


alien-ike:
### Ike Plugins (test-keeper)

Thank you @SMahil for this contribution!



rohitkrai03:
@SMahil Code looks good to me. But i found a related but different bug that's there on prod as well. While redirecting to some page from an error page when there is some error param in URL, the redirect doesn't work for some reason. 

So, clicking the take me home button on https://openshift.io/_error works fine but for https://openshift.io/_error?error=some_error it doesn't.

cc: @joshuawilson

joshuawilson:
@rohitkrai03 do you think we could add a unit test to check for that?

rohitkrai03:
@joshuawilson Adding a test might be a bit tricky here because while debugging i could see the router is redirecting the URL but the desired component somehow fails to render. So, i am not sure how to reproduce this thing in unit test. 
Also, i've created an issue related to this - https://github.com/openshiftio/openshift.io/issues/3144

PR #2737, opened 2018-04-12
fix(flow-selector): revert changes to the space attribute of flow-sel…

aptmac:
…ector

This PR fixes https://github.com/openshiftio/openshift.io/issues/3075 [0].

It reverts the changes to 5 html components [1] where `[space]="space"` was changed to become `[space]=space?.attributes.name` when I (wrongfully) noticed that the space attribute expected a string type instead of a space type. As such, based on a recommendation the space type has been changed to `any`for the moment to remove any further confusion.

[0] https://github.com/openshiftio/openshift.io/issues/3075
[1] https://github.com/fabric8-ui/fabric8-ui/pull/2730/files


corinnekrych:
+1

PR #2691, opened 2018-04-04
feat(feature-flag): implement sticky footer round button for componen…

corinnekrych:
…t features
Fixes https://github.com/openshiftio/openshift.io/issues/2400
@mindreeper2420 From the issue, you can find the link to the screens: https://redhat.invisionapp.com/share/KAF97RBGB#/screens/276638411_Options_Selection

PR #2738, opened 2018-04-12
fix(codebases): enable backaction outlet in routing

corinnekrych:
fixes https://github.com/openshiftio/openshift.io/issues/3100

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -10,6 +10,7 @@
 
 <main>
   <div>
+    <router-outlet name="action"></router-outlet>
     <router-outlet></router-outlet>
     <f8-feature-footer [featurePageConfig]="featureConfig"></f8-feature-footer>
   </div>


joshuawilson:
This doesn't completely solve it.

aptmac:
@joshuawilson what functionality doesn't get resolved? additionally, it resolves the add-work-item action issue for me too, as seen in https://github.com/openshiftio/openshift.io/issues/3021 [0].

[0] https://github.com/openshiftio/openshift.io/issues/3021

joshuawilson:
If you click the plus for codebases, after the page changes, click on pipelines. The URL keeps the codebases URL with an ending like this `create/(pipelines//action:add-codebase)`.

We need to find the root cause.

dlabrecq:
I build/ran this branch locally, but get a blank home page. I added name="action" manually, to a clean master branch, and encountered the same issue. 

I can see the URL change when attempting to navigate to my profile or spaces, but still get a blank page.

I don't see any errors in the console, but believe the feature flags is preventing the home page from displaying?

joshuawilson:
I was able to trace the cause to https://github.com/fabric8-ui/fabric8-ui/pull/2691 and it appears this fix puts things back to the way they were.

PR #2741, opened 2018-04-13
fix(launcher):  CSS issue with overlay screen

invincibleJai:
### Launcher

- CSS issue with the overlay screen

Tracks : 
- https://github.com/openshiftio/openshift.io/issues/3050


mindreeper2420:
Will this fix the screen running in OSIO, or just the demo app? I believe the Create overlay is part of fabric8-ui?

PR #2745, opened 2018-04-13
Create Space Overlay: Enable space name validation

dlabrecq:
The create space overlay already has code to validate the space name. However, it didn’t include necessary directives. The forge wizard already declares UniqueSpaceNameValidatorDirective and ValidSpaceNameValidatorDirective, so that should be declared in the create space overlay since the forge wizard is being replaced.

Fixes: https://github.com/openshiftio/openshift.io/issues/3055


alien-ike:
### Ike Plugins (test-keeper)

Thank you @dlabrecq for this contribution!

PR #2754, opened 2018-04-17
fix(pipelines): handles routes not appearing for stage/prod

invincibleJai:
handles routes not appearing for stage/prod in pipelines screen for new launcher flow

Tracks : 
- https://github.com/openshiftio/openshift.io/issues/3003
- https://github.com/openshiftio/openshift.io/issues/2920

### DIFF
--- a/src/a-runtime-console/kubernetes/model/pipelinestage.model.ts
+++ b/src/a-runtime-console/kubernetes/model/pipelinestage.model.ts
@@ -43,10 +43,11 @@ export class PipelineStage {
             this.environmentName = name;
             let urlMap = se.serviceUrls;
             this.serviceUrlMap = urlMap;
-
-            let buildConfigName = build.buildConfigName;
-            if (buildConfigName) {
-              this.serviceUrl = urlMap[buildConfigName] || '';
+            let serviceUrlKeys: string[] = Object.keys(urlMap);
+            if (serviceUrlKeys && serviceUrlKeys[0]) {
+              this.serviceUrl = urlMap[serviceUrlKeys[0]];
+            } else {
+              this.serviceUrl = '';
             }
           }
         }


alien-ike:
### Ike Plugins (test-keeper)

Thank you @invincibleJai for this contribution!



invincibleJai:
@jiekang can you pls review this, it handles routes not being shown up with new launcher flow if `app_name !== repo_name !== artifact_id`

ravsa:
I've tested this PR with multiple builds and now links are showing on the pipeline screen :)

invincibleJai:
@ravsa thanks for testing :)

src/a-runtime-console/kubernetes/model/pipelinestage.model.ts
```diff
@@ -43,10 +43,9 @@ export class PipelineStage {
             this.environmentName = name;
             let urlMap = se.serviceUrls;
             this.serviceUrlMap = urlMap;
-
-            let buildConfigName = build.buildConfigName;
-            if (buildConfigName) {
-              this.serviceUrl = urlMap[buildConfigName] || '';
+            let servUrlName: string[] = Object.keys(urlMap);
+            if (servUrlName && servUrlName[0]) {
+              this.serviceUrl = urlMap[servUrlName[0]] || '';
```
jiekang: The assumption on the 0 indexed item being the correct service url feels a little dangerous, but after analyzing the code it's not like the rest of it is capable of handling multiple service url's at this time. It would be nice to have a unit test that fails before this change and passes afterwards; you can log some of the data when connected to production so it's not too hard to write.

invincibleJai: @jiekang yes as per my understanding earlier I mean in the legacy wizard as you are aware we had `app_name == repo_name == artifact_id` and here `build.buildConfigName == app_name` and `urlMap: { artifact_id : 'route_stage'}` it used to work as both buildConfigName and key name under urlMap used to be same.
> this.serviceUrl = urlMap[buildConfigName] || '';

now with new launcher flow as these values are not same so I fetched `Object.keys(urlMap)` and as far I've seen it always has just one key and associated value i.e `urlMap: {artifact_id: 'route_stage'}` so used index[0].

And this is my first contribution in pipelines code, in order to fix this issue, can see any associated spec files else would have updated the tests with this scenario.

src/a-runtime-console/kubernetes/model/pipelinestage.model.ts
```diff
@@ -43,10 +43,11 @@ export class PipelineStage {
             this.environmentName = name;
             let urlMap = se.serviceUrls;
             this.serviceUrlMap = urlMap;
-
-            let buildConfigName = build.buildConfigName;
-            if (buildConfigName) {
-              this.serviceUrl = urlMap[buildConfigName] || '';
+            let serviceUrlKey: string[] = Object.keys(urlMap);
```
jyasveer: @invincibleJai Just a small suggestion. if you rename the variable from `serviceUrlKey` to `serviceUrlKeys`, it could be more meaningful.

invincibleJai: @jyasveer have made the changes

joshuawilson:
Finding now tests is not a reason to not write them. Please add the spec and the tests for this change. You don't have to write tests for everything else.

joshuawilson:
This code is wrong. As it is difficult to test in it's current location and structure it should be refactored so that it is testable and maintainable. Please add an issue to fix it and add the reference here.

I understand the value of this fix and the difficulty in testing it. Once the refactoring issue is added, I will merge this one.

fabric8cd:
@invincibleJai fabric8/fabric8-ui:SNAPSHOT-PR-2754-8 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2754-fabric8-ui.badger.fabric8.io

PR #2755, opened 2018-04-17
Refactor EnvironmentWidget to remove DeploymentsService dependency

andrewazores:
This refactors the EnvironmentWidget found on the current Analyze dashboard to remove its dependency on the DeploymentsService, which is intended to be consumed only by Deployments-area components. The function and structures that were previously added to the DeploymentsService to support this widget have been moved into a new, very small service, which is specific to the EnvironmentWidget, and which uses the new DeploymentApiService to handle HTTP requests to the backend Deployments API.

### DIFF
--- a/src/app/dashboard-widgets/environment-widget/application-overview.service.spec.ts
+++ b/src/app/dashboard-widgets/environment-widget/application-overview.service.spec.ts
@@ -1,0 +1,122 @@
+import {
+  fakeAsync,
+  TestBed,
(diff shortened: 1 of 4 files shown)


alien-ike:
### Ike Plugins (test-keeper)

jiekang:
Filed a separate issue found when reviewing this. https://github.com/openshiftio/openshift.io/issues/3173 Just FYI.

PR #2764, opened 2018-04-18
feat(bootstrap): add bootstrap module and update OauthConfigStore

jiekang:
This adds a bootstrap module which provides an `APP_INITIALIZER` promise that Angular will guarantee to wait for before continuing past initialization.

The oauth-config-store is added such that the oauth-config-store http request must complete before continuing past initialization. This fixes https://github.com/openshiftio/openshift.io/issues/2200

As well, the oauth-config-store is refactored to subscribe to the UserService loggedInUser observable to be robust against https://github.com/openshiftio/openshift.io/issues/3161 Tests are added as well.

Note: This is one solution for https://github.com/openshiftio/openshift.io/issues/2200 It doesn't address the fact that the websocket code is written strangely to repeatedly, and infinitely try to connect to some url, even if it fails. It only addresses the reason behind why the connection fails, because the url is sometimes valid and other times invalid. I'm still trying to understand the a-runtime-console code to improve the websockets because if at some point, this url is invalid due to another reason (API changes to somewhere else and it's not updated, someone makes some change without knowing it's full effects, etc.), this issue of repeated websocket connection attempts will pop up again.

This was manually tested with a purposely delayed http connection via:

```
oauth-config-store.ts:
[...]
  private load() {
    let configUri = '/_config/oauth.json';
    this.http.get(configUri).delay(5000)
    [...]
```

the issue is fixed with the patch, not fixed without. A side effect is that, in this instance, the application takes 5 more seconds to load because of the 5000 ms delay and how it must wait for completion before continuing. In practice, it will depend on how fast the user's internet is.

I'm wondering if this initialization delay is worth it. I really think the pipelines work needs to be cleanly refactored out of the a-runtime-console and then we can drop this oauth-config-store as well. This PR could be a 'good enough' fix in the meantime, or maybe not. Thoughts?

### DIFF
--- a/src/a-runtime-console/kubernetes/store/oauth-config-store.spec.ts
+++ b/src/a-runtime-console/kubernetes/store/oauth-config-store.spec.ts
@@ -1,0 +1,283 @@
+import { ErrorHandler } from '@angular/core';
+import { fakeAsync, inject, TestBed } from '@angular/core/testing';
+
+import {
+  Http,
+  HttpModule,
+  Response,
+  ResponseOptions,
+  ResponseType,
+  XHRBackend
+} from '@angular/http';
+
+import {
+  MockBackend,
+  MockConnection
+} from '@angular/http/testing';
+
+import {
+  Observable,
+  Subscription
+} from 'rxjs';
+
+
+import { createMock } from 'testing/mock';
+
+import {
+  Logger,
+  Notification,
+  NotificationType
+} from 'ngx-base';
+import {
+  Profile,
+  User,
+  UserService
+} from 'ngx-login-client';
+
+import {
+  OAuthConfig,
+  OAuthConfigStore
+} from './oauth-config-store';
+
+import { NotificationsService } from 'app/shared/notifications.service';
+
+describe('OauthConfigStore', () => {
+
+  let mockUserService: UserService;
+  let mockBackend: MockBackend;
+  let oauthStore: OAuthConfigStore;
+
+  let mockLogger: jasmine.SpyObj<Logger>;
+  let mockErrorHandler: jasmine.SpyObj<ErrorHandler>;
+  let mockNotificationsService: jasmine.SpyObj<NotificationsService>;
+
+  let user: User = {
+    attributes: {
+      fullName: 'mock',
+      imageURL: 'mock',
+      username: 'mock',
+      cluster: 'http://api.example.com/cluster/'
+    },
+    id: 'mock',
+    type: 'mock'
+  };
+
+  let data = {};
+
+  let subscription: Subscription;
+
+  beforeEach(() => {
+    mockLogger = jasmine.createSpyObj<Logger>('Logger', ['error']);
+    mockErrorHandler = jasmine.createSpyObj<ErrorHandler>('ErrorHandler', ['handleError']);
+    mockNotificationsService = jasmine.createSpyObj<NotificationsService>('NotificationsService', ['message']);
+  });
+
+  describe('success state', () => {
+    beforeEach(() => {
+      mockUserService = createMock(UserService);
+      mockUserService.loggedInUser = Observable.of(user).publish();
+
+      TestBed.configureTestingModule({
+        imports: [HttpModule],
+        providers: [
+          {
+            provide: XHRBackend, useClass: MockBackend
+          },
+          {
+            provide: UserService, useClass: mockUserService
+          },
+          {
+            provide: Logger, useValue: mockLogger
+          },
+          {
+            provide: ErrorHandler, useValue: mockErrorHandler
+          },
+          {
+            provide: NotificationsService, useValue: mockNotificationsService
+          },
+          {
+            // provide OAuthConfigStore with a factory inside the fakeAsync zone
+            // so the mockBackend can catch http requests made inside the constructor
+            provide: OAuthConfigStore, useFactory: fakeAsync((
+              http: Http,
+              mockBackend: MockBackend,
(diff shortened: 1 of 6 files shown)


src/a-runtime-console/kubernetes/store/oauth-config-store.ts
```diff
@@ -104,16 +116,23 @@ export class OAuthConfigStore {
...
+  private load() {
     let configUri = '/_config/oauth.json';
     this.http.get(configUri)
+      .catch((error: Response) => {
+        this.errorHandler.handleError(error);
+        this.logger.error(error);
+        this.notifications.message({
+          type: NotificationType.DANGER,
+          header: 'Error: Configuration setup',
+          message: 'Could not find OAuth configuration at' + configUri
```
andrewazores: Missing a space between "at" and the URI here

src/a-runtime-console/kubernetes/store/oauth-config-store.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+    });
+
+    it('should eventually set openshift console on init', (done: DoneFn) => {
+      mockUserService.loggedInUser.connect();
+
+      oauthStore.loading.subscribe((val: boolean) => {
+        if (!val) {
+          oauthStore.resource.subscribe((config: OAuthConfig) => {
+            expect(config.loaded).toBeTruthy();
+            if (config.openshiftConsoleUrl) {
```
andrewazores: Same idea here with the `config.openShiftConsoleUrl` check - this should probably become an `expect().toBeTruthy()`.

src/a-runtime-console/kubernetes/store/oauth-config-store.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+
+      mockBackend = TestBed.get(XHRBackend);
+      oauthStore = TestBed.get(OAuthConfigStore);
+    });
+
+    it('should load and set latest oauthconfig on init', (done: DoneFn) => {
+      mockUserService.loggedInUser.connect();
+
+      oauthStore.loading.subscribe((val: boolean) => {
+        if (!val) {
```
andrewazores: Does this `if (!val)` check need to be present in this test? The truthiness of `val` should be well known in this test case, and the test is geared to fail by timeout if it's wrong. Perhaps the check can be removed and a `expect().toBeFalsey()` can be added instead, so that if it is *not* as expected the reason for the test failure is more explicitly reported than a vague timeout?

jiekang: I think I did these if statements for resource && loading because they're both BehaviorSubjects so I expected them to emit their initial value to the subscriber and eventually emit the value I want. Looks like I was wrong though; fixed.

joshuawilson:
I guess if I had to pick, working 100% but starting slow is the lesser of 2 evils. Have you tested if you can use less time, maybe 2 secs or 3?

@dgutride can you take a look at this, please.

christianvogt:
lgtm
I agree with @andrewazores and @joshuawilson  above. Not ideal to block the entire UI until the conf is loaded but serves the purpose.

An issue should be raised to track an improvement; be it hydrating the UI with a starting config through a server rendered page or something else.

jiekang:
@joshuawilson Re: "Have you tested if you can use less time, maybe 2 secs or 3?" I think my comment regarding time was unclear. In practice, it will purely depend on your internet connection. For me, it's less than a second. The 5 seconds was part of manual testing, set by me to verify it works with the patch and doesn't work without. It's not in the patch.

joshuawilson:
Thanks @jiekang for clearing that up.

joshuawilson:
local tests fail with 

```
OauthConfigStore
    user service error
      ✖ should notify on user service error
        PhantomJS 2.1.1 (Linux 0.0.0)
      Expected undefined to equal 'http://console.example.com/cluster/console'.
```

jiekang:
@joshuawilson nice catch, thanks! I've added a fix for that. If anyone could verify on their system I'd appreciate it!

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2764-8 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2764-fabric8-ui.badger.fabric8.io

src/a-runtime-console/kubernetes/store/oauth-config-store.spec.ts
```diff
@@ -0,0 +1,286 @@
...
+  Subscription
+} from 'rxjs';
+
+
+import { createMock } from 'testing/mock';
+
+import {
+  Logger,
+  Notification,
+  NotificationType
```
joshuawilson: It appears you are not using these imports.
`Notification, NotificationType`
Nor `Profile` from below.

jiekang: Nice catch as well. I really need to set some time to find an extension that will suggest unused imports or something else that can automatically do that for me.

jiekang: I've removed them with the latest commit.

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2764-9 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2764-fabric8-ui.badger.fabric8.io

PR #2767, opened 2018-04-18
fix(environment-widget): disable links when deployment has 0 pods

andrewazores:
fixes https://github.com/openshiftio/openshift.io/issues/3173

### DIFF
--- a/src/app/dashboard-widgets/environment-widget/application-overview.service.spec.ts
+++ b/src/app/dashboard-widgets/environment-widget/application-overview.service.spec.ts
@@ -34,7 +34,8 @@ describe('ApplicationOverviewService', () => {
             {
               attributes: {
                 name: 'stage',
-                version: '1'
+                version: '1',
+                pod_total: 1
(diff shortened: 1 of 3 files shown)


andrewazores:
Note that this doesn't hide the link as suggested in the bug report, but instead makes the link falsey (null) so that the component template will still display the deployment environment name and version, but make it non-clickable.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2767-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2767-fabric8-ui.badger.fabric8.io

andrewazores:
Huh, I guess Badger's back.

jiekang:
I think Joshua had trouble removing it; I went ahead and checked the logs and filed an issue for some of the errors: https://github.com/fabric8io/fabric8-build-team/issues/53

Looks like it works now.

PR #2768, opened 2018-04-19
fix master build

jarifibrahim:
The master build is failing at chrome installation with "Requires: libappindicator3.so.1()(64bit)"
This PR installs epel-release and libappindicator to the builder container.


jarifibrahim:
Planner builds (PR and Master) were failing because of the same issue. We've merged the patch (https://github.com/fabric8-ui/fabric8-planner/pull/2570)  on fabric8-planner and the builds are working now.

PR #2774, opened 2018-04-19
fix(version): update fabric8-planner to 0.45.0

fabric8cd:


sanbornsen:
Do not merge it yet. We found some bug with infotips.

PR #2779, opened 2018-04-20
LoginService does not "hang" when presented with an "error" URL param

andrewazores:
This PR addresses https://github.com/openshiftio/openshift.io/issues/3144 . The LoginService sees the "error" URL param and sends a notification, but continues to process login logic as expected, so that the user may continue to use OSIO. Otherwise, before the first patch in this series, the LoginService is left in a bad state with an undefined `openshiftToken` property, and with the `authService` `logIn` or `onLogIn` functions not called.

Additionally, in order to make this change testable, I have moved the `WindowService` to the `src/app/shared` directory, and since it is just a wrapper around the global `window` object, I have added it as an application root module level provide and removed it from the three other component modules which previously provided their own instances of the service. Then I refactored the `LoginService` to also use the `WindowService`, so that I could control the query parameter exposed to the `LoginService` without interfering with PhantomJS.


fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2779-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2779-fabric8-ui.badger.fabric8.io

PR #2784, opened 2018-04-23
fix(settings): Redirect user to appropriate feature opt in page

chrislessard:
This patch changes the route used by the feature warning page to redirect users to the feature opt in page under their profile settings, instead of their profile.

Fixes: https://github.com/openshiftio/openshift.io/issues/3260


alien-ike:
### Ike Plugins (test-keeper)

Thank you @chrislessard for this contribution!



joshuawilson:
@chrislessard you know that if the settings pages gets put behind a feature flag then this will need that too?

PR #2792, opened 2018-04-24
fix(pipelines): refactor pipelines service and use it in pipelines widgets

jiekang:
This fixes https://github.com/openshiftio/openshift.io/issues/3265

The service is refactored so code can be shared between the pipelines widget, recent pipelines widget and pipelines page. Tests are updated as appropriate.


src/app/dashboard-widgets/pipelines-widget/pipelines-widget.component.ts
```diff
@@ -39,12 +43,11 @@ export class PipelinesWidgetComponent implements OnInit {
 
   ngOnInit() {
     this.contextPath = this.context.current.map(context => context.path);
-    this.buildConfigs = this.pipelinesService.current.share();
+    this.buildConfigs = this.pipelinesService.getCurrentPipelines();
     // buildConfigsCount triggers changes in the DOM; force Angular Change Detection
     // via setTimeout encapsulation
     this.buildConfigs
       .map(buildConfigs => buildConfigs.length)
-      .share()
```
andrewazores: Why don't these Observables need to be shared anymore?

src/app/space/create/pipelines/services/pipelines.service.ts
```diff
@@ -24,6 +24,10 @@ import { AuthenticationService } from 'ngx-login-client';
 
 import { NotificationsService } from 'app/shared/notifications.service';
 
+import { BuildConfig } from 'a-runtime-console/index';
```
joshuawilson: Do you think we need to bring this into the main app? Out of runtime-console.

jiekang: Hmm good question. I would ultimately like to bring the entire pipelines work out of runtime-console so we can delete the runtime-console folder. In the process, I would like to refactor it to be simpler as well.

I'm not sure how much benefit there will be to moving this specific export out of the runtime console; where it lives doesn't change much as far as I understand it.

src/app/dashboard-widgets/pipelines-widget/pipelines-widget.component.ts
jiekang: Right; I removed because they were in the wrong place but I forgot to actually add them to the right place. It should be correct now.

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2792-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2792-fabric8-ui.badger.fabric8.io

src/app/space/create/pipelines/services/pipelines.service.ts
```diff
@@ -62,6 +67,22 @@ export class PipelinesService {
     this.apiUrl = witUrl + 'user/services';
   }
 
+  getCurrentPipelines(): Observable<BuildConfig[]> {
+    return this.runtimePipelinesService.current
+      .distinctUntilChanged()
+      .combineLatest(
+        this.getOpenshiftConsoleUrl(),
+        this.setupBuildConfigLinks);
```
andrewazores: Would it make sense to add the `.share()` operator to the Observable as returned by these two functions, rather than doing it at the component callsite?

Also, it shouldn't make any real difference since `getOpenShiftConsoleUrl` will only emit once, but I think it would read easier if `distinctUntilChanged` was after `combineLatest`.

jiekang: I was reluctant to `share()` in the service since I felt like it was consumer-specific behavior. In this case two of the consumers do want to `share`, however the Pipelines page doesn't need it (as far as I understand the situation; I could be wrong).

andrewazores: Okay, fair enough.

jiekang: For the `distinctUntilChanged`, I think the placement changes it's behavior. I want to apply `distinctUntilChanged` only to the first observable being combined and not to the combined observable. That way the projection function will be run less often. At least that's my understanding. Thoughts?

jiekang: Hmm, maybe it's a bit confusing and I should convert it to the Observable form
```
Obserable.combineLatest(
  obs1.distinctUntilChanged()
  obs2
  projectionFn
)
```

andrewazores: Right, but if one of those combined Observables only emits once, it's running the project function a constant one-fewer time, right? Not really a big performance concern.

andrewazores: Yea that could work too. Personally I find that static combineLatest generally just easier to read anyway, and in that form the distinctUntilChanged is pretty clear regardless of whether it's on the first inner Observable or the final one and what's going to happen overall. I still think it should be fine to apply it to the combined Observable though.

jiekang: It will run the project function anytime any one of the 'inner' observables emits. So if the buildConfig emits `[]` 10 times, it will project that with the console url 10 times. No?

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2792-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2792-fabric8-ui.badger.fabric8.io

src/app/space/create/pipelines/services/pipelines.service.ts
andrewazores: Ahh yes, I see what you mean now. Yea, that is true.

andrewazores: Sorry, I was thinking in terms of the output Observable, and I don't know what I was thinking with the projection function part...

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2792-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2792-fabric8-ui.badger.fabric8.io

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2792-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2792-fabric8-ui.badger.fabric8.io

PR #2794, opened 2018-04-24
fix(routing): prevent side-effects of action outlet for add-work-item and add-codebases

aptmac:
fixes https://github.com/openshiftio/openshift.io/issues/3021 [0]

By fixing up the side-effects that named router outlets are currently causing, this PR accomplishes two things:

1. Makes the add-work-item action into a separate page

Previously, the add-work-item action would display an "overlay", and cause a 404 when trying to go to the planner, settings, or create page while active [0]. Now, the overlay can be treated as it's own page, which maintains the routing we would expect. 

**Note, at the time of writing this something is wrong in the f8-planner side of the add-work-item [1], specifically trying to retrieve the work item types is returning a 404; it looks like it may get amended by [2] . There is an issue logged for it here: [3].

2. Removes the add-codebases action from the dashboard, and modifies the header component so that switching between the codebases/pipelines/deployments pages with the outlet active will update the header accordingly.

The add-codebases action, similar to the add-work-items action, has routing problems when used in the dashboard. To patch this as discussed in [0] the add-codebases action has been replaced with the code wizard. On the create/ pages however, I patched up the url verification for determining active menu items, so the add-codebases action can still work in codebases and update the header accordingly. 

[0] https://github.com/openshiftio/openshift.io/issues/3021
[1] https://github.com/fabric8-ui/fabric8-planner/blob/fa4c909b8297f60e29b5d67dfd43db0e8f0850e4/src/app/components/work-item-create/work-item-create-selector/work-item-create-selector.component.html
[2] https://github.com/fabric8-ui/fabric8-planner/pull/2574
[3] https://github.com/openshiftio/openshift.io/issues/3300

### DIFF
--- a/src/app/dashboard-widgets/add-codebase-widget/add-codebase-widget.component.html
+++ b/src/app/dashboard-widgets/add-codebase-widget/add-codebase-widget.component.html
@@ -2,7 +2,7 @@
   <div id="spacehome-codebases" class="add-codebase-widget card-pf f8-card" user-level>
     <div class="card-pf-heading f8-card-heading">
       <div class="card-pf-heading-details f8-card-heading-details">
-        <a id="spacehome-codebases-add-space-button" class="btn btn-link f8-card-heading-btn-link" [routerLink]="[contextPath, 'create', { outlets: { action: 'add-codebase' } }]">
+        <a id="spacehome-codebases-add-space-button" class="btn btn-link f8-card-heading-btn-link" (click)="addToSpace.emit()">
           <i class="pficon pficon-add-circle-o"></i>
         </a>
       </div>
@@ -14,10 +14,7 @@
       <div class="f8-blank-slate-card" *ngIf="codebases.length === 0">
         <h3>This Space has no Codebases</h3>
         <div class="f8-blank-slate-main-action">
-          <button id="spacehome-codebases-create-button" class="btn btn-primary btn-lg" (click)="addToSpace.emit()">Add Codebase</button>
-        </div>
-        <div class="f8-blank-slate-main-action">
-          <button id="spacehome-codebases-import-button" class="btn btn-primary btn-lg" [routerLink]="[{ outlets: { action: 'add-codebase' } }]">Import Codebase</button>
+          <button id="spacehome-codebases-create-button" class="btn btn-primary btn-lg" (click)="addToSpace.emit()">Add a Codebase</button>
         </div>
       </div>
       <div id="spacehome-codebases-card-list" class="list-group list-view-pf list-view-pf-view list-view-pf-striped" *ngIf="codebases.length > 0">
@@ -44,7 +41,7 @@
   <div id="spacehome-codebases-card" class="add-codebase-widget card-pf f8-card" default-level>
     <div class="card-pf-heading f8-card-heading">
       <div class="card-pf-heading-details f8-card-heading-details">
-        <a id="spacehome-codebases-add-button" class="btn btn-link f8-card-heading-btn-link" [routerLink]="[contextPath, 'create', { outlets: { action: 'add-codebase' } }]">
+        <a id="spacehome-codebases-add-button" class="btn btn-link f8-card-heading-btn-link" (click)="addToSpace.emit()">
           <i class="pficon pficon-add-circle-o"></i>
         </a>
       </div>
@@ -57,10 +54,7 @@
       <div class="f8-blank-slate-card" *ngIf="codebases.length === 0">
         <h3>This Space has no Codebases</h3>
         <div class="f8-blank-slate-main-action">
-          <button id="spacehome-my-codebases-create-button" class="btn btn-primary btn-lg" (click)="addToSpace.emit()">Add Codebase</button>
-        </div>
-        <div class="f8-blank-slate-main-action">
-          <button id="spacehome-my-codebases-import-button" class="btn btn-primary btn-lg" [routerLink]="[{ outlets: { action: 'add-codebase' } }]">Import Codebase</button>
+          <button id="spacehome-my-codebases-create-button" class="btn btn-primary btn-lg" (click)="addToSpace.emit()">Add a Codebase</button>
         </div>
       </div>
       <ul id="spacehome-codebases-list" class="list-group list-view-pf list-view-pf-view list-view-pf-striped" *ngIf="codebases.length > 0">
--- a/src/app/dashboard-widgets/create-work-item-widget/create-work-item-overlay/create-work-item-overlay.component.ts
+++ b/src/app/dashboard-widgets/create-work-item-widget/create-work-item-overlay/create-work-item-overlay.component.ts
@@ -1,3 +1,4 @@
+import { Location } from '@angular/common';
 import { AfterViewInit, Component, OnInit, ViewChild, ViewEncapsulation } from '@angular/core';
 import { Router } from '@angular/router';
 
@@ -30,7 +31,8 @@ export class CreateWorkItemOverlayComponent implements OnInit, AfterViewInit {
     private router: Router,
     private spaces: Spaces,
     private broadcaster: Broadcaster,
-    private workItemService: WorkItemService
+    private workItemService: WorkItemService,
+    private location: Location
   ) { }
 
   ngOnInit() {
@@ -43,7 +45,7 @@ export class CreateWorkItemOverlayComponent implements OnInit, AfterViewInit {
   }
 
   onClose() {
-    this.router.navigateByUrl(removeAction(this.router.url));
+    this.location.back();
   }
 
   onSelect(workItemType: WorkItemType) {
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -11,7 +11,24 @@ import { SigninComponent } from './signin/signin.component';
 
 
 export function removeAction(url: string) {
-  return trimEnd(url.replace(/\(action:[a-z-]*\)/, ''), '/');
+  const nestedOutletRegex: RegExp = (/^(.*\([A-z-]*?)\/{0,2}action:[A-z-]*\)$/);
+  if (url.indexOf('action:') !== -1) {
+    if (nestedOutletRegex.test(url)) {
+      // if the action outlet contains the current page
+      // e.g., /${user}/${space}/create/(pipelines//action:add-codebase)
+      url = url.match(nestedOutletRegex)[1];
+      let i = url.lastIndexOf('(');
+      url = url.substring(0, i) + url.substring(i + 1);
+      if ((url.lastIndexOf('/') + 1) === url.length) {
+        url = url.slice(0, -1); // trim trailing '/' if applicable
+      }
+    } else {
+      // if the outlet is isolated at the end of the url
+      // e.g., /${user}/${space}/create/(action:add-codebase)
+      url = trimEnd(url.replace(/\(action:[a-z-]*\)/, ''), '/');
+    }
+  }
+  return url;
 }
 
 export const routes: Routes = [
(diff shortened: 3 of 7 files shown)


fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2794-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2794-fabric8-ui.badger.fabric8.io

src/app/dashboard-widgets/add-codebase-widget/add-codebase-widget.component.html
```diff
@@ -14,10 +14,7 @@ <h2 class="card-pf-title">
       <div class="f8-blank-slate-card" *ngIf="codebases.length === 0">
         <h3>This Space has no Codebases</h3>
         <div class="f8-blank-slate-main-action">
-          <button id="spacehome-codebases-create-button" class="btn btn-primary btn-lg" (click)="addToSpace.emit()">Add Codebase</button>
-        </div>
-        <div class="f8-blank-slate-main-action">
-          <button id="spacehome-codebases-import-button" class="btn btn-primary btn-lg" [routerLink]="[{ outlets: { action: 'add-codebase' } }]">Import Codebase</button>
```
dlabrecq: Why was the import codebase button removed?

FYI, @mindreeper2420

src/app/dashboard-widgets/add-codebase-widget/add-codebase-widget.component.html
```diff
@@ -57,10 +54,7 @@ <h2 class="card-pf-title">
       <div class="f8-blank-slate-card" *ngIf="codebases.length === 0">
         <h3>This Space has no Codebases</h3>
         <div class="f8-blank-slate-main-action">
-          <button id="spacehome-my-codebases-create-button" class="btn btn-primary btn-lg" (click)="addToSpace.emit()">Add Codebase</button>
-        </div>
-        <div class="f8-blank-slate-main-action">
-          <button id="spacehome-my-codebases-import-button" class="btn btn-primary btn-lg" [routerLink]="[{ outlets: { action: 'add-codebase' } }]">Import Codebase</button>
```
dlabrecq: Similarly, why was this removed from the old widget?

src/app/dashboard-widgets/create-work-item-widget/create-work-item-overlay/create-work-item-overlay.component.ts
```diff
@@ -43,7 +45,7 @@ export class CreateWorkItemOverlayComponent implements OnInit, AfterViewInit {
   }
 
   onClose() {
-    this.router.navigateByUrl(removeAction(this.router.url));
+    this.location.back();
```
dlabrecq: When choosing to create a work item, I encounter the error below. I don't see the overlay, but an empty page that says "Choose a new work item type".

ERROR Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'false'. Current value: 'true'.

src/app/dashboard-widgets/work-item-widget/work-item-widget.component.html
```diff
@@ -12,7 +12,7 @@ <h3>Create Work Items</h3>
       </p>
       <div class="f8-blank-slate-main-action">
         <button id="spacehome-workitems-create-button" class="btn btn-primary btn-lg"
-                [routerLink]="[{ outlets: { action: 'add-work-item' } }]">Create work item</button>
+                [routerLink]="['add-work-item']">Create work item</button>
```
dlabrecq: When choosing to create a work item from the new work-item-widget, I encounter the same error below. The same empty page displays with "Choose a new work item type" and nothing more.

ERROR Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'false'. Current value: 'true'.

dlabrecq:
FYI, I tested both locally and the snapshot and see the same result.

src/app/dashboard-widgets/work-item-widget/work-item-widget.component.html
aptmac: Both of those exist currently, the logged error isn't displayed in production however.

The first bug is noted here: https://github.com/openshiftio/openshift.io/issues/3300

And the second is noted here: https://github.com/fabric8-ui/fabric8-planner/issues/2571

jiekang:
The add work item page with this patch is blank. It exposes a component from the Planner project:

```
  <create-selector-widget #detailAddTypeSelector [workItemTypes]="workItemTypes | async" (onSelect)="onSelect($event)" (onClose)="onClose()"></create-selector-widget>
```

There is an internal `*ngFor` that has a null array.

jiekang:
The request to get workitemtypes is 404 for me:

https://api.openshift.io/api/spaces/<id>/workitemtypes

jiekang:
Looks like that endpoint no longer exists. The code is in Planner. Hmm...

joshuawilson:
@dlabrecq can you please take a look at the conflicts since it looks like they came from the changes to the dashboards widgets.

dlabrecq:
This doesn't appear to be related to the recent dashboard changes. The create-selector-widget is used with the old dashboard, not the new layout. This Planner component was something added to fabric8-create-work-item-widget about a year ago; although, it's broken in prod as well.

The work items widget still works, mostly. Clicking on the card header takes you to Planner. Clicking on a work item also takes you to Planner. It's just the overlay that's not shown properly when clicking the "add work item" icon. Like @jiekang stated above, it seems the end point no longer exists.

Should we just remove the "add" icon for now?

src/app/dashboard-widgets/add-codebase-widget/add-codebase-widget.component.html
jiekang: I think this is touched in the long discussion of https://github.com/openshiftio/openshift.io/issues/3021

https://github.com/openshiftio/openshift.io/issues/3100#issuecomment-380884315
suggests import codebase functionality can be removed.

On 3021: Alex says:

"Additionally, if we want a fix in place for summit but haven't fully hashed out the details of url patterns, would it be acceptable to make a patch that removes the add-codebases action (or replace it's functionality with displaying the forge-wizard), and have the add-work-item "overlay" be it's own page?"

Joshua replies

"I think it would be. That would at least not break user flow."


So I think these are intentional removals.

joshuawilson: I still think if you can't fix it fast, then removing it for now is best.

@catrobson @qodfathr do you agree?

dlabrecq:
FYI, there appears to be another, valid URL we can use with the fabric8-create-work-item-widget:

this.workItemTypes = this.workItemService.getWorkItemTypes2(this.space.relationships.workitemtypes.links.related);

I can navigate to the planner without further changes, but run into the same fabric8-ui/fabric8-planner#2571 issue describe above.

src/app/dashboard-widgets/add-codebase-widget/add-codebase-widget.component.html
catrobson: +1 that Add Codebase should go to Launcher Experience (if that flag is enabled for that user on the button) or to the old Forge wizard (if the feature flag is not enabled for that user on the button).  If we can't go to the old forge wizard, removing the button from the old widget would be acceptable to me over keeping the old 'create codebase' fly-out panel.  @dlabrecq does this make sense to you - I want to make sure I'm understanding this correctly.

dlabrecq: There are two issues here; the "import codebase" button and the "add work item" icon. 

It's still not clear as to why the "import codebases" button was removed because it still works for me without any changes. If the team wants to remove the "import codebase" button, I'm fine with that. I understand that it's not needed with the new app-launcher, but believe that's outside the scope of this PR.

Regarding the "add work item" icon, I think that should be removed. Even with these fixes, the user cannot create a work item because two APIs no longer exists. I was able to work around one of them, but not the second missing AP. These APIs are no longer in the swagger docs.

jiekang:
@dlabrecq Thanks so much for your help here!

It looks like with #2839 this area is going to be touched quite heavily. The primary goal of this PR is to prevent a 404 Not Found from occurring when trying to use these actions. This PR achieves that goal. It's unfortunate that one of the actions is currently broken, making it hard to test but I do believe that a separate PR is best to solve that as they are separate issues in the same area of code. Dan has #2843 for that already, which is great, even if it's WIP.

Are there still objections to merging this PR that haven't been addressed? I'd appreciate repeating them so I can do my best to address them!

src/app/dashboard-widgets/add-codebase-widget/add-codebase-widget.component.html
jiekang: I think the issue with the "import codebase" is that if you use it and then route to pipelines or deployments, your URL looks something like:

`https://openshift.io/jkang@redhat.com/arstneio/create/(pipelines//action:add-codebase)`

and the header menu then highlights the Codebases tab even when you're on Pipelines or Deployments. However, as far as I can tell, you're right, everything else works.

dlabrecq: That functionality looks okay to me. Although, it seems we didn't have these routing issues until the two router-outlets were added to app-component.html?

aptmac:
Yeah sorry about that, I've fixed up that test and rebased.

src/app/dashboard-widgets/add-codebase-widget/add-codebase-widget.component.html
jiekang: Huh; I really can't say re. the router outlets being added and these issues being seen. That's another thing to investigate I guess.

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2794-8 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2794-fabric8-ui.badger.fabric8.io

jiekang:
@joshuawilson Any thoughts on merging this? It should be safe to go in.

dlabrecq:
Considering @michaelkleinhenz 's comments in https://github.com/openshiftio/openshift.io/issues/3383#issuecomment-385960011, do we need these changes? That is, considering that the "add work item" work flow will not work anymore, this change may now be unnecessary after we remove the feature?

jiekang:
I think the changes to the header implementation and the codebase action could still be useful. This makes sure the tabs have the correct selection graphic. It's fixing a much smaller issue though and I'd say, not a P0.

jiekang:
It's probably best to close this PR and work on another smaller issue/PR for the add-codebase/header work.

dlabrecq:
I'm all for a new PR. I'm still concerned that the app-routing.module changes hinder us from using outlets in the future. I didn't find any navigation issues with the existing codebase flow, so just want to make certain before introducing such a major change.

jiekang:
Yeah that makes sense. I think it's something we can investigate more before continuing. As for the outlets my understanding is that the main issue hindering us was how we've named

Analyze as `/`
Planner as `/planner`
Create as `/create`

If Analyze was `/analyze` this wouldn't have been a problem. I honestly think it makes more sense with `/analyze` instead of `/`, but it's too big of a change to make at this time.

There isn't a navigation issue. It's just that when using the add codebase action, it directs you to the create/codebases page. If you then navigate to create/pipelines or create/deployments, the header tab still highlights Codebases, instead of the page you are actually on. Very minor but from a user experience it could look quite bad. How could they get something so simple wrong kind of thing.

dlabrecq:
I agree

aptmac:
I'm going to close this PR. The issues with the add-work-items flow will be fixed by removing the functionality completely [0] as discussed in [1].

I've noticed that the old add-codebases slider is only available with feature-flags set to production features, and this can still lead to a 404 when trying to navigate to /create for example from /${space}/(action:add-codebase). Additionally, having the add-codebases slider open on the codebases page and navigating to /create/pipelines still results in incorrect header highlighting. These can be addressed in a follow-up issue/PR. In feature-opt-in level beta and above, the slider is replaced with an overlay (and don't use the action) and so these issues don't occur.

[0] https://github.com/fabric8-ui/fabric8-ui/pull/2855
[1] https://github.com/openshiftio/openshift.io/issues/3383

PR #2804, opened 2018-04-25
fix(cleanup): add scrollbar to page when many spaces

aptmac:
fixes https://github.com/openshiftio/openshift.io/issues/3269

When a user has a large number of spaces, the reset tenant button can be unacceptable due to the lack of a scrollbar on that page. To fix, the overflow-y: auto; CSS rule has been added to include a vertical scrollbar if and when needed. 

With the overflow css, the page now looks like this when there are a large number of spaces:
fix-without-padding

I thought it looked kind of odd with the reset button being exactly at the bottom edge of the page, so I added a 30px padding (similar to other components on that page) to create a bit of room at the bottom.
fix-with-padding

Here it is in action:
cleanup-scroll


mindreeper2420:
@aptmac I'm not sure why this page didn't automatically scroll - there shouldn't be a need to add any kind of `overflow` or extra padding to the bottom of the page.

aptmac:
Neat! Like you I cannot replicate this issue .. despite seeing it all yesterday afternoon. I'm going to poke around a bit and see if I can figure out what was going on.

mindreeper2420:
@aptmac did you happen to visit the Planner before going to this page? There is a known issue where CSS conflicts are causing pages to not be scrollable, until the user manually refreshes their screen.

aptmac:
@mindreeper2420 Yeah that's what I'm thinking .. I made a temp fix for that on the f8-planner side and had been verifying it yesterday, I may have forgotten to refresh my browser when I saw the issue [0] that I referenced at the top of the pr. I'll close this PR and issue now.

[0] https://github.com/openshiftio/openshift.io/issues/3269

PR #2805, opened 2018-04-26
Deploy on RHEL based container

jmelis:
Enable build on RHEL if TARGET=rhel.

As part of an effort to migrate the services running in OSIO from CentOS
to RHEL, we are ready to push out the RHEL based image for this service
into staging.

This PR includes a new RHEL based dockerfile for the deployment of the service.
This dockerfile will be built when this variable is set: TARGET=rhel. The build
scripts have been adapted to take this into consideration, and to push the image
to different namespaces if the TARGET is rhel or not.

Currently the deployment builds happen on empty baremetal machines. The build
script will be executed a second time to build and push the deployment file if
TARGET=rhel, so some changes in the build script are to ensure that it will work
if executed a second time.

The OpenShift template has also been patched so it allows parametrization for
the container IMAGE, enabling the possibility of having different IMAGE urls for
staging and for production.

By merging this PR this service will be deployed to staging. Once we verify that
the RHEL based image works correctly, we can push to prod.


alien-ike:
### Ike Plugins (test-keeper)

Thank you @jmelis for this contribution!



joshuawilson:
[test]

PR #2814, opened 2018-04-26
fix(overview): Disable add to space button when in foreign space

chrislessard:
This patch disables the "Add to space" button when viewing another user's space, in the analyse overview page, and also adds some basic tests for the component. It addresses one of three sections of the code base affected by https://github.com/openshiftio/openshift.io/issues/1122.


src/app/space/analyze/analyze-overview/analyze-overview.component.html
```diff
@@ -36,8 +36,15 @@
         <fabric8-edit-space-description-widget></fabric8-edit-space-description-widget>
       </div>
       <div class="col-xs-4 col-xs-offset-4 col-sm-2 col-sm-offset-0">
-        <button *ngIf="this.authentication.isLoggedIn()" id="analyze-overview-add-to-space-button" class="btn btn-primary btn-lg pull-right"
-          (click)="openForgeWizard(addSpace)">Add to space</button>
+        <button
+          *ngIf="this.authentication.isLoggedIn()"
+          [disabled]="isInOtherUsersSpace()"
+          [title]="isInOtherUsersSpace() ? 'Add to space' : 'You cannot add to another user\'s space'"
```
andrewazores: Is this right? Seems like the title values are reversed. If you are in another user's space it says you can add, else if you are not (ie you are in your own) it says you cannot?

src/app/space/analyze/analyze-overview/analyze-overview.component.ts
```diff
@@ -60,4 +68,10 @@ export class AnalyzeOverviewComponent implements OnInit, OnDestroy {
     this.selectedFlow = $event.flow;
   }
 
+  isInOtherUsersSpace(): boolean {
+    if (this.context && this.loggedInUser) {
+      return !(this.context.space.relationships['owned-by'].data.id === this.loggedInUser.id);
+    }
+    return false;
+  }
```
andrewazores: I think this would be more easily readable if the function was `userOwnsSpace()` or something - ie that it returns true iff the current space is owned by the user viewing it, rather than the reverse.

src/app/space/analyze/analyze-overview/analyze-overview.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+})
+class ImportWizard {
+}
+
+@Component({
+  selector: 'quickstart-wizard',
+  template: ''
+})
+class QuickstartWizard {
+}
```
andrewazores: I think you can probably get away with not mocking and declaring a lot of these fake components using `NO_ERRORS_SCHEMA` (https://angular.io/api/core/NO_ERRORS_SCHEMA). Please see which you can get away with not mocking here to reduce the amount of boilerplate in this file.

src/app/space/analyze/analyze-overview/analyze-overview.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+
+describe('AnalyzeOverviewComponent', () => {
+  type ctx = TestContext<AnalyzeOverviewComponent, HostComponent>;
```
andrewazores: `type ctx` should have a capitalized type name, like `type Ctx` or `type TestingContext`

src/app/space/analyze/analyze-overview/analyze-overview.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  let modalService: jasmine.SpyObj<BsModalService>;
+  let broadcaster: jasmine.SpyObj<Broadcaster>;
+  let authentication: jasmine.SpyObj<AuthenticationService>;
+  let ctxSubj: Subject<Context>;
+
+  let authenticationServiceMock = (): jasmine.SpyObj<AuthenticationService> => {
+    authentication.isLoggedIn = function() { return true; } as (() => boolean) & jasmine.Spy;
+    return authentication;
+  };
```
andrewazores: This is a little awkward. This is named as a mock but is not actually a mock - it's a function that returns a mock. Why does this need to be extracted out rather than inlined into the module configuration on line 160?

src/app/space/analyze/analyze-overview/analyze-overview.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+describe('AnalyzeOverviewComponent', () => {
+  type ctx = TestContext<AnalyzeOverviewComponent, HostComponent>;
+
+  let modalService: jasmine.SpyObj<BsModalService>;
+  let broadcaster: jasmine.SpyObj<Broadcaster>;
+  let authentication: jasmine.SpyObj<AuthenticationService>;
+  let ctxSubj: Subject<Context>;
+
+  let authenticationServiceMock = (): jasmine.SpyObj<AuthenticationService> => {
+    authentication.isLoggedIn = function() { return true; } as (() => boolean) & jasmine.Spy;
```
andrewazores: If the only property you need on this mock is `isLoggedIn`, which is actually a function, and is not even a spy object, there's no need to declare the mock as a `jasmine.SpyObj` to begin with, is there?

```
{ provide: AuthenticationService: useValue: ({ isLoggedIn: () => true }) }
```

seems like it would do the same job with fewer lines and less forced (and inaccurate) type casting.

src/app/space/analyze/analyze-overview/analyze-overview.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+  };
+
+  let contextsMock = (): jasmine.SpyObj<Contexts> => {
+    const contexts: jasmine.SpyObj<Contexts> = createMock(Contexts);
+
+    ctxSubj = new Subject<Context>();
+    contexts.current = ctxSubj as Subject<Context> & jasmine.Spy;
+
+    return contexts;
+  };
```
andrewazores: `{ provide: Contexts, useValue: ({ current: ctxSubj }) }`

src/app/space/analyze/analyze-overview/analyze-overview.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  let userServiceMock = () => {
+    const userService: jasmine.SpyObj<UserService> = createMock(UserService);
+    userService.loggedInUser = Observable.of({
+      id: 'loggedInUser'
+    } as User).publish() as ConnectableObservable<User> & jasmine.Spy;
+
+    return userService;
```
andrewazores: similar to above

src/app/space/analyze/analyze-overview/analyze-overview.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+        }
+      } as Space
+    } as Context);
+
+    this.detectChanges();
+    expect(this.testedDirective.isInOtherUsersSpace).toHaveBeenCalled();
+  });
+
+  it('should recognize that the user owns the space', function(this: ctx) {
+    const contexts: jasmine.SpyObj<Contexts> = TestBed.get(Contexts);
```
andrewazores: unused?

src/app/space/analyze/analyze-overview/analyze-overview.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+    } as Context);
+
+    userService.loggedInUser.connect();
+    this.detectChanges();
+
+    expect(this.testedDirective.isInOtherUsersSpace()).toBe(false);
+  });
+
+  it('should recognize that the user does not own the space', function(this: ctx) {
+    const contexts: jasmine.SpyObj<Contexts> = TestBed.get(Contexts);
```
andrewazores: unused?

src/app/space/analyze/analyze-overview/analyze-overview.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+      id: 'loggedInUser'
+    } as User).publish() as ConnectableObservable<User> & jasmine.Spy;
+
+    return userService;
+  };
+
+  beforeEach(() => {
+    modalService = createMock(BsModalService);
+    broadcaster = createMock(Broadcaster);
+    authentication = createMock(AuthenticationService);
```
andrewazores: Using a beforeEach here to create the mocks and then using a useFactory below to return a reference to the freshly created instance is awkward. Why not just have the factory call createMock itself and return a new instance directly? Or, since you don't seem to be making any assertions about any of the functions on these mocks, why not just use `useValue` and return one instance once without using a factory at all?

src/app/space/analyze/analyze-overview/analyze-overview.spec.ts
```diff
@@ -3,7 +3,8 @@ import { AnalyzeOverviewComponent } from './analyze-overview.component';
 import {
```
andrewazores: I think this file should be renamed `analyze-overview.component.spec.ts`

src/app/space/analyze/analyze-overview/analyze-overview.spec.ts
```diff
@@ -24,147 +25,34 @@ import {
...
-      FakeFabric8AnalyticalReportWidget,
-      FakeFabric8AddCodebaseWidget,
-      FakeFabric8CreateWorkItemWidget,
-      FakeFabric8EnvironmentWidget,
-      FakeFabric8FeatureToggle,
-      FlowSelector,
-      ImportWizard,
-      QuickstartWizard
-    ],
+    declarations: [],
```
andrewazores: No need to include this when it's empty

src/app/space/analyze/analyze-overview/analyze-overview.spec.ts
```diff
@@ -223,6 +109,6 @@ describe('AnalyzeOverviewComponent', () => {
     userService.loggedInUser.connect();
     this.detectChanges();
 
-    expect(this.testedDirective.isInOtherUsersSpace()).toBe(true);
+    expect(this.testedDirective.userOwnsSpace()).toBe(false);
```
andrewazores: Please also add tests to exercise the not-logged-in case:

`if (this.context && this.loggedInUser) {`

So far you've tested the logic within this branch, where there is a context set and a user is logged in, and verified that the function works as expected if this user owns the space. There should also be tests where this branch is not entered, so `this.context` and/or `this.loggedInUser` are falsey.

src/app/space/analyze/analyze-overview/analyze-overview.component.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+  initContext,
+  TestContext
+} from 'testing/test-context';
+
+@Component({
+  template: '<alm-analyzeOverview></alm-analyzeOverview>'
+})
+class HostComponent { }
+
+fdescribe('AnalyzeOverviewComponent', () => {
```
andrewazores: This "fdescribe" needs to be set back to "describe"

src/app/space/analyze/analyze-overview/analyze-overview.component.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+        }
+      } as Space
+    } as Context);
+
+    this.detectChanges();
+    expect(this.testedDirective.userOwnsSpace).toHaveBeenCalled();
+  });
+
+  it('should disable the button if user service unavailable', function(this: TestingContext) {
+    fakeUserObs = null;
```
andrewazores: I don't think setting the observable reference to null here is going to do anything - the component accesses this observable via the UserService it receives via DI when constructed. I think emitting a falsey value from this observable - same as you've done with ctxSubject below - should do what was intended here.

PR #2823, opened 2018-04-27
fix(description): fix CSS to show the pencil when hovering a space de…

aptmac:
…scription

When hovering the description of a space, the behaviour is that a little pencil icon should pop up to show that it can be edited. However, on the new dashboard the pencil icon is missing on hover.

The reason for this is that the current CSS moves the pencil to a place that isn't visible.
before

To fix, the CSS has been amended to remove the displacement of the icon, and now the pencil shows up as expected.
after


fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2823-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2823-fabric8-ui.badger.fabric8.io

PR #2826, opened 2018-04-27
fix(stylesheets): contain feature-opt-in list stylesheet modifications

aptmac:
Introduced in the feature-opt-in commit [0], the feature-opt-in selection has been updated to use pfng-lists to display the current/available feature levels, and their descriptions. There was introduced a modification to the less file [1] to change the padding of the list items. However, the CSS modifications persist and affects other pfng-list (as shown on the codebases page) here:
skinny-list

where it regularly looks like this:
regular-list

This PR introduces a class tag to the html of that particular pfng-list so it can be specifically targeted by the .less file without modifying other lists.

[0] https://github.com/fabric8-ui/fabric8-ui/commit/9332ade583ba9ab04ac1918b0d4744cd878f4374
[1] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/profile/settings/feature-opt-in/feature-opt-in.component.less#L41


mindreeper2420:
@aptmac Good catch. Can you wrap all of the styles for the feature opt-in component with the `.feature-opt-in` class? That way, we won't have to revisit this if any of the other classes cause conflicts.

aptmac:
@mindreeper2420 Yeah that sounds good to me, I've wrapped the styles in a feature-opt-in class.

src/app/profile/settings/feature-opt-in/feature-opt-in.component.less
```diff
@@ -1,58 +1,51 @@
...
-    line-height: 2em;
+.feature-opt-in {
+  .checkbox {
+    border-right: 1px solid @color-pf-black-300;
+    float: left;
+    margin-top: 20px;
+    margin-right: 15px;
+    margin-bottom: 20px;
+    padding: 3px 10px 3px 0;
+  }
```
dlabrecq: The checkbox selector raised a red flag for me. The list-pf-content selector of pfng-list already provides a border when useExpandRows is enabled.

It appears that you have implemented your own checkbox and expansion toggle. Why use the pfng-list at all if just bypassing all of its features?

corinnekrych:
Good catch @aptmac ! i think we should merge that PR

corinnekrych:
@dlabrecq pls provide more guidance on how er should use it. I've looked at: https://github.com/patternfly/patternfly-ng/blob/master/src/app/list/basic-list/example/list-compound-example.component.html#L29
I don'tunserdand the suggestion. 
1. put `true` to: https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/profile/settings/feature-opt-in/feature-opt-in.component.ts#L66
2. and then how do i link to the `expandId` template ?
I'm not clear how that should be used

dlabrecq:
You really shouldn't be using pfng-list-expand-toggle directly. That is just a helper component used by pfng-list to show the toggle icon when useExpandItems=true. All you need to do is set useExpandItems and provide expandTemplate for the expandable area.

Take another look at the example at http://www.patternfly.org/patternfly-ng/#/list, then click the "Simple Expansion" checkbox. The code for the example is shown in the HTML and Typescript tabs, under the Code section.

corinnekrych:
@dlabrecq @aptmac @mindreeper2420 closing this PR in favor of https://github.com/fabric8-ui/fabric8-ui/pull/2857

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2826-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2826-fabric8-ui.badger.fabric8.io

PR #2827, opened 2018-04-27
feat(deployments): disable "Scale Up" when environment quotas are rea…

andrewazores:
…ched

fixes https://github.com/openshiftio/openshift.io/issues/3345

screenshot from 2018-04-27 18-15-43

This PR also does some cleanup of the component, removing an unused variable (`scalable`) and tracking/unsubscribing Subscriptions. The feature is in the new `atQuota` property and the `combineLatest` Observable in ngOnInit that sets it, along with the template changes to disable the button and change the tooltip text based on `atQuota`.

### DIFF
--- a/src/app/space/create/deployments/deployments-donut/deployments-donut.component.html
+++ b/src/app/space/create/deployments/deployments-donut/deployments-donut.component.html
@@ -2,16 +2,16 @@
 </deployments-donut-chart>
 <div column class="deployments-donut-scale-controls fade-inline" *ngIf="!isIdled && !mini">
   <div>
-    <a id="scaleUp" (click)="$event.preventDefault(); scaleUp()" [ngClass]="{ disabled: !scalable }" [attr.title]="!scalable ? undefined : 'Scale up'"
-      [attr.aria-disabled]="!scalable ? 'true' : undefined" role="button">
+    <a id="scaleUp" (click)="$event.preventDefault(); scaleUp()" [ngClass]="{ disabled: atQuota }" [attr.title]="atQuota ? 'You don\'t have enough resources. Scale down the number of pods on another deployment to make resources available.' : 'Scale up'"
+      [attr.aria-disabled]="atQuota ? 'true' : undefined" role="button">
       <i class="fa fa-chevron-up"></i>
       <span class="sr-only">Scale up</span>
     </a>
   </div>
   <div>
     <!-- Remove the title when disabled because the not-allowed styled cursor overlaps the tooltip on some browsers. -->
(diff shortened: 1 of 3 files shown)


fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2827-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2827-fabric8-ui.badger.fabric8.io

jiekang:
@andrewazores 
When manual user testing, I am still able to trigger the scale up action when it's set to Disabled.

I think `scalable` was being used and may have been inadvertently broken somewhere along the line.

andrewazores:
@jiekang fixed, that was silly. Thanks for being more thorough with your manual testing than I was. I saw it looked disabled and just figured it was... whoops.

jiekang:
@andrewazores 

What do you think about the check incorporating: It should prevent users from scaling up past the quota, assuming that each scale would add the fixed amount for application. This would cover the ability for a user to start from 0/1 applications, and quickly click the scale up button to 5. This would be similar to how it already doesn't let users scale below 0, just the constant changes based on data from the last poll cycle info.

I guess handling the various scenarios can become complicated. And there's the timing, maybe a particularly fast user (automated clicks?) could get past that kind of check anyways. Could be another issue if you think it might be worth looking into.

andrewazores:
@jiekang yes, that's related to what Todd and I were just discussing on the related issue (https://github.com/openshiftio/openshift.io/issues/3345). If we can query for how much quota will be consumed by a pod, then we can incorporate this type of logic.

Marking as "Do not merge" as Todd has asked for confirmation from Cat first.

jiekang:
I think for an interm of quota consumed by a pod, it would be accurate to divide the used quota by the number of pods. Is it possible to scale a deployment up and have different quotas taken by the individual pods of that deployment? The UI developed for the OSO console suggests scaling a pod involves a complete replica of the deployment.

andrewazores:
I believe each pod would have the same allotment as the rest for a given deployment, but perhaps that's not always the case? Even if so, that approach wouldn't work for a deployment that currently has 0 pods, since we wouldn't be able to infer what the usage of each pod is at that point, even though the behaviour we're talking about would disable the "scale up" button if a single pod would exceed the quota. It's possible as another partial solution, but I'm not sure about it.

fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2827-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2827-fabric8-ui.badger.fabric8.io

jiekang:
Ah yes, excellent point regarding the 0 case. Definitely needs API work then, right?

andrewazores:
Yea. https://github.com/openshiftio/openshift.io/issues/3345#issuecomment-385568945

PR #2836, opened 2018-04-30
fix(a-runtime-console): also check for undefined user values in oauth-config-store

jiekang:
This addresses the oauth-config-store error reported in https://github.com/openshiftio/openshift.io/issues/3347

The `first()` predicate check now also checks for `undefined` and `null` instead of just `null`.

A test was added, however, in test runs it strangely throws an EmptyError. In actual runs, visiting /openshiftio/openshiftio while not logged in, no error is thrown. Prior to the fix, the test run error would be the stacktrace of 'can't access undefined...' so it's still a valuable test. If anyone has ideas what could be causing the EmptyError, I'd greatly appreciate the input!

### DIFF
--- a/src/a-runtime-console/kubernetes/store/oauth-config-store.spec.ts
+++ b/src/a-runtime-console/kubernetes/store/oauth-config-store.spec.ts
@@ -16,8 +16,9 @@ import {
 } from '@angular/http/testing';
 
 import {
-  EmptyError,
+  BehaviorSubject,
   Observable,
+  ReplaySubject,
   Subscription
 } from 'rxjs';
 
@@ -79,7 +80,7 @@ describe('OauthConfigStore', () => {
   describe('success state', () => {
     beforeEach(() => {
       mockUserService = createMock(UserService);
-      mockUserService.loggedInUser = Observable.of(user).publish();
+      mockUserService.loggedInUser = new BehaviorSubject(user).multicast(() => new ReplaySubject(1));
 
       TestBed.configureTestingModule({
         imports: [HttpModule],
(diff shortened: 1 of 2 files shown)


andrewazores:
EmptyError doesn't sound nice. Maybe along the lines of https://github.com/ReactiveX/rxjs/issues/2262 ?

jiekang:
@andrewazores it suggests the error is thrown because it knows that the observable used in the test is always going to output a bad value `undefined`. This is actually the expected behavior for the test to make sure the subscriber handles an `undefined` emission is okay. Thoughts?

andrewazores:
What do you mean "it knows it will always output a bad value"? That doesn't sound right...

It's definitely weird that the code behaves differently under test vs at actual runtime. This suggests to me that something about how the tests are set up is not mimicking the real scenario closely enough. Perhaps some Observable is emitting nothing before completing, rather than emitting an empty value and then completing, for example.

jiekang:
I'm just guessing based on that comment:

"you're seeing this error because your last predicate always returns undefined. See the docs for the last predicate here"

Re always output a bad value, the test run mocks the Observable to emit `{} as User` which is always bad. I'll try to investigate a more complex mock.

jiekang:
Looks like it's a consequence of `Observable.of({})` and then using `first()`. The actual implementation multicasts this behind a ReplaySubject. Will be doing that in the test as well.

jiekang:
@andrewazores thanks for the help. I've updated the PR, how does it look?

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2836-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2836-fabric8-ui.badger.fabric8.io

fabric8cd:
@jiekang fabric8/fabric8-ui:SNAPSHOT-PR-2836-6 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2836-fabric8-ui.badger.fabric8.io

PR #2837, opened 2018-04-30
fix: make github requests in UI instead from server

edewit:
This way request will no longer come from our cluster but from the users own IP.

fixes: https://github.com/openshiftio/openshift.io/issues/3367

### DIFF
--- a/src/app/space/app-launcher/app-launcher.module.ts
+++ b/src/app/space/app-launcher/app-launcher.module.ts
@@ -18,6 +18,8 @@ import { ImportAppModule } from './import-app/import-app.module';
 import { AuthAPIProvider } from './services/app-launcher-authprovider.service';
 import { NewForgeConfig } from './shared/new-forge.config';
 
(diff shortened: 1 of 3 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Thank you @edewit for this contribution!



edewit:
although this will just move the problem not really resolve it. The tests must run with a realistic amount of users otherwise github will increase the response time as this is a basically a DNS attack

edewit:
part of https://github.com/fabric8-launcher/launcher-planning/issues/158

gastaldi:
@edewit I think you're right, let's hold this PR until we confirm that this can minimize the problem

joshuawilson:
As Ike said, this needs tests.

maxandersen:
I grok the sentiment for this fix; but if this uses the same github token I don't think this will help AND we'll want to support multiple different git providers in the future - having the frontend keep doing github specific calls should be a short term thing (if at all needed).

PR #2838, opened 2018-04-30
fix(deployments): fix bug where Deployment card charts become corrupt

andrewazores:
Delay processing of stats data until after chart objects have loaded, and modify chart properties
for Y-axis max and colours via C3 API rather than Sparkline chart config object. This keeps the C3
chart object's state consistent with the DeploymentDetailsComponent state.

fixes https://github.com/openshiftio/openshift.io/issues/3230

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,7 +125,7 @@
     "angular2-websocket": "0.9.5",
     "chunk-manifest-webpack2-plugin": "1.0.1",
     "core-js": "2.5.1",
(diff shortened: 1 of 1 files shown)


fabric8cd:
@andrewazores fabric8/fabric8-ui:SNAPSHOT-PR-2838-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2838-fabric8-ui.badger.fabric8.io

PR #2840, opened 2018-04-30
fix(launcher): disable submit while processing request

christianvogt:
When the user clicks the OK button, a request to save the space is made which is async and therefore allows the user to continuously press the OK button while the request is being processed. Simply disabled the button until the request completes.

Also added `overflow-y: auto` to the dialog as I was unable to scroll vertically in a smaller window while testing.

fixes openshiftio/openshift.io#3198

### DIFF
--- a/src/app/space/add-space-overlay/add-space-overlay.component.less
+++ b/src/app/space/add-space-overlay/add-space-overlay.component.less
@@ -8,6 +8,7 @@
   bottom: 0;
   left: 0;
   background-color: rgba(0, 0, 0, .8);
+  overflow-y: auto;
   .create-launcher {
     color: @color-pf-white;
     .alert-danger {
--- a/src/app/space/add-space-overlay/add-space-overlay.component.html
+++ b/src/app/space/add-space-overlay/add-space-overlay.component.html
@@ -70,7 +70,7 @@
       <footer class="container-fluid padding-top-standard-offset padding-bottom-standard-offset">
         <div class="row">
           <div class="col-sm-12 create-launcher-step-tool-bar button-right">
-            <button id="createSpaceButton" class="btn btn-primary" [disabled]="!spaceForm.form.valid" type="submit">Ok</button>
+            <button id="createSpaceButton" class="btn btn-primary" [disabled]="!spaceForm.form.valid || !canSubmit" type="submit">Ok</button>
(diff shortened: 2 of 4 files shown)


src/app/space/add-space-overlay/add-space-overlay.component.less
```diff
@@ -8,6 +8,7 @@
   bottom: 0;
   left: 0;
   background-color: rgba(0, 0, 0, .8);
+  overflow-y: auto;
```
mindreeper2420: @christianvogt with this, you are essentially allowing for a double page scroll. I understand why, but I just wanted to let you know that in case you have another idea. When I make my browser small enough, I can scroll the overlay and, when that has finished scrolling, the page then scrolls.

christianvogt: @mindreeper2420 ya I'm aware of the double scroll. I was thinking of applying a class to the body to set `overflow: hidden` however then I found `add-app-overlay.component.less` was already setting `overflow-y: auto` and to fix one would also then require applying the fix there as well.

We should look at creating a single common component for displaying such overlays that take care of all of these quirks instead of one off fixes. Isn't that what patternfly is suppose to do for us?

I think the double scroll is the lesser evil here until such time.

mindreeper2420: Agreed that there should be single common component to handle these quirks. Regarding PatternFly, this does not actually exisit in PatternFly and is purely an OSIO component, so we have complete control over it (though it can always be submitted to PF for inclusion).

Agreed that this is the lesser of two evils for the time being, but also that we should standardize this interaction.

joshuawilson:
@mindreeper2420 do approve?

mindreeper2420:
@joshuawilson yup - thought I had.

fabric8cd:
@christianvogt fabric8/fabric8-ui:SNAPSHOT-PR-2840-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2840-fabric8-ui.badger.fabric8.io

fabric8cd:
@christianvogt fabric8/fabric8-ui:SNAPSHOT-PR-2840-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2840-fabric8-ui.badger.fabric8.io

PR #2841, opened 2018-04-30
fix(spaces): fix alignment of my spaces list item contents

christianvogt:
Cleaned up styles for my-spaces list items.
Removed unnecessary custom CSS rules which were trying to fight against patternfly styles. Instead simply stacked the content for all breakpoints as per the UX design.

fixes openshiftio/openshift.io#3381

image


src/app/profile/my-spaces/my-spaces-item/my-spaces-item.component.html
```diff
@@ -1,14 +1,11 @@
...
-      <a [routerLink]="['/', space?.relationalData?.creator?.attributes?.username, space?.attributes?.name]">
-        {{space?.attributes?.name | spaceName}}</a>
-      <span class="avatar">
-        <img src="{{space?.relationalData?.creator?.attributes?.imageURL}}">
-        <span class="margin-left-10">{{space?.relationalData?.creator?.attributes?.username}}</span>
-      </span>
-    </div>
-    <div class="list-pf-description margin-left-0">{{space?.attributes?.description}}</div>
+<div class="my-spaces-item" *ngIf="space?.relationalData?.creator?.attributes">
+  <div class="list-pf-title">
```
dlabrecq: I believe list-pf-title is meant to be located within list-pf-main-content. And the description should appear within list-pf-description.

You can refer to the Patternfly layout here:
https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/list-pf.html

Here is a snapshot of that same layout:
screen shot 2018-04-30 at 4 58 50 pm

mindreeper2420: @dlabrecq normally, yes, but the design calls for the description to be underneath the Title and Owner, rather than to the right of (per PatternFly List View spec).

What could be used instead of this customized standard list (and thus following PatternFly design spec) would be either the 3rd or 6th variation (found here: http://www.patternfly.org/pattern-library/content-views/list-view/#code). That would most likely be best as a separate issue, improving the code to make it more extensible for the final design implementation.

mindreeper2420: When inspecting the code, the majority of these elements are present (handled by PatternFly-ng). The only piece missing is `list-pf-description`, but that would then cause the page to deviate from the design. See my other comment on alternate List View implementations that would be worth exploring.

christianvogt: @dlabrecq the problem I had is that the patternfly-ng list does not apply the class `list-pf-stacked` to the list and therefore the title and description do not stack at all breakpoints. So either I choose to not use the patternfly classes for the list item content, or I fight the CSS by providing my own overrides. I chose the simple approach of managing my own content layout.

@mindreeper2420 to use those variations I would need to use `list-view-pf` which patternfly-ng list does not use. So again I'd have to do something abnormal just to make use of the patternfly classes.

Seems the patternfly-ng list doesn't provide enough API to control the various patterns one can achieve using the raw classes.

PR #2875, opened 2018-05-04
fix(deployments): rename applications to deployments in header

jiekang:
This fixes https://github.com/openshiftio/openshift.io/issues/3460

It dawned on me in the design [1], the header is Deployments, not Applications.

[1] https://redhat.invisionapp.com/share/YSDL0KRKW#/screens/260764292

### DIFF
--- a/src/app/space/create/deployments/apps/deployments-apps.component.html
+++ b/src/app/space/create/deployments/apps/deployments-apps.component.html
@@ -6,7 +6,7 @@
     <div class="row row-cards-pf">
       <div class="col-xs-12 col-sm-4 col-md-2">
         <h3>
-          <b>Applications</b>
+          <b>Deployments</b>
         </h3>
       </div>
       <div class="col-xs-12 col-sm-4 col-md-2">


jiekang:
On second thought, maybe the applications label is more accurate. The cards to the right are the deployments information while the list is of the applications [name].

jiekang:
@mceledonia thoughts?

jiekang:
The toolbar filter is also by 'Application Name'. This should probably be consistent, either Deployment Name if the header is Deployments or as is if the header is Applications.

mceledonia:
@jiekang I'm not too sure I'm following, I will explain how I'm interpreting the terms:

This page shows the user all of their deployments of their applications to their deployment environments. The label "applications" refers to the list in the left column - applications themselves. Deployments are the cards which represent deployments of those applications to a specific environment (of which it can be deployed to multiple at once). 

So in terms of rows/columns, each row is an application, and the columns of that row are the deployments, excluding the first column which is the application name. 

So when you filter for an application, you are filtering your view to see just that row, but you still might see multiple deployments of that application because that just might be the case. 

The issue here might be the "Applications" column header - maybe it looks too much like a page title vs. just the column header?

jiekang:
@mceledonia Hi! I think I may have made things overly confusing with my rambling.

At the moment: The header currently says 'Applications'. In the invision design, I just noticed that it says 'Deployments'. Which one is right? From your explanation it seems 'Applications' is more appropriate.

mceledonia:
@jiekang Haha no worries, that is relieving that it's simpler than I thought. But it did also bring to light a way to improve the page, so that's good too.

"Applications" is correct!

jiekang:
Awesome! Looks like this PR can be closed then. Thanks!

PR #2876, opened 2018-05-04
Limit application names to 40 characters or less (#3391)

stooke:
This addresses
https://github.com/openshiftio/openshift.io/issues/3391

There's a change in project-info-step-component.ts that I've made but am not sure if it's applicable.
Also, I do not know where the docs are, I'm happy to create PR for those, too.

I am worried that limiting this input might cause issues with existing longer-named applications, but I think the odds are low.

### DIFF
--- a/src/app/space/add-app-overlay/add-app-overlay.component.html
+++ b/src/app/space/add-app-overlay/add-app-overlay.component.html
@@ -33,14 +33,14 @@
                     Names must:
                     <ul>
                       <li>Be Unique</li>
-                      <li>Contain 4-65 characters</li>
+                      <li>Contain 4-40 characters</li>
                       <li>Contain only letters, numbers, spaces, underscores, or hyphens</li>
                       <li>Not start or end with spaces, underscores, or hyphens</li>
                     </ul>
                   </ng-template>
                   <span class="code-imports--step_tooltip-label" [popover]="namingTemplate" popoverTitle="Naming Requirements" placement="bottom" triggers="mouseenter:mouseleave">Naming Requirements<span class="pficon pficon-info"></span></span>
                   <span class="help-block code-imports--step_help-block" *ngIf="isProjectNameValid === false">
-                    Names may not start or end with spaces, underscores, or hyphens.
+                    Invalid name. See a full list of naming requirements below the input field.
                   </span>
                 </div>
               </form>
--- a/src/app/space/app-launcher/services/app-launcher-dependency-check.service.ts
+++ b/src/app/space/app-launcher/services/app-launcher-dependency-check.service.ts
@@ -27,8 +27,8 @@ export class AppLauncherDependencyCheckService implements DependencyCheckService
    * @returns boolean
    */
   validateProjectName(projectName: string): boolean {
-    // allows only '-'
-    const pattern = /^[a-zA-Z0-9][a-zA-Z0-9-_\s]{2,62}[a-zA-Z0-9]$/;
+    // allows only '-', '_', ' ' and 4-40 characters (must start and end with alphanumeric)
+    const pattern = /^[a-zA-Z0-9][a-zA-Z0-9-_\s]{2,38}[a-zA-Z0-9]$/;
     return pattern.test(projectName);
   }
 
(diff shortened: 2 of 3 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Thank you @stooke for this contribution!



jiekang:
I think the new application wizard is in a different repository? Is that affected?

joshuawilson:
ah, yes. you are correct

joshuawilson:
Please add a length validation check to the unit test at https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/space/app-launcher/services/app-launcher-dependency-check.service.spec.ts

stooke:
I think the new application wizard is part of this repo. 
Isn't this it?
https://github.com/fabric8-ui/fabric8-ui/tree/master/src/app/space/forge-wizard

jiekang:
The new 'launcher' is 'imported' from https://github.com/fabric8-launcher/ngx-launcher and used in https://github.com/fabric8-ui/fabric8-ui/tree/master/src/app/space/app-launcher

The naming is a bit weird as it's still called ngx-forge, though the repository is ngx-launcher.

jiekang:
The validation between the two is inconsistent too. The error messages are also not entirely useful

Old:

0123 is not accepted with: `OpenShift Project name should not contain uppercase or special characters`

The issue is that it can't start with a number while this suggests numbers aren't allowed at all.

New:

0123 is accepted. I've confirmed it goes up to 64 characters as well. Starting with a number should not be allowed though.

An unacceptable input like '123' or 'abc-@' has an even worse error message than the old experience: `Please enter a valid Project Name`
`

Would be nice to tell me what is valid. Anyways this is all unrelated to the goal of this PR.

stooke:
Jie, which input are you commenting on?  If it's project name, I didn't really touch that.

I haven't changed project names nor space name validation - we can expand the scope of the original issue.
I definitely can't enter more than 40 characters for application name on at least one screen.
I'll update the launcher but so far I've only found help text - not validation code.

How do I know which launcher I'm getting?

jiekang:
If you are 'Production' feature level, you will see the old launcher. If you are 'Experimental' you will see the new launcher. I think there's an issue with switching between feature levels not affecting already loaded things so I'd recommend hard-refreshing your tab after you make a switch.

jiekang:
Here's an image of the new launcher:

new-launcher

I expect the validation code will be in https://github.com/fabric8-launcher/ngx-launcher

@joshuawilson is anyone who works on that able to fix this? Does it need to be fixed there?

jiekang:
An image of the old launcher for good measure:

old-launcher

jiekang:
@stooke  Ah, it looks like the app overlay is part of the new launcher but within the launcher components you can change the name again.

jiekang:
Actually on investigation that service is injected into the new launcher so it's okay. There doesn't need to be any changes to the ngx-launcher repository.

@joshuawilson @stooke Okay nevermind me. Sorry about that fuss.

stooke:
There are hardcoded length messages in the ngx-launcher repo,  which seems odd as (I presume) the validation is done in the fabric8-ui repo.

joshuawilson:
Adding @arunkumars08 @jyasveer as they are on the Launcher team for the UI.

fabric8cd:
@stooke fabric8/fabric8-ui:SNAPSHOT-PR-2876-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2876-fabric8-ui.badger.fabric8.io

joshuawilson:
I see you added the tests, so I'm good with this. Is it ready to merge?

stooke:
const pattern = /^[a-zA-Z0-9][a-zA-Z0-9-_\s]{2,38}[a-zA-Z0-9]$/;

This validation pattern seems too permissive.  
I'd like to get rid of the whitepsace and no allow leading numbers.

i.e
const pattern = /^[a-zA-Z][a-zA-Z0-9-_]{2,38}[a-zA-Z0-9]$/;

joshuawilson:
That should go in another PR. We need to keep the changes atomic. 

Additionally, we need to vet the requirements. What are the other systems allowing or expecting.

stooke:
Sounds fine to me.  Do you want a PR for my proposed regex changes or just let that pass for now?  I think being more restrictive only be a good thing so I would suggest I do a PR for that.

Also, I don't have merge rights for this repo.  Could someone please merge for me?

stooke:
@joshuawilson I have created a new issue #3516 to address proposed regex changes.
https://github.com/openshiftio/openshift.io/issues/3516

PR #2882, opened 2018-05-06
fix(workitems): filter out closed, removed or done work items

sanbornsen:
- This is not the ideal way of doing it.
- The backend request should be modified in future for this filter but this will require meta state concept and we don't have that currently.
- see openshiftio/openshift.io#3473

### DIFF
--- a/src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
+++ b/src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
@@ -7,6 +7,8 @@ import { Broadcaster } from 'ngx-base';
 import { Contexts } from 'ngx-fabric8-wit';
 import { UserService } from 'ngx-login-client';
 
+import { filterOutClosedItems } from '../../shared/workitem-utils';
+
 class WorkItemFilter {
   paramKey: string;
   value: string;
@@ -46,6 +48,7 @@ export class CreateWorkItemWidgetComponent implements OnInit {
       .switchMap(filters => this.workItemService
         .getWorkItems(100000, filters))
       .map(val => val.workItems)
+      .map(workItems => filterOutClosedItems(workItems))
       // Resolve the work item type, creator and area
       .do(workItems => workItems.forEach(workItem => this.workItemService.resolveType(workItem)))
       .do(workItems => workItems.forEach(workItem => this.workItemService.resolveAreaForWorkItem(workItem)))
--- a/src/app/home/work-item-widget/work-item-widget.component.ts
+++ b/src/app/home/work-item-widget/work-item-widget.component.ts
@@ -6,6 +6,8 @@ import { User, UserService } from 'ngx-login-client';
 import { Subscription } from 'rxjs';
 import { Observable } from 'rxjs/Observable';
 
+import { filterOutClosedItems } from '../../shared/workitem-utils';
+
 @Component({
   encapsulation: ViewEncapsulation.None,
   selector: 'alm-work-item-widget',
@@ -99,6 +101,7 @@ export class WorkItemWidgetComponent implements OnDestroy, OnInit  {
       .switchMap(filters => this.workItemService
         .getWorkItems(100000, filters))
       .map(val => val.workItems)
+      .map(workItems => filterOutClosedItems(workItems))
       // Resolve the work item type, creator and area
       .do(workItems => workItems.forEach(workItem => this.workItemService.resolveType(workItem)))
       .do(workItems => workItems.forEach(workItem => {
--- a/src/app/profile/overview/work-items/work-items.component.ts
+++ b/src/app/profile/overview/work-items/work-items.component.ts
@@ -7,6 +7,8 @@ import { User, UserService } from 'ngx-login-client';
 import { Subscription } from 'rxjs';
 import { Observable } from 'rxjs/Observable';
 
+import { filterOutClosedItems } from '../../../shared/workitem-utils';
+
 @Component({
   encapsulation: ViewEncapsulation.None,
   selector: 'alm-work-items',
@@ -103,6 +105,7 @@ export class WorkItemsComponent implements OnDestroy, OnInit  {
       .switchMap(filters => this.workItemService
         .getWorkItems(100000, filters))
       .map(val => val.workItems)
+      .map(workItems => filterOutClosedItems(workItems))
       // Resolve the work item type, creator and area
       .do(workItems => workItems.forEach(workItem => this.workItemService.resolveType(workItem)))
       .do(workItems => workItems.forEach(workItem => {
--- a/src/app/shared/workitem-utils.spec.ts
+++ b/src/app/shared/workitem-utils.spec.ts
@@ -1,0 +1,37 @@
+import { filterOutClosedItems } from './workitem-utils';
+
+describe('Workitem utils: it', () => {
+
+  const inplist = [{
+    id: 'id-1',
+    attributes: {
+      'system.state': 'removed'
+    }
+  }, {
+    id: 'id-2',
+    attributes: {
+      'system.state': 'closed'
+    }
+  }, {
+    id: 'id-3',
+    attributes: {
+      'system.state': 'done'
+    }
+  }, {
+    id: 'id-4',
+    attributes: {
+      'system.state': 'new'
+    }
+  }];
+
+  const outputlist = [{
+    id: 'id-4',
+    attributes: {
+      'system.state': 'new'
+    }
+  }];
+
+  it('closed work items should be filtered out from the list', () => {
+    expect(filterOutClosedItems(inplist)).toEqual(outputlist);
+  });
+});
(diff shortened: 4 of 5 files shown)


nimishamukherjee:
[test]

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
```diff
@@ -44,8 +44,12 @@ export class CreateWorkItemWidgetComponent implements OnInit {
         active: true
       }])
       .switchMap(filters => this.workItemService
-        .getWorkItems(100000, filters))
+      .getWorkItems(100000, filters))
```
joshuawilson: why indent this?

joshuawilson:
Ike says you need some tests.

sanbornsen:
Anyway, this end point is deprecated thus the components should be re-written with the new endpoint and filter expressions. Besides, the concept of meta state will change a lot in this component. Therefore it makes no sense to put effort on writing a test bed for these components just to support this hack.

fabric8cd:
@sanbornsen fabric8/fabric8-ui:SNAPSHOT-PR-2882-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2882-fabric8-ui.badger.fabric8.io

sanbornsen:
@joshuawilson why do you think I wouldn't add a test if it was really needed here?

nimishamukherjee:
@joshuawilson can someone from platform add tests for this PR?

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
```diff
@@ -44,8 +44,13 @@ export class CreateWorkItemWidgetComponent implements OnInit {
         active: true
       }])
       .switchMap(filters => this.workItemService
-        .getWorkItems(100000, filters))
+        .getWorkItems(100000, filters)
```
aslakknutsen: 100000 is a silly number. 100 is Max you will get from backend

sanbornsen: Yeah very true. I'll change it.

fabric8cd:
@sanbornsen fabric8/fabric8-ui:SNAPSHOT-PR-2882-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2882-fabric8-ui.badger.fabric8.io

joshuawilson:
It's simple, you want a fix in that will go directly to prod, right before a huge demo, during a code freeze? Tests are all the more important now. 

You can try to change my mind.

joshuawilson:
the framework is already in place, just create a spec file for each ts file and put tests in it to test the function you are making changes to

nimishamukherjee:
@joshuawilson as you mentioned there is no spec file for this widget: https://github.com/fabric8-ui/fabric8-ui/tree/master/src/app/dashboard-widgets/create-work-item-widget
Our request is for someone from platform create the spec file for this component, as they would be aware of what basic components and services to include - this is the time consuming part for @sanbornsen 
@sanbornsen can then include the specific test for this PR

sanbornsen:
Yeah, that's what I meant by the test bed. A test bed should be prepared with the mocked dependencies for these components. Thanks @nimishamukherjee

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
```diff
@@ -44,8 +44,13 @@ export class CreateWorkItemWidgetComponent implements OnInit {
         active: true
       }])
       .switchMap(filters => this.workItemService
-        .getWorkItems(100000, filters))
+        .getWorkItems(100, filters)
```
maxandersen: does this change matter for this fix ? backend will ignore above 100 anyway. 
I agree it should be fixed but lets keep changes to bare minimum to not avoid suprirsing sideeffects.

src/app/home/work-item-widget/work-item-widget.component.ts
```diff
@@ -97,8 +97,13 @@ export class WorkItemWidgetComponent implements OnDestroy, OnInit  {
         active: true
       }])
       .switchMap(filters => this.workItemService
-        .getWorkItems(100000, filters))
+        .getWorkItems(100, filters)
```
maxandersen: same as above - is it relevant for this change?

src/app/home/work-item-widget/work-item-widget.component.ts
```diff
@@ -97,8 +97,13 @@ export class WorkItemWidgetComponent implements OnDestroy, OnInit  {
...
+      )
       .map(val => val.workItems)
+      .map(workItems => workItems.filter(item => {
```
maxandersen: this is pure duplicated code from above - any reason  why this bad temporary hack of filtering is not at least put into a shared method to avoid future deviation in the multiple places?

src/app/profile/overview/work-items/work-items.component.ts
```diff
@@ -101,8 +101,13 @@ export class WorkItemsComponent implements OnDestroy, OnInit  {
         active: true
       }])
       .switchMap(filters => this.workItemService
-        .getWorkItems(100000, filters))
+        .getWorkItems(100, filters)
```
maxandersen: same as above.

src/app/profile/overview/work-items/work-items.component.ts
```diff
@@ -101,8 +101,13 @@ export class WorkItemsComponent implements OnDestroy, OnInit  {
...
+      )
       .map(val => val.workItems)
+      .map(workItems => workItems.filter(item => {
```
maxandersen: third time duplication ;)

maxandersen:
I'm okey getting this change in if:

a) someone please verify that *all* logic on client side is okey we are now filtering out closed items in three different places - it feels like this would possible remove closed items where you might at times need them ?

b) please put the same hack into one place/method if possible so it does not deviate.

c) work should start ASAP from both platform and planner side on getting testbed in place so we avoid more back and forths.

src/app/home/work-item-widget/work-item-widget.component.ts
sanbornsen: No specific reason. I couldn't find any shared utils to put this method, but yeah I feel the problem and I'll create one.

sanbornsen:
> a) someone please verify that all logic on client side is okey we are now filtering out closed items in three different places - it feels like this would possible remove closed items where you might at times need them ?

A verification appreciated. In my knowledge by default, we don't want to show closed items anywhere.

maxandersen:
> A verification appreciated. In my knowledge by default, we don't want to show closed items anywhere.

@michaelkleinhenz can you help here ? I would personally not like it if I cannot find closed items as thats the only way to actually see what is done.

fabric8cd:
@sanbornsen fabric8/fabric8-ui:SNAPSHOT-PR-2882-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2882-fabric8-ui.badger.fabric8.io

maxandersen:
my concern (and it might be unwarrented) is that if this code removes closed workitems how do i actually see what work has been done ? how will this affect when filter has state = closed ? will it show empty list ?

michaelkleinhenz:
@maxandersen @sanbornsen that is correct, by default, we don't want to show closed items anywhere. The checkbox in Planner to enable showing of closed items is only a temporary solution until a query builder is in place.

michaelkleinhenz:
On the tests: someone wrote a client component, did not use the API correctly and did not write any tests for it. That implementation did break eventually and that issue was largely ignored - also because there were no tests in the first place.

Now the issue gets hot and someone else steps in to fix - which is not more than a very small bandaid for a component that should better be completely replaced. Yet that person gets blamed for not providing a full component test - which would need a complete testbed, which is a *lot* of work.

I don't understand this logic, sorry.

rgarg1:
I validated these changes and the fix appears to be working fine. Work items in the following states are no longer displayed on the OSIO dashboard page:  'removed', 'closed', 'done'

maxandersen:
now that we got it veriifed it *only* filters in the current dashboard widgets and got the code cleaned up I've gone ahead and merged this.

@joshuawilson lets get that testbed in place asap - planner team is happy to add those tests if can be pointed to how. if all is in place as you mention that should be easy to do with a simple PR showing what it is you were looking for.

joshuawilson:
The refactoring pulled the changes out into a new file and added a test. If the code was left where it was, you could have just added that test 3 times. This wasn't hard. 

We do have untested code, we are working on it. One way we work on it is to add a test for each function we touch. Just like this. It was not required that you write tests for every function in the file.

sanbornsen:
@joshuawilson not true. I just added test for a pure function. If I had it there I had to write a spec file for each of the components and they have injected providers which had to be mocked in the test, which I meant by preparing the test bed.

PR #2889, opened 2018-05-07
fix(jenkins-status): checks for Jenkins status before promoting

kishansagathiya:
Fix for https://github.com/openshiftio/openshift.io/issues/3356

### DIFF
--- a/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
+++ b/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
@@ -89,38 +89,50 @@ export class InputActionDialog {
 
   checkIfJenkinsIsUpAndProceedURL() {
     let url = 'https://jenkins.prod-preview.openshift.io';
-    console.log(this.build);
+    let notification;
     this.http.get(url)
       .subscribe((response) => {
-        if (response.status == 200) {
-          this.notifications.message({
-            message: `Got ${response.status}, Jenkins is up and running...`,
-            type: NotificationType.SUCCESS
-          } as Notification);
-          this.invokeUrl(this.inputAction.proceedUrl);
-          return;
-        } else if (response.status == 307 || response.status == 302) {
-          this.notifications.message({
-            message: `Got ${response.status}, Connecting to Jenkins...`,
-            type: NotificationType.INFO
-          } as Notification);
-        } else if (response.status == 503) {
-          this.notifications.message({
-            message: `Got ${response.status}, Cluster resources capacity is full. waiting along...`,
-            type: NotificationType.INFO
-          } as Notification);
-        } else if (response.status == 202) {
-          this.notifications.message({
-            message: `Got ${response.status}, Jenkins is currently idled. Please wait while we start it. waiting along...`,
-            type: NotificationType.WARNING
-          } as Notification);
-        } else {
-          this.notifications.message({
-            message: `Got status ${response.status}`,
-            type: NotificationType.DANGER
-          } as Notification);
-          return;
+        switch (response.status) {
+          case 200:
+            notification = {
+              message: `Got ${response.status}, Jenkins is up and running...`,
+              type: NotificationType.SUCCESS
+            };
+            this.invokeUrl(this.inputAction.proceedUrl);
+            break;
+          case 307:
+          case 302:
+            notification = {
+              message: `Got ${response.status}, Connecting to Jenkins.`,
+              type: NotificationType.INFO
+            };
+            break;
+          case 503:
+            notification = {
+              message: `Got ${response.status}, Cluster resources capacity is full. Waiting along.`,
+              type: NotificationType.INFO
+            };
+            break;
+          case 504:
+            notification = {
+              message: `Got ${response.status}, Request timed out. Trying again.`,
+              type: NotificationType.INFO
+            };
+            break;
+          case 202:
+            notification = {
+              message: `Got ${response.status}, Jenkins is currently idled. Please wait while we start it. waiting along...`,
+              type: NotificationType.WARNING
+            };
+            break;
+          default:
+            notification = {
+              message: `Got status ${response.status}`,
+              type: NotificationType.DANGER
+            };
+            break;
         }
+        this.notifications.message(notification as Notification);
         setTimeout (this.checkIfJenkinsIsUpAndProceedURL(), 10000);
       });
     this.close();
--- a/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
+++ b/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
@@ -4,6 +4,7 @@ import { AuthenticationService } from 'ngx-login-client';
 
 import { FABRIC8_FORGE_API_URL } from 'app/shared/runtime-console/fabric8-ui-forge-api';
 
+import { Notification, Notifications, NotificationType } from 'ngx-base';
 import { OnLogin } from '../../../../shared/onlogin.service';
 import { Build, PendingInputAction } from '../../../model/build.model';
 import { PipelineStage } from '../../../model/pipelinestage.model';
@@ -22,6 +23,7 @@ export class InputActionDialog {
   @ViewChild('inputModal') modal: any;
 
   constructor(private http: Http,
+              private notifications: Notifications,
               private authService: AuthenticationService,
               @Inject(FABRIC8_FORGE_API_URL) private forgeApiUrl: string
   ) {
@@ -90,20 +92,36 @@ export class InputActionDialog {
     this.http.get(url)
       .subscribe((response) => {
         if (response.status == 200) {
-          console.log('Got ' + response.status + ', Jenkins is up and running...');
+          this.notifications.message({
+            message: `Got ${response.status}, Jenkins is up and running...`,
+            type: NotificationType.SUCCESS
+          } as Notification);
           this.invokeUrl(this.inputAction.proceedUrl);
           return;
         } else if (response.status == 307 || response.status == 302) {
-          console.log('Got ' + response.status + ', Jenkins Proxy is logging you in...');
+          this.notifications.message({
+            message: `Got ${response.status}, Connecting to Jenkins...`,
+            type: NotificationType.INFO
+          } as Notification);
         } else if (response.status == 503) {
-          console.log('Got ' + response.status + ', Cluster resources capacity is full. waiting along..');
+          this.notifications.message({
+            message: `Got ${response.status}, Cluster resources capacity is full. waiting along...`,
+            type: NotificationType.INFO
+          } as Notification);
         } else if (response.status == 202) {
-          console.log('Got 202, Jenkins is currently idled. Please wait while we start it. waiting along..');
+          this.notifications.message({
+            message: `Got ${response.status}, Jenkins is currently idled. Please wait while we start it. waiting along...`,
+            type: NotificationType.WARNING
+          } as Notification);
         } else {
-          console.error ('Got status ' + response.status);
+          this.notifications.message({
+            message: `Got status ${response.status}`,
+            type: NotificationType.DANGER
+          } as Notification);
           return;
         }
         setTimeout (this.checkIfJenkinsIsUpAndProceedURL(), 10000);
       });
-    }
+    this.close();
+  }
 }
--- a/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
+++ b/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
@@ -146,7 +146,7 @@ export class InputActionDialog {
       access_token: this.authService.getToken(),
       token_type: 'Bearer'
     };
-    let url = this.build.jenkinsBuildURL + '?token_json=' + JSON.stringify(token_json);
+    let url = this.build.jenkinsBuildURL + '?token_json=' + encodeURIComponent(JSON.stringify(token_json));
     return this.http
       .get(url)
       .map((response: Response) => {
--- a/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
+++ b/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
@@ -89,6 +89,7 @@ export class InputActionDialog {
 
   checkIfJenkinsIsUpAndProceedURL() {
     let url = 'https://jenkins.prod-preview.openshift.io';
+    console.log(this.build);
     this.http.get(url)
       .subscribe((response) => {
         if (response.status == 200) {
--- a/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
+++ b/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
@@ -88,9 +88,8 @@ export class InputActionDialog {
   }
 
   checkIfJenkinsIsUpAndProceedURL() {
-    let url = 'https://jenkins.prod-preview.openshift.io';
     let notification;
-    this.http.get(url)
+    this.http.get(this.build.jenkinsBuildURL)
       .subscribe((response) => {
         switch (response.status) {
           case 200:
--- a/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
+++ b/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
@@ -98,8 +98,9 @@ export class InputActionDialog {
               message: `Got ${response.status}, Jenkins is up and running...`,
               type: NotificationType.SUCCESS
             };
+            this.notifications.message(notification as Notification);
             this.invokeUrl(this.inputAction.proceedUrl);
-            break;
+            return;
           case 307:
           case 302:
             notification = {
@@ -130,7 +131,8 @@ export class InputActionDialog {
               message: `Got status ${response.status}`,
               type: NotificationType.DANGER
             };
-            break;
+            this.notifications.message(notification as Notification);
+            return;
         }
         this.notifications.message(notification as Notification);
         setTimeout (this.checkIfJenkinsIsUpAndProceedURL(), 10000);
--- a/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
+++ b/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
@@ -1,5 +1,5 @@
 import { Component, Inject, ViewChild } from '@angular/core';
-import { Headers, Http, RequestOptions } from '@angular/http';
+import { Headers, Http, RequestOptions, Response } from '@angular/http';
 import { AuthenticationService } from 'ngx-login-client';
 
 import { FABRIC8_FORGE_API_URL } from 'app/shared/runtime-console/fabric8-ui-forge-api';
@@ -9,6 +9,8 @@ import { OnLogin } from '../../../../shared/onlogin.service';
 import { Build, PendingInputAction } from '../../../model/build.model';
 import { PipelineStage } from '../../../model/pipelinestage.model';
 import { pathJoin } from '../../../model/utils';
+
+import { Observable } from 'rxjs';
 
 @Component({
   selector: 'input-action-dialog',
@@ -89,7 +91,7 @@ export class InputActionDialog {
 
   checkIfJenkinsIsUpAndProceedURL() {
     let notification;
-    this.http.get(this.build.jenkinsBuildURL)
+    this.callJenkins()
       .subscribe((response) => {
         switch (response.status) {
           case 200:
@@ -137,5 +139,25 @@ export class InputActionDialog {
         setTimeout (this.checkIfJenkinsIsUpAndProceedURL(), 10000);
       });
     this.close();
+  }
+
+  callJenkins(): Observable<Response> {
+    const token_json = {
+      'access_token': this.authService.getToken(),
+      'token_type': 'Bearer'
+    };
+    let url = this.build.jenkinsBuildURL + JSON.stringify(token_json);
+    return this.http
+      .get(url)
+      .map((response: Response) => {
+        return response;
+      })
+      .catch((error) => {
+        this.notifications.message({
+          message: 'There seems to be a problem with Jenkins',
+          type: NotificationType.DANGER
+        } as Notification);
+        return Observable.throw(error.message || error);
+      });
   }
 }
--- a/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
+++ b/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
@@ -88,7 +88,7 @@ export class InputActionDialog {
   }
 
   checkIfJenkinsIsUpAndProceedURL() {
-    let url = 'https://jenkins.openshift.io';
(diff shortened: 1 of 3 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Thank you @kishansagathiya for this contribution!



src/app/dashboard-widgets/applications-widget/applications-list-item/applications-list-item.component.ts
```diff
@@ -84,13 +86,35 @@ export class ApplicationsListItemComponent implements DoCheck, OnInit {
     }
     let inputAction = input.build.firstPendingInputAction;
     if (isValidInputAction(inputAction) && input.build.jenkinsNamespace) {
+      console.log('####-12');
```
kishansagathiya: To be removed

src/app/dashboard-widgets/applications-widget/applications-list-item/applications-list-item.component.ts
```diff
@@ -84,13 +86,35 @@ export class ApplicationsListItemComponent implements DoCheck, OnInit {
...
+      this.checkIfJenkinsIsUp();
       this.inputActionDialog.build = input.build;
       this.inputActionDialog.inputAction = inputAction;
       this.inputActionDialog.stage = input.stage;
-      this.inputActionDialog.proceed();
     }
   }
 
+  checkIfJenkinsIsUp(): any {
+  let url = 'https://jenkins.openshift.io';
```
kishansagathiya: Not sure how to retrieve this URL instead of hardcoding it

joshuawilson:
Why is the package-lock file being updated?

kishansagathiya:
Do not review yet. I just realized that promote button is at two different places. I have to change the code.

src/app/dashboard-widgets/applications-widget/applications-list-item/applications-list-item.component.ts
```diff
@@ -84,13 +86,35 @@ export class ApplicationsListItemComponent implements DoCheck, OnInit {
...
+  this.http.get(url)
+    .subscribe((response) => {
+      if (response.status == 200) {
+        console.log('Got ' + response.status + ', Jenkins is up and running...');
+        this.inputActionDialog.proceed();
+        return true;
+      }
+      if (response.status == 307 || response.status == 302) {
+        console.log('Got ' + response.status + ', Jenkins Proxy is logging you in...');
```
gastaldi: No indication in the UI that this is happening?

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
```diff
@@ -84,4 +84,23 @@ export class InputActionDialog {
...
+  checkIfJenkinsIsUp() {
+    let url = 'https://jenkins.openshift.io';
+    this.http.get(url)
+      .subscribe((response) => {
+        if (response.status == 200) {
+          console.log('Got ' + response.status + ', Jenkins is up and running...');
+          this.invokeUrl(this.inputAction.proceedUrl);
+          return;
+        } else if (response.status == 307 || response.status == 302) {
+          console.log('Got ' + response.status + ', Jenkins Proxy is logging you in...');
```
gastaldi: No indication in the UI that this happened?

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
```diff
@@ -84,4 +84,23 @@ export class InputActionDialog {
   close() {
     this.modal.close();
   }
+
+  checkIfJenkinsIsUp() {
+    let url = 'https://jenkins.openshift.io';
```
gastaldi: A new environment needs to be added to F8UI (the Jenkins URL)

kishansagathiya: I wanted to inject an environment variable like https://github.com/fabric8-ui/fabric8-ui/blob/master/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts#L5-L26
Where and how do I do this

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
rohitkrai03: @gastaldi We could probably use NotificationService to show these messages to user rather than just logging it to console. @kishansagathiya What do you think?

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
gastaldi: @kishansagathiya I think @edewit or @joshuawilson may know this better than me :wink:

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
```diff
@@ -84,4 +84,25 @@ export class InputActionDialog {
...
+        } else if (response.status == 307 || response.status == 302) {
+          console.log('Got ' + response.status + ', Jenkins Proxy is logging you in...');
+        } else if (response.status == 503) {
+          console.log('Got ' + response.status + ', Cluster resources capacity is full. waiting along..');
+        } else if (response.status == 202) {
+          console.log('Got 202, Jenkins is currently idled. Please wait while we start it. waiting along..');
+        } else {
+          console.error ('Got status ' + response.status);
+        }
+        setTimeout (this.checkIfJenkinsIsUp(), 10000);
```
lordofthejars: Looks like infinite loop isn't it? It just finishes if we get a `200`. Shouldn't we finish this loop if we get an unexpected error? And also add some feedback to user? Something like spinning animation in case of unidling?

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
lordofthejars: yes some kind of feedback will be great so user knows that he has pushed the button and we are working on it.

kishansagathiya:
@joshuawilson 
Have tested this locally, 
With prod-preview Jenkins proxy I was able to get the proper status code
but on prod Jenkins proxy it gave me Cross-Origin Resource Sharing error.
Since once deployed domain would be same, so we won't see this. There might be few things that I might be missing in this PR (like hardcoded URL). I don't know how to do that.
Somebody with UI experience would be able to do it much better and faster than.

src/app/dashboard-widgets/applications-widget/applications-list-item/applications-list-item.component.ts
```diff
@@ -90,7 +90,6 @@ export class ApplicationsListItemComponent implements DoCheck, OnInit {
       this.inputActionDialog.proceed();
     }
   }
-
```
joshuawilson: why was this line removed? It is there for readability?

kishansagathiya: Unintentional. I had changed this file. Might have left this. Let get it how it was

kishansagathiya: Yup, I know why. It gives me linting errors without that.
93,1 trailing whitespace

jiekang:
In theory CORS won't be hit on production since it will be openshift.io making a request to jenkins.openshift.io. Unfortunately allowing CORS is a server-side thing so testing it locally does not seem possible.

rohitkrai03:
@jiekang Since most of the API URLs we have come from external dependencies like ngx-login-client or ngx-fabric8-wit, i am not sure what's the best place to put the jenkins URL.

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
maxandersen: lets not reveal internal implementation details like "Jenkins proxy" to user in the UI here. 

Better to just say "Connecting to Jenkins..." or similar.

jiekang:
Hmm, a few issues varying in degree of severity. @rohitkrai03 

1. Basic GET request to jenkins.openshift.io **always** results in 307 response, regardless of state of jenkins pods. I don't know there's another endpoint that can be used but this one may not be correct. The request also doesn't have any auth header but I'm doubtful it would change anything.
2. Clicking promote no longer runs `this.close()` so the modal is not closed. It should close at the end, regardless of success.
3. I agree with your earlier comment regarding using NotificationService to inform users what happened after pressing promote.
4. This is a band-aid and I hope the issue is tracked for a proper long-term solution. In the event a user closes the tab containing this application, this repeated setTimeout will be destroyed and Promote will not occur.

jiekang:
Also when I ran locally targeting production, I was able to see the request made when hitting 'Promote' and I didn't have a CORS issue, so I guess my earlier comment re. CORS can be disregarded. Sorry about that;;

kishansagathiya:
jenkins.openshift.io?token_json=$OSIO_TOKEN_JSON should return status code
other than 307

jiekang:
Ah it looks like initial request is okay but the CORS comes in due to the 307 as it doesn't allow the redirect to `https://auth.openshift.io/api/login?redirect=https:%2F%2Fjenkins.openshift.io%2F`

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
```diff
@@ -84,4 +86,42 @@ export class InputActionDialog {
...
+            type: NotificationType.SUCCESS
+          } as Notification);
+          this.invokeUrl(this.inputAction.proceedUrl);
+          return;
+        } else if (response.status == 307 || response.status == 302) {
+          this.notifications.message({
+            message: `Got ${response.status}, Connecting to Jenkins...`,
+            type: NotificationType.INFO
+          } as Notification);
+        } else if (response.status == 503) {
```
kishansagathiya: We need one more retry logic for status is 504 Gateway Timeout

fabric8cd:
@kishansagathiya fabric8/fabric8-ui:SNAPSHOT-PR-2889-9 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2889-fabric8-ui.badger.fabric8.io

kishansagathiya:
Todo
1. Add one more condition for retry (if status code is 504 Gateway timeout)
2. Remove hardcoded url and  inject it from environment variable 
3 . Add Tests

fabric8cd:
@kishansagathiya fabric8/fabric8-ui:SNAPSHOT-PR-2889-10 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2889-fabric8-ui.badger.fabric8.io

fabric8cd:
@kishansagathiya fabric8/fabric8-ui:SNAPSHOT-PR-2889-11 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2889-fabric8-ui.badger.fabric8.io

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
```diff
@@ -84,4 +86,55 @@ export class InputActionDialog {
...
+            };
+            break;
+          default:
+            notification = {
+              message: `Got status ${response.status}`,
+              type: NotificationType.DANGER
+            };
+            break;
+        }
+        this.notifications.message(notification as Notification);
```
lordofthejars: Notice that now there is no return as it happens in if/else so infinite loop might occurs because break only occurs in switch context.

fabric8cd:
@kishansagathiya fabric8/fabric8-ui:SNAPSHOT-PR-2889-12 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2889-fabric8-ui.badger.fabric8.io

fabric8cd:
@kishansagathiya fabric8/fabric8-ui:SNAPSHOT-PR-2889-13 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2889-fabric8-ui.badger.fabric8.io

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
```diff
@@ -88,7 +88,8 @@ export class InputActionDialog {
   }
 
   checkIfJenkinsIsUpAndProceedURL() {
-    let url = 'https://jenkins.openshift.io';
+    let url = 'https://jenkins.prod-preview.openshift.io';
```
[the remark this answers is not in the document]
joshuawilson: We need to set the URL via the env var that are set with SaaSHearder in https://github.com/openshiftio/saas-openshiftio/blob/master/dsaas-services/fabric8-ui.yaml#L10
and referenced like https://github.com/fabric8-ui/fabric8-ui/blob/master/openshift/fabric8-ui.app.yaml#L57

fabric8cd:
@kishansagathiya fabric8/fabric8-ui:SNAPSHOT-PR-2889-15 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2889-fabric8-ui.badger.fabric8.io

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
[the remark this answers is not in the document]
sbose78: Looked everywhere but didn't find where we are setting the jenkins proxy url.
What we found is that `this.build.jenkinsBuildURL` in the current context has the right jenkins proxy url which can be hit to un-idle.


```
let url = this.build.jenkinsBuildURL + '?token_json=' + JSON.stringify(token_json);
``` 
is being used to call the jenkins proxy with the user's token - this will trigger un-idling.

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
joshuawilson: This only sets the `notification` variable. There is nothing returned at all. The `default` sets it too. This looks correct to me.

src/app/dashboard-widgets/applications-widget/applications-list-item/applications-list-item.component.ts
joshuawilson: Is there a single white space in the line? If so, just remove the single space and leave the line.

kishansagathiya:
@joshuawilson 
Let's **NOT** merge this until the demo is done. This should not affect the demo. This might affect users during workshops and new users which will be added.

rohitkrai03:
@kishansagathiya I've added a few tests for the component. I think all the todo items have been taken care of now. Do let me know if there is anything else that needs to be done. cc: @joshuawilson

fabric8cd:
@kishansagathiya fabric8/fabric8-ui:SNAPSHOT-PR-2889-16 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2889-fabric8-ui.badger.fabric8.io

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
```diff
@@ -22,6 +25,7 @@ export class InputActionDialog {
   @ViewChild('inputModal') modal: any;
 
   constructor(private http: Http,
+              private notifications: Notifications,
               private authService: AuthenticationService,
               @Inject(FABRIC8_FORGE_API_URL) private forgeApiUrl: string
```
gastaldi: I believe `FABRIC8_FORGE_API_URL` needs to be completely removed, since it is no longer used

joshuawilson:
@jiekang can you please take another look too.

jiekang:
Okay, taking a look now.

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
```diff
@@ -84,4 +88,76 @@ export class InputActionDialog {
...
+            break;
+          case 504:
+            notification = {
+              message: `Got ${response.status}, Request timed out. Trying again.`,
+              type: NotificationType.WARNING
+            };
+            break;
+          case 202:
+            notification = {
+              message: `Got ${response.status}, Jenkins is currently idled. Please wait while we start it. waiting along.`,
```
jiekang: s/Please wait while we start it. waiting along./Please wait while it starts./

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
```diff
@@ -84,4 +88,76 @@ export class InputActionDialog {
...
+    };
+    let url = this.build.jenkinsBuildURL + '?token_json=' + encodeURIComponent(JSON.stringify(token_json));
+    return this.http
+      .get(url)
+      .map((response: Response) => {
+        return response;
+      })
+      .catch((error) => {
+        this.notifications.message({
+          message: 'There seems to be a problem with Jenkins',
```
jiekang: In a local instance connected to production, a redirect still occurs that results in this error being hit and unidling behavior not occurring. 

pipelines-redirect

The only way to fix this is to configure the server to send the right response header, but that's not an option. It may work in production though.

It looks like the jenkins proxy is configured s.t.
1. If a request to jenkins.openshift.io/*, redirect to auth.
2. If redirected to auth, redirect back to jenkins.openshift.io/* with ?token set
3. If a request to jenkins.openshift.io/* with ?token set, redirect to jenkins.openshift.io/*

It would be nice if there was an endpoint that didn't involve redirecting as long as the auth header was set with the correct token. Maybe the entire redirect mechanism can be looked at too, it's a bit long and browsers will show 'Too many redirects' even though it eventually goes to the right place. I digress...

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
```diff
@@ -84,4 +88,76 @@ export class InputActionDialog {
...
+
+  callJenkins(): Observable<Response> {
+    const token_json = {
+      access_token: this.authService.getToken(),
+      token_type: 'Bearer'
+    };
+    let url = this.build.jenkinsBuildURL + '?token_json=' + encodeURIComponent(JSON.stringify(token_json));
+    return this.http
+      .get(url)
+      .map((response: Response) => {
```
jiekang: Using `map` to take `response` and return` response` is redundant, `http.get()` already returns the Response object, no?

`http.get(...).error(...)` should do the same thing.

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
jiekang: Also something like: s/There seems to be a problem with Jenkins/There is a problem accessing Jenkins./

It's best to avoid 'probably' messages ('seems' in this case) for the user. It's not as polished.

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
```diff
@@ -84,4 +88,76 @@ export class InputActionDialog {
...
+          case 200:
+            notification = {
+              message: `Got ${response.status}, Jenkins is up and running.`,
+              type: NotificationType.SUCCESS
+            };
+            this.notifications.message(notification as Notification);
+            this.invokeUrl(this.inputAction.proceedUrl);
+            return;
+          case 307:
+          case 302:
```
jiekang: I think the default behavior of the http library when hitting a 302 from `get(...)` is to follow the redirect automatically. I think we've observed this in practice here as well. So theoretically, it won't ever be 302. Up to you to keep this or not.

rohitkrai03:
@jiekang I tried writing tests for other status codes but ran into a problem because all the others cases call `setTimeout(this.checkIfJenkinsIsUpAndProceedURL(), 10000);` which makes it run out of stack in a test case. Is there a way to test this?

jiekang:
@rohitkrai03 Oh that's fun :) I think that might also be an issue in production. If jenkins never hits 200, will a user also run out of stack, not just in tests? Maybe it would be best to only retry a set number of times, maybe 5 or 10 and then prompt the user to try again later. Thanks for trying to write tests here!

src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
```diff
@@ -84,4 +88,73 @@ export class InputActionDialog {
...
+              type: NotificationType.WARNING
+            };
+            break;
+          default:
+            notification = {
+              message: `Got status ${response.status}`,
+              type: NotificationType.DANGER
+            };
+            this.notifications.message(notification as Notification);
+            this.close();
```
jiekang: This is within an asynchronous block called from `checkIfJenkinsIsUpAndProceedURL()`. This function always calls `this.close()` before returning (line 141) so I don't think the default case needs to also call `this.close()`. Thoughts?

rohitkrai03: @jiekang I agree. Removing `this.close()` from default case. I added it earlier thinking that if default case returns then the modal might not be closed but that's not the case since switch is under .subscribe() and `this.close()` in the end always gets called.

jiekang:
@rohitkrai03 yes I think limiting retries would be much preferred then. If you want to dive into Observables, I think you can restructure this into an Observable that retries via `observable.retryWhen(...)` instead of using a setTimeout on the function. That setTimeout will also result in `this.close()` being called many times, which is probably safe but not the cleanest.

Also now that I think about it, the behavior of Observable returned by `http.get()` might need to be investigated. If it recieves an error status 4xx or 5xx, will it call the success function or the error function? I have a feeling it might call the error function so the `switch` for 4xx and 5xx errors is never hit.

fabric8cd:
@kishansagathiya fabric8/fabric8-ui:SNAPSHOT-PR-2889-18 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2889-fabric8-ui.badger.fabric8.io

kishansagathiya:
Does https://github.com/fabric8-services/fabric8-jenkins-proxy/issues/257 affect merge of this PR in any way?

jiekang:
I think if https://github.com/fabric8-services/fabric8-jenkins-proxy/issues/257 can be resolved positively than it's likely that this merge will actually work in production or prod-preview. There's still no way to test this locally but it's not necessarily a blocker as merging and testing on prod-preview is possible.

joshuawilson:
I'm going to merge this since we need to test it in prod-preview to fully check it. @rohitkrai03 please create an issue (and/or PR to fix the infinite issue so they don't run out of stack).

jiekang:
So it looks like the CORS issue still occurs on prod-preview:

```
`Failed to load https://jenkins.prod-preview.openshift.io/job/chrislessard/job/osio-prod-2/job/master/2/?token_json=<token>: Redirect from 'https://jenkins.prod-preview.openshift.io/job/chrislessard/job/osio-prod-2/job/master/2/?token_json=<token>' to 'https://jenkins.prod-preview.openshift.io/job/chrislessard/job/osio-prod-2/job/master/2/' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'https://prod-preview.openshift.io' is therefore not allowed access.`
```

jiekang:
This means for this to work at all, https://github.com/fabric8-services/fabric8-jenkins-proxy/issues/257 needs to be positively resolved.

PR #2926, opened 2018-05-21
Remove hardcoded data in template

piyush1594:
Fixes https://github.com/openshiftio/openshift.io/issues/3431


joshuawilson:
This is something you can only see under Experimental. That flag is for work in progress. 

That field needs to stay so that the developer knows where to put the correct data when it get's implemented. If you want to work with @mindreeper2420 to implement it, please submit a new PR.

PR #2928, opened 2018-05-22
fix(version): update ngx-forge to 0.0.72

fabric8cd:


joshuawilson:
@invincibleJai @jyasveer @arunkumars08 Can you please fix your project to update the https://github.com/fabric8-ui/fabric8-npm-dependencies/blob/master/package.json

You set this here and since it is different what is in master it tries to update whenever anyone updates a library. I have been closing the PRs. If you do not intend to use this then I will remove it from the dependencies list.

PR #2935, opened 2018-05-23
feat(deployments): introduce empty-state screen for when there are no…

aptmac:
… deployments

This PR addresses [issue 3127](https://github.com/openshiftio/openshift.io/issues/3127) [0], which introduces an empty-state page to the deployments page for the event when there applications but none of them are deployed. This implementation is based on the [ui mockups](https://redhat.invisionapp.com/share/YSDL0KRKW#/screens/286564375) [1].

Here is what the deployments page looks like with this patch applied:
after

Testing this PR should be "easy" at the moment because the vertx starter app is broken (for me at least) and will fail during the early stages of the pipeline build. As a result, it's not too difficult to create applications that will not be deployed, and the deployments page should appear as seen in the image above.

This PR required some modifications to the deployments card container component, specifically to the layout of the application names and deployment cards. When there are active deployments, the application name and deployment cards fill a `row`, but this can be troublesome because if there are no deployment cards instead we need a column of application names and then the empty state image + instructions. To deal with this, the `emptystate` template allocates a specific amount of space to the application titles, and then the empty state image + description fills in to the right of it. When there are applications present, the environment title cards are displayed and the application title + deployment cards are created as before. Let me know if this is okay, the designs for this empty-state page wasn't as plug-n-play as the one for no applications [3] and required more html modifications.

To include the functionality [3] for the hyperlinks in the empty state text, the `UserService` has been introduced for the purposes of generating the url to the user's pipelines page, and the `spaceName` is now passed from the deployments component. In order to verify if there are active deployments, a function was added to the deployments service called `isApplicationsDeployed` which returns an observable of a boolean when passed the spaceId and environments to check.

[0] https://github.com/openshiftio/openshift.io/issues/3127
[1] https://redhat.invisionapp.com/share/YSDL0KRKW#/screens/286564375
[2] https://github.com/openshiftio/openshift.io/issues/3126
[3] https://github.com/openshiftio/openshift.io/issues/2188#issuecomment-385427415


src/app/space/create/deployments/apps/deployment-card-container.component.html
```diff
@@ -1,18 +1,54 @@
...
+          </a>
+        </h4>
+      </div>
+    </div>
+    <div class="deployments-empty-state col-sm-4 col-sm-offset-2">
+      <img src="../../../../../assets/images/deployments_emptystate_nodeployments.png"/>
+      <div class="deployments-empty-state-text">
+        <h2 class="deployments-empty-state-text-heading">None of your applications have been deployed yet.</h2>
+        <p>
+          Check for upcoming deployments on the <a [routerLink]="['/', username, spaceName, 'create', 'pipelines']">Pipelines</a> page,
```
andrewazores: I'd prefer to use a "contextPath" (ie inject `Contexts` and use `contexts.current.map(c => c.path)`) rather than `'/', username, spaceName`). Using the current logged in username means this link will be invalid if you're looking at deployments of a space owned by another user.

src/app/space/create/deployments/apps/deployment-card-container.component.html
```diff
@@ -1,18 +1,54 @@
...
+          and use the <a [routerLink]="['/', username, '_settings', 'resources']">Resource Usage panel</a> to verify you have the space for another deployment.
```
andrewazores: Can we check with the design team if this sentence should be displayed in the case where the user viewing the page is not the user that owns the space?

src/app/space/create/deployments/apps/deployment-card-container.component.less
```diff
@@ -8,3 +8,18 @@
...
+  text-align: center;
+  margin-top: 60px;
+  margin-bottom: 80px;
+}
+
+.deployments-empty-state-text-heading {
+  font-weight: 500;
+}
+
+#notDeployedApplicationTitle {
```
andrewazores: Can this be done with a class rather than an id?

src/app/space/create/deployments/apps/deployment-card-container.component.ts
```diff
@@ -14,7 +16,22 @@ import { Environment } from '../models/environment';
 })
 export class DeploymentCardContainerComponent {
   @Input() spaceId: string;
+  @Input() spaceName: string;
   @Input() environments: Observable<string[]>;
-  @Input() application: string;
-  collapsed: boolean = true;
+  @Input() applicationsList: string[];
+  collapsed: {[applicationName: string]: boolean} = {};
```
andrewazores: Can a Map be used here rather than an array? Or does that make accessing it from within the template a lot uglier?

src/app/space/create/deployments/apps/deployment-card-container.component.ts
```diff
@@ -14,7 +16,22 @@ import { Environment } from '../models/environment';
 })
 export class DeploymentCardContainerComponent {
   @Input() spaceId: string;
+  @Input() spaceName: string;
   @Input() environments: Observable<string[]>;
-  @Input() application: string;
-  collapsed: boolean = true;
+  @Input() applicationsList: string[];
```
andrewazores: `applications` rather than `applicationsList` is fine IMO.

src/app/space/create/deployments/apps/deployment-card-container.component.ts
```diff
@@ -14,7 +16,22 @@ import { Environment } from '../models/environment';
...
+  collapsed: {[applicationName: string]: boolean} = {};
+  private isApplicationsDeployed: Observable<boolean>;
+  private username: string;
+
+  constructor(private userService: UserService, private deploymentsService: DeploymentsService) {}
```
andrewazores: Please break each constructor argument onto a new line here, ie

```
constructor(
  private userService: UserService,
  private deploymentsService: DeploymentsService
) {}
```

src/app/space/create/deployments/services/deployments.service.ts
```diff
@@ -129,6 +129,18 @@ export class DeploymentsService implements OnDestroy {
       .distinctUntilChanged();
   }
 
+  isApplicationsDeployed(spaceId: string, environments: string[]): Observable<boolean> {
```
andrewazores: `isApplicationsDeployed` is a pretty awkward name. `hasDeployments` perhaps?

src/app/space/create/deployments/services/deployments.service.ts
```diff
@@ -129,6 +129,18 @@ export class DeploymentsService implements OnDestroy {
...
+    let bool = false;
+    for (let i = 0; i < environments.length; i++) {
+      this.isDeployedInEnvironment(spaceId, environments[i]).subscribe(b => {
+        if (b) {
+          bool = true;
+        }
+      });
+    }
+    return Observable.of(bool);
+  }
```
andrewazores: This doesn't look like it's going to work properly. `isDeployedInEnvironment` is async, but you're looping and returning synchronously. You should be using RxJS streaming operators for this logic, something like:

```
Observable.of(environments)
  .map((environment: string): [string, string] => [spaceId, environment])
  .flatMap((pair: [string, string]): Observable<boolean> => this.isDeployedInEnvironment(pair[0], pair[1]))
  .every((deployed: boolean): boolean => deployed);
```
```

Comment (andrewazores) [reply src/app/space/create/deployments/services/deployments.service.ts]:
```
Observable.combineLatest(
  environments.map(environment => this.isDeployedInEnvironment(spaceId, environment))
)
  .map((deployed: boolean[]): boolean => deployed.some(v => v))
```

is probably more like it.

Comment (aptmac) [reply src/app/space/create/deployments/services/deployments.service.ts]: Good call, that's a nice solution thanks - I've gone forward with your function in this latest revision.

File: src/app/space/create/deployments/apps/deployment-card-container.component.html (hunk 1 above)
Comment (aptmac) [reply src/app/space/create/deployments/apps/deployment-card-container.component.html]: Ah good call, I've gone with the contexts in this next revision.

File: src/app/space/create/deployments/services/deployments.service.spec.ts [hunk 9]
```diff
@@ -782,6 +782,38 @@ describe('DeploymentsService', () => {
     });
   });
 
+  describe('#hasDeployments', () => {
+    it('should return true if there are deployed applications', (done: DoneFn) => {
+      spyOn(svc, 'isDeployedInEnvironment').and.returnValue(Observable.of(true));
```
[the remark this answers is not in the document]
aptmac: Yeah sure, I've adjusted the tests to work with the api service instead.

src/app/space/create/deployments/apps/deployment-card-container.component.spec.ts
```diff
@@ -35,34 +39,83 @@ class FakeDeploymentCardComponent {
...
+          { provide: Contexts, useValue: mockContexts },
+          { provide: DeploymentsService, useValue: mockDeploymentsService }
+        ],
+        schemas: [NO_ERRORS_SCHEMA]
+      },
+      (component: DeploymentCardContainerComponent) => {
+        component.spaceId = 'space';
+        component.environments = Observable.of(environments);
+        component.applications = ['app'];
+        component.collapsed['app'] = true;
```
andrewazores: Does this `collapsed` property need to be set here? IIRC this customizer hook gets run before the `ngOnInit` lifecycle hook does (or that was the intention when I added the customizer function), so I think the component should take of initializing its own `collapsed` map property.

src/app/space/create/deployments/apps/deployment-card-container.component.spec.ts
```diff
@@ -35,34 +39,83 @@ class FakeDeploymentCardComponent {
...
```
andrewazores: Same as above.

src/app/space/create/deployments/apps/deployment-card-container.component.spec.ts
```diff
@@ -35,34 +39,83 @@ class FakeDeploymentCardComponent {
...
+      const arrayOfComponents: DebugElement[] =
+        this.fixture.debugElement.queryAll(By.directive(FakeDeploymentCardComponent));
+      expect(arrayOfComponents.length).toEqual(environments.length);
 
-    environments.forEach((envData: string, index: number) => {
-      const cardComponent = arrayOfComponents[index].componentInstance;
-      expect(cardComponent.applicationId).toEqual('app');
-      expect(cardComponent.environment).toEqual(environments[index]);
+      environments.forEach((envData: string, index: number) => {
+        const cardComponent = arrayOfComponents[index].componentInstance;
```
andrewazores: Missing type annotation.

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2935-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2935-fabric8-ui.badger.fabric8.io

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2935-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2935-fabric8-ui.badger.fabric8.io

mindreeper2420:
@aptmac can the expansions carets be removed from the application names when this empty state is displayed? It makes it looks like I can expand something that isn't there :-) Otherwise, it looks good!

aptmac:
@mindreeper2420 Yeah sure that shouldn't take long, I'll add it as an extra commit here.

andrewazores:
Looks good to me.

PR #2938, opened 2018-05-24
fix(launcher): Disable continue button on load

jyasveer:
- Don't populate the application name by default
- Let the user enter it
- Disable the `continue` button by default.
- Includes tests

**Addresses** https://github.com/openshiftio/openshift.io/issues/3607

### DIFF
--- a/src/app/space/add-app-overlay/add-app-overlay.component.ts
+++ b/src/app/space/add-app-overlay/add-app-overlay.component.ts
@@ -23,14 +23,15 @@ import { Application, DeploymentApiService } from '../create/deployments/service
 })
 export class AddAppOverlayComponent implements OnDestroy, OnInit {
   currentSpace: Space;
-  isProjectNameValid: boolean;
+  isProjectNameValid = false;
   loggedInUser: User;
   projectName: string = '';
   selectedFlow: string = '';
   spaces: Space[] = [];
   subscriptions: Subscription[] = [];
   applications: string[] = [];
-  isProjectNameAvailable: boolean;
+  isProjectNameAvailable = false;
+  onLoad = true;
 
   constructor(private context: ContextService,
               private dependencyCheckService: DependencyCheckService,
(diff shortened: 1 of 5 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Thank you @jyasveer for this contribution!



src/app/space/add-app-overlay/add-app-overlay.component.ts
```diff
@@ -23,14 +23,15 @@ import { Application, DeploymentApiService } from '../create/deployments/service
 })
 export class AddAppOverlayComponent implements OnDestroy, OnInit {
   currentSpace: Space;
-  isProjectNameValid: boolean;
+  isProjectNameValid = false;
```
joshuawilson: why did you remove the type?

src/app/space/add-app-overlay/add-app-overlay.component.ts
```diff
@@ -23,14 +23,15 @@ import { Application, DeploymentApiService } from '../create/deployments/service
...
   loggedInUser: User;
   projectName: string = '';
   selectedFlow: string = '';
   spaces: Space[] = [];
   subscriptions: Subscription[] = [];
   applications: string[] = [];
-  isProjectNameAvailable: boolean;
+  isProjectNameAvailable = false;
```
joshuawilson: needs a type

src/app/space/add-app-overlay/add-app-overlay.component.ts
jyasveer: when we assign a primitive value, type is not needed by default. Type is defined from the primitive value we assign

src/app/space/add-app-overlay/add-app-overlay.component.ts
jyasveer: when we assign a primitive value, type is not needed by default. Type is defined from the primitive value we assign

src/app/space/add-app-overlay/add-app-overlay.component.ts
joshuawilson: I prefer the verbose style but yes you are right, that works too.

PR #2941, opened 2018-05-24
fix(deployments): move http error handling from deployments service to deployments api service

jiekang:
This refactor is a part of addressing https://github.com/openshiftio/openshift.io/issues/2713

A follow up PR will then rework the notification handling. Thoughts?


src/app/space/create/deployments/services/deployment-api.service.ts
```diff
@@ -138,7 +140,9 @@ export class DeploymentApiService {
   constructor(
     private http: Http,
     @Inject(WIT_API_URL) private witUrl: string,
-    private auth: AuthenticationService
+    private auth: AuthenticationService,
+    private readonly logger: Logger,
+    private readonly errorHandler: ErrorHandler,
```
andrewazores: May as well make the other properties here `readonly` too then

PR #2943, opened 2018-05-25
fix(version): Update planner version to 0.48.1

jarifibrahim:
NOTE - The changes to package-lock.json are a result of `npm install`. The transitive dependencies have been updated because `rh-ngx-datatable` version is updated in planner.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4912,9 +4912,9 @@
       "dev": true
     },
     "fabric8-planner": {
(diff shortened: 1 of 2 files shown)


jarifibrahim:
[test]

sanbornsen:
Last time also CICO passed but master failed.

joshuawilson:
It appears to run clean locally, so I'm going to approve it.

PR #2947, opened 2018-05-25
fix(webpack): fix webpack config which prevented updating ngx-bootstrap

christianvogt:
A while back ngx-bootstrap was updated to 1.9.3 and then downgraded back to 1.8.0.

Without fixing the webpack config, ngx-bootstrap 1.9.3 would throw errors because a rule that was meant for moment also matched a module in ngx-bootstrap 1.9.3. I updated the regex pattern to avoid matching ngx-bootstrap.

I had to update a test when upgrading ngx-bootstrap.

Both launcher and planner already depend on ngx-bootstrap 1.9.3 for popover outsideClick functionality.

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -474,7 +474,7 @@ module.exports = function (options) {
 
       extractCSS,
 
-      new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/),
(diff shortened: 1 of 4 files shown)


christianvogt:
@joshuawilson updated package-lock.json

PR #2966, opened 2018-05-29
fix(work-item-widget): fix the component to display all of a user's spaces instea…

aptmac:
…d of only recent spaces

This PR addresses issues [3621](https://github.com/openshiftio/openshift.io/issues/3621) [0] and [3504](https://github.com/openshiftio/openshift.io/issues/3504) [1], in which the home page work items widget only displays a subset of the user's spaces in the dropdown.

This is due to the widget only retrieving a list of 'recent' spaces via the `Spaces.current`[2]. As a result, only a subset of the user's spaces are displayed in the dropdown, which could be confusing because the title of the widget doesn't make reference to it only displaying recent spaces work items. This has been addressed by using the `SpaceService` to retrieve a list of the user's spaces instead.

Before:
before

After:
after-1

[0] https://github.com/openshiftio/openshift.io/issues/3621
[1] https://github.com/openshiftio/openshift.io/issues/3504
[2] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/home/work-item-widget/work-item-widget.component.ts#L32

### DIFF
--- a/src/app/home/work-item-widget/work-item-widget.component.html
+++ b/src/app/home/work-item-widget/work-item-widget.component.html
@@ -3,13 +3,13 @@
     <div class="card-pf-heading-bg">
       <div class="card-pf-heading f8-card-heading">
         <h2 class="card-pf-title">My Work Items
-        <span *ngIf="recentSpaces.length !== 0">
+        <span *ngIf="spaces.length !== 0">
           <select class="col-xs-3 col-sm-6 form-control work-item-combobox"
-              [disabled]="recentSpaces.length === 0"
+              [disabled]="spaces.length === 0"
               [(ngModel)]="currentSpaceId"
               (ngModelChange)="fetchWorkItems()">
             <option value="default">Select a space to work with...</option>
-            <option *ngFor="let space of recentSpaces" [value]="space.id">
+            <option *ngFor="let space of spaces" [value]="space.id">
               {{space.attributes.name | spaceName}}
             </option>
           </select>
@@ -57,13 +57,13 @@
     <div class="card-pf-heading-bg">
       <div class="card-pf-heading f8-card-heading">
         <h2 class="card-pf-title">My Work Items
-          <span *ngIf="recentSpaces.length !== 0">
+          <span *ngIf="spaces.length !== 0">
             <select class="col-xs-3 col-sm-6 form-control work-item-combobox"
-                    [disabled]="recentSpaces.length === 0"
+                    [disabled]="spaces.length === 0"
                     [(ngModel)]="currentSpaceId"
                     (ngModelChange)="fetchWorkItems()">
               <option value="default">Select a space to work with...</option>
-              <option *ngFor="let space of recentSpaces" [value]="space.id">
+              <option *ngFor="let space of spaces" [value]="space.id">
                 {{space.attributes.name | spaceName}}
               </option>
             </select>
--- a/src/app/home/work-item-widget/work-item-widget.component.ts
+++ b/src/app/home/work-item-widget/work-item-widget.component.ts
@@ -1,7 +1,7 @@
 import { Component, OnDestroy, OnInit, ViewEncapsulation } from '@angular/core';
 
 import { WorkItem, WorkItemService } from 'fabric8-planner';
-import { Space, Spaces } from 'ngx-fabric8-wit';
+import { Space, Spaces, SpaceService } from 'ngx-fabric8-wit';
 import { User, UserService } from 'ngx-login-client';
 import { Subscription } from 'rxjs';
 
@@ -20,15 +20,19 @@ export class WorkItemWidgetComponent implements OnDestroy, OnInit  {
   loggedInUser: User;
   recentSpaces: Space[] = [];
   recentSpaceIndex: number = 0;
+  spaces: Space[] = [];
   subscriptions: Subscription[] = [];
   workItems: WorkItem[] = [];
 
   constructor(
+      private spaceService: SpaceService,
       private spacesService: Spaces,
       private workItemService: WorkItemService,
       private userService: UserService) {
-    this.subscriptions.push(userService.loggedInUser.subscribe(user => {
-      this.loggedInUser = user;
+    this.loggedInUser = this.userService.currentLoggedInUser;
+    this.subscriptions.push(this.spaceService.getSpacesByUser(this.loggedInUser.attributes.username).subscribe(spaces => {
+      this.spaces = spaces;
+      this.loading = false;
     }));
     this.subscriptions.push(spacesService.recent.subscribe(spaces => {
       this.recentSpaces = spaces;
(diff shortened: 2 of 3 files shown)


joshuawilson:
Please rebase.

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2966-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2966-fabric8-ui.badger.fabric8.io

joshuawilson:
@dlabrecq can you please review.

src/app/home/work-item-widget/work-item-widget.component.html
```diff
@@ -3,13 +3,13 @@
     <div class="card-pf-heading-bg">
       <div class="card-pf-heading f8-card-heading">
         <h2 class="card-pf-title">My Work Items
-        <span *ngIf="recentSpaces.length !== 0">
+        <span *ngIf="spaces.length !== 0">
```
dlabrecq: Is there a specific issue about using recent spaces? Issues #3621 and #3504 don't appear to be related.

Using recent spaces was intentional here to limit the number of potential options shown in the drop down menu.

@mindreeper2420 @catrobson

catrobson:
This dropdown should show all spaces that I am a collaborator in.  It looks like now we're going to show all spaces, regardless of whether I am a collaborator or not.  This is not expected, only those that I participate in should be shown.

src/app/home/work-item-widget/work-item-widget.component.html
dlabrecq: I'm fine with whatever folks decide to do here, just want to confirm with UXD that this is what we want for the UI. 

The other widgets for this dashboard (i.e., "recent spaces" and "Recent Active Pipelines") are showing recent spaces, so this could also be addressed with a label change.

dlabrecq:
Unfortunately, we don't have an API to determine what space the user is a collaborator in. Thus, I believe that's the reason why we chose to show only recent spaces.

catrobson:
Yes, showing recent is better than showing all.  @aptmac is there any chance of changing the API availability to implement this correctly?  If not, I agree with @dlabrecq that recent spaces is the best option we have.

src/app/home/work-item-widget/work-item-widget.component.ts
```diff
@@ -20,15 +20,19 @@ export class WorkItemWidgetComponent implements OnDestroy, OnInit  {
...
       private userService: UserService) {
-    this.subscriptions.push(userService.loggedInUser.subscribe(user => {
-      this.loggedInUser = user;
+    this.loggedInUser = this.userService.currentLoggedInUser;
+    this.subscriptions.push(this.spaceService.getSpacesByUser(this.loggedInUser.attributes.username).subscribe(spaces => {
+      this.spaces = spaces;
+      this.loading = false;
     }));
     this.subscriptions.push(spacesService.recent.subscribe(spaces => {
       this.recentSpaces = spaces;
```
dlabrecq: If we end up showing all user spaces, then we don't need to retrieve recent spaces. The fetchRecentSpace() method (not shown here) should be removed as well.

aptmac:
@catrobson Regarding the API I'm not sure at the moment, but that would be nice functionality to have. However, I still don't think recent spaces is good enough, or even better than showing all of the user's spaces. If showing all the spaces that belong to the user, we manage to display a subset of the spaces that the user is a collaborator on and/or owns. If showing recent, we can end up displaying sometimes zero spaces (like in #[387](https://github.com/openshiftio/openshift.io/issues/387)), or spaces the user doesn't even collaborate on.

To illustrate, I went and viewed someone else's space - one that I'm not a collaborator on, and have no affiliation with.

not-my-space

However, it's a recent space for me now, and now shows up in my recent spaces widget and as only 1 of 2 options in the dropdown for my work items widget.

my-recent-spaces

catrobson:
@aptmac I agree recent spaces makes no practical sense, and we'd like to fix this with the right API as soon as possible.  

However, is "all space" just spaces that I own, or ALL spaces in the system?  The problem is that I'm expecting it is all spaces in the system, which will a) overload the dropdown beyond belief (hundreds, thousands of list items?), and b) be completely unusable in pretty much every situation to find a space I care about.  At least recent spaces gives me a fighting chance to have a small list of relevant/selectable options.

Am I missing something?

aptmac:
@catrobson Ah that'd be it, I probably wasn't clear enough on that; the function from the service for retrieving spaces is `getSpacesByUser`, so it would be retrieving all of the spaces that the current user owns.

catrobson:
@aptmac Ok, that makes more sense to me then.  I'm ok with only showing the spaces I own as a more solid, expected, list until we get the right API in place.  Any chance you could also create an issue for the correct API for this so we don't lose that need?

aptmac:
@catrobson That sounds good to me, I'll create an issue for that.

src/app/home/work-item-widget/work-item-widget.component.ts
aptmac: @dlabrecq hm, I didn't realize there wasn't an easy way of retrieving all work-items owned by a particular user; I see the concern regarding many API calls.

What are your thoughts on treating this like the Profile work-items component, where the combobox is populated based on the results of `userService.getSpacesByUser` but the auto-population of work items is performed using `spacesService.recent`? This way there can be fewer API calls upfront to populate the widget with recent work items, but leave the functionality and freedom for the user to afterwards select any of their spaces to view work items?

joshuawilson:
What needs to be done to finish this?

aptmac:
@joshuawilson I asked for some follow up feedback [0] before addressing the review comments; once sorted out I will make the required adjustments

[0] https://github.com/fabric8-ui/fabric8-ui/pull/2966#discussion_r195541476

src/app/home/work-item-widget/work-item-widget.component.ts
aptmac: ^ ping @dlabrecq

dlabrecq: At one time, the dashboard widget was labeled "Recent Work Items'. It allows me to view work items I recently created in the openshift.io space, for example. The profile page doesn't allow me to do that because I can only see my spaces.

If we only show the user's spaces, I cannot view my work items from the openshift.io space. The workaround we discussed with UXD was to show recent spaces, so users can potentially reach all their work items. That's what we have implemented today.

Changing the dashboard widget to show only the user's spaces is a question for @catrobson. The code for the two widgets is almost identical otherwise -- they both use ``spacesService.recent``.

Ideally, we would have an API that provides all of the user's work items, across all system spaces. I want to view ALL my work items on the dashboard as well as the profile page.

aptmac: @dlabrecq Ah that's a good point, I had forgotten about the case where you may have a WI in another user's space. In which case, there is an open issue regarding an API as you have described (https://github.com/openshiftio/openshift.io/issues/3760), but the collaborators API is being deprecated so any functionality would be put into the new API.

In which case, sorry to be going around in loops, but it appears that your original critique may be a valid substitute for the time being:
> The other widgets for this dashboard (i.e., "recent spaces" and "Recent Active Pipelines") are showing recent spaces, so this could also be addressed with a label change.

joshuawilson:
@aptmac are you still working on this? Please rebase.

jiekang:
I can pick up the review after a rebase...

aptmac:
The initial two reported issues [0][1] that this PR aimed to fix display intentional behaviour, that being that the work items component on the home page displays 'recent' information as it's neighbouring components (recent spaces & recent workspaces). Additionally, through discussion with UX it was expressed that the widget could be reworked to display all work items belonging to a user, but this currently lacks API implementation [2] which would make this action potentially costly. Until a time the widget can be tweaked to display this behaviour, it is working as intended and this PR is obsolete.

[0] https://github.com/openshiftio/openshift.io/issues/3504
[1] https://github.com/openshiftio/openshift.io/issues/3621
[2] https://github.com/openshiftio/openshift.io/issues/3760

PR #2968, opened 2018-05-30
fix(work-items): fix profile work-items combobox width and display all spaces

aptmac:
This short PR addresses [issue 1886](https://github.com/openshiftio/openshift.io/issues/1886), where if the user has a long space name then the combobox will overlap the title of the work item component in the profile view.

To fix, the width of the combobox has been made to be `width: 45%` just like the work items widget on the [home page](https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/home/work-item-widget/work-item-widget.component.less#L7) [1]. Additionally, the number of spaces being retrieved to display in the combobox was hardcoded at 10 spaces, so I've removed the limitation so all spaces should show up in the combobox.

Before:
profile-before-1

profile-before-2

After:
profile-after

[0] https://github.com/openshiftio/openshift.io/issues/1886
[1] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/home/work-item-widget/work-item-widget.component.less#L7

### DIFF
--- a/src/app/profile/overview/work-items/work-items.component.ts
+++ b/src/app/profile/overview/work-items/work-items.component.ts
@@ -15,8 +15,7 @@ import { filterOutClosedItems } from '../../../shared/workitem-utils';
   encapsulation: ViewEncapsulation.None,
   selector: 'alm-work-items',
   templateUrl: './work-items.component.html',
-  styleUrls: ['./work-items.component.less'],
-  providers: [SpaceService]
+  styleUrls: ['./work-items.component.less']
 })
 export class WorkItemsComponent implements OnDestroy, OnInit  {
   context: Context;
--- a/src/app/profile/overview/work-items/work-items.component.ts
+++ b/src/app/profile/overview/work-items/work-items.component.ts
@@ -41,14 +41,14 @@ export class WorkItemsComponent implements OnDestroy, OnInit  {
     this.viewingOwnAccount = this.contextService.viewingOwnContext();
     this.subscriptions.push(contexts.current.subscribe(val => this.context = val));
     if (this.context.user.attributes) {
-      this.subscriptions.push(spaceService.getSpacesByUser(this.context.user.attributes.username, 10).subscribe(spaces => {
+      this.subscriptions.push(spaceService.getSpacesByUser(this.context.user.attributes.username).subscribe(spaces => {
         this.spaces = spaces;
       }));
     }
     this.subscriptions.push(this.broadcaster.on('contextChanged').subscribe(val => {
       this.context = val as Context;
       if (this.context.user.attributes) {
-        this.subscriptions.push(spaceService.getSpacesByUser(this.context.user.attributes.username, 10).subscribe(spaces => {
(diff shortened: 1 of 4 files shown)


alien-ike:
### Ike Plugins (test-keeper)

src/app/profile/overview/work-items/work-items.component.ts
```diff
@@ -37,7 +37,7 @@ export class WorkItemsComponent implements OnDestroy, OnInit  {
     this.subscriptions.push(userService.loggedInUser.subscribe(user => {
       this.loggedInUser = user;
       if (user.attributes) {
-        this.subscriptions.push(spaceService.getSpacesByUser(user.attributes.username, 10).subscribe(spaces => {
+        this.subscriptions.push(spaceService.getSpacesByUser(user.attributes.username).subscribe(spaces => {
```
dlabrecq: This limitation was intentional. I don't recall the exact reason, but there could be quite a lot of spaces shown here. We use recent spaces on the home dashboard for this very reason.

Perhaps @mindreeper2420 or @catrobson can best answer this question?

mindreeper2420: The limit of 10 spaces was set so that there would be no scrolling within the card. On smaller screens, it has the possibility of introducing double scrollbars (scroll page content and scroll card content), so the cards are fixed at a height to display that desired number of items.

joshuawilson:
Please rebase.

src/app/profile/overview/work-items/work-items.component.ts
aptmac: @mindreeper2420 Hm yeah I guess that makes sense, but isn't it dependant on the system as well? For example, I logged in on my phone and the combobox is nicely scrollable; it'd easily be able to accommodate all my spaces. 

android

Or would there be a way to fix the height of the combobox and make it scrollable instead while still allowing for all spaces to be shown? I feel like otherwise we're not delivering the information we're advertising.

aptmac:
@mindreeper2420 I made the combobox cap out at showing 10 items, but scrollable too so it can include all items (based on an example I found on StackOverflow [2])

scrollable

@dlabrecq Yeah I don't think the `get space()` is being used, but I think this should be done in a follow up commit. There is a bit of refactoring to do I think .. after the new profile PR [3] there is quite a bit of duplication with regards to subscribing to the spaceService which could use some cleaning up IMO. Also I've been hitting a null id error occasionally (on the home WI widget too) so I'll open an issue about that once I have a better idea of what's going on.

[2] https://stackoverflow.com/a/27635760
[3] https://github.com/fabric8-ui/fabric8-ui/commit/0f8a284b99d14244cd3e38dfcdc2d1765aa6e437#diff-dbcba9469d0ef1b03d69e14bb0d46b9cR44

joshuawilson:
what needs to be done to merge this?

aptmac:
@joshuawilson this needs a review - I also have a spec file in the works that would be nice to go in with this too

PR #2972, opened 2018-05-30
fix(codebases): fix opening workspaces in a new tab

aptmac:
This PR addresses [issue 3565](https://github.com/openshiftio/openshift.io/issues/3565) [0], in which attempting to open a Che workspace for a codebase with ad-blocker enabled will not work, and an error notification gets displayed to the screen.

After a bit of digging, it seems that in order for a new page to appear with ad-block enabled is to have the window.open functionality called immediately after the component is clicked. Previously, our call to open a new window is nested inside of a subscribe is doesn't execute in time. This causes ad-block to block the new window and the `codebases-item-workspaces` to catch an error. In this PR, the `openWorkspace()` only performs the action of opening a new window based on the eclipse che url that is set when a combobox item is selected. Because of the sync nature, creating and opening a workspace in one function is not feasible, and has been separated.

Before (notice that it's being caught by ad-block):
before

After (here I am opening two different workspaces associated to the same codebase):
after

[0] https://github.com/openshiftio/openshift.io/issues/3565

### DIFF
--- a/src/app/space/create/codebases/codebases-item-workspaces/codebases-item-workspaces.component.ts
+++ b/src/app/space/create/codebases/codebases-item-workspaces/codebases-item-workspaces.component.ts
@@ -136,6 +136,7 @@ export class CodebasesItemWorkspacesComponent implements OnDestroy, OnInit {
           }
         });
       } else {
+        workspaceWindow.close();
         this.workspaceBusy = false;
         // display error message
         this.notifications.message({
--- a/src/app/space/create/codebases/codebases-item-workspaces/codebases-item-workspaces.component.ts
+++ b/src/app/space/create/codebases/codebases-item-workspaces/codebases-item-workspaces.component.ts
@@ -48,7 +48,6 @@ export class CodebasesItemWorkspacesComponent implements OnDestroy, OnInit {
   }
 
   ngOnInit(): void {
-
     this.workspaceBusy = false;
     this.workspacesAvailable = false;
 
@@ -68,9 +67,9 @@ export class CodebasesItemWorkspacesComponent implements OnDestroy, OnInit {
   // Actions
 
   /**
-   * Create workspace and open in editor
+   * Create workspace
    */
-  createAndOpenWorkspace(): void {
+  createWorkspace(): void {
     this.workspaceBusy = true;
     this.subscriptions.push(this.cheService.getState().switchMap(che => {
       if (!che.clusterFull) {
@@ -81,7 +80,6 @@ export class CodebasesItemWorkspacesComponent implements OnDestroy, OnInit {
             this.workspaceBusy = false;
             if (workspaceLinks != undefined) {
               let name = this.getWorkspaceName(workspaceLinks.links.open);
-              this.windowService.open(workspaceLinks.links.open, name);
               this.notifications.message({
                 message: `Workspace created!`,
                 type: NotificationType.SUCCESS
@@ -123,16 +121,16 @@ export class CodebasesItemWorkspacesComponent implements OnDestroy, OnInit {
   }
 
   /**
-   * Open workspace in editor
+   * Opens Eclipse Che workspace in a new tab
    */
   openWorkspace(): void {
+    let workspaceWindow = this.windowService.open('about:blank', '_blank');
     this.workspaceBusy = true;
     this.subscriptions.push(this.cheService.getState().switchMap(che => {
       if (!che.clusterFull) {
         // create
         return this.workspacesService.openWorkspace(this.workspaceUrl).map(workspaceLinks => {
           this.workspaceBusy = false;
-          let workspaceWindow = this.windowService.open('about:blank', '_blank');
           if (workspaceLinks != undefined) {
             workspaceWindow.location.href = workspaceLinks.links.open;
           }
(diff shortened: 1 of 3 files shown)


src/app/space/create/codebases/codebases-item-workspaces/codebases-item-workspaces.component.ts
```diff
@@ -123,39 +122,44 @@ export class CodebasesItemWorkspacesComponent implements OnDestroy, OnInit {
...
+  }
+
+  /**
+   * Set the URL to the Eclipse Che workspace
+   */
+  setEclipseCheUrl(): void  {
+    if (this.workspaceUrl !== 'default') {
+      this.workspaceBusy = true;
+      this.subscriptions.push(this.cheService.getState().switchMap(che => {
+        return this.workspacesService.openWorkspace(this.workspaceUrl).map(workspaceLinks => {
```
corinnekrych: @aptmac where do you check whether `che.clusterFull` ?

aptmac: `che.clusterFull` is still checked when the workspace is created [0]; this `setEclipseCheUrl` is fired when the user selects an existing workspace in the combobox so the Url can be preloaded and not have to be resolved as an Observable when launching the workspace in a new tab.

[0] https://github.com/aptmac/fabric8-ui/blob/02e6c8447d0911b85d6ce06f9229ac426b18f04e/src/app/space/create/codebases/codebases-item-workspaces/codebases-item-workspaces.component.ts#L76

fabric8cd:
@aptmac fabric8/fabric8-ui:SNAPSHOT-PR-2972-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2972-fabric8-ui.badger.fabric8.io

joshuawilson:
@ibuziuk can you please review.

src/app/space/create/codebases/codebases-item-workspaces/codebases-item-workspaces.component.ts
ibuziuk: @corinnekrych  `clusterFull` should be checked in both places - creating new workspace & opening existing

corinnekrych: @aptmac let's add it back when opening then....

joshuawilson:
@ibuziuk can we merge this?

src/app/space/create/codebases/codebases-item-workspaces/codebases-item-workspaces.component.ts
aptmac: Sure, I'll work on adding that check back, thanks for clarifying.

aptmac:
@corinnekrych I've made some changes to the PR and brought back the code to check the che cluster. I wanted to get some feedback on this approach before continuing forward, because the `recent-workspaces-widget` and `codebases-item-actions` will require similar tweaks.

Having the windowService opening a new tab from within the cheService subscription was causing a delay that resulted in the pop-up blocker blocking the new tab from opening. However, if the windowService opens as a direct consequence of the button click (i.e., at the beginning of the function) and opening a blank screen, we can later supply the url of the workspace when it has been resolved [0]. This is how the `recent-workspaces-widget` currently operates [1] but there is no check to see if the cluster is full; so it will need to be tweaked.

[0] https://github.com/fabric8-ui/fabric8-ui/pull/2972/files#diff-46deac7af9920bed67be95ee1e799c54R127
[1] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/dashboard-widgets/recent-workspaces-widget/recent-workspaces-widget.component.ts#L90

corinnekrych:
👀

catrobson:
@corinnekrych I'm not sure I have a full idea of the scenarios we can run into here.  Can you verify below?

1.  User can come to page and a moment or two passes before workspace starts loading <-- momentary loading state with some sort of status needed
2. User can come to page and workspace loads perceptibly immediately to the user <-- nothing needed
3. User can come to page and workspace will never load because they are out of space to run it (cluster full)? <-- page telling the user what happened needed

corinnekrych:
@catrobson Correct! The scenario we found disturbing is number 3. Leaving a blank page without information does not seem right.

catrobson:
@corinnekrych - great.  Should we turn this around ASAP to get this PR finalized, or can we create a new UX story to create a screen for this and add it incrementally after this PR has been merged?

aptmac:
Hm, perhaps it's not a completely urgent matter, but would be nice to have a design of what to display if the cluster is full.

@catrobson @corinnekrych the ability to close a window had completely slipped my mind at the time, and now inside of the `else` block that is triggered when the cluster is full I added a line to close the opened workspace. If the cluster is full, the action of trying to open a workspace will look like this:
clusterfull

What are your thoughts on this? There is a toast notification that tells the user that the cluster is out of capacity.

catrobson:
@aptmac it's a little jarring that a window closed by itself.  I think its a fine way to handle it for now and we can give you a full screen design for the window that gets opened to replace this with as soon as we can get to it.

joshuawilson:
@ibuziuk is there an API to check if the cluster if full?

aptmac:
@catrobson Okay that sounds reasonable, thanks.

@joshuawilson fwiw the `clusterFull` is coming from the che service [0][1]

[0] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/space/create/codebases/services/che.service.ts#L34
[1] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/space/create/codebases/services/che.ts

joshuawilson:
@corinnekrych is there something else that needs to be done on this PR?

PR #2974, opened 2018-05-31
Remove webpack custom module resolution

christianvogt:
related to openshiftio/openshift.io#3686

- This change depends on all OSIO projects to use peer dependencies for angular in order to allow `npm link --production` to function correctly.
- Removed custom module resolution from webpack common. We're going back to default node_module hierarchy resolution. This will allow components to work with their own dependencies and versions properly.
- Since I removed the custom module resolution look up paths, I had to update all modules in src/ to use relative imports when referring to other modules within src/. Using absolute paths is wrong to use in this case anyways.
- Updated ng2-restangular to ngx-restangular because it was causing module not found issues. The version of ngx-restangular was taken from the docs for use with angular 4.
- Had to set `symlinks` to `false` in webpack dev config to allow `npm link` to function correctly with webpack module loading.

To test I linked locally the following projects:

- fabric8-planner
- ngx-widgets
- ngx-forge / ngx-launcher
- ngx-login-client
- ngx-fabric8-wit
- fabric8-recommender
- ngx-base
- fabric8-analytics-dependency-editor

With linking all these projects I was able to load a dev environment. I was also able to test watching the planner build and have the page reload when changes were made.

`at-loader` was throwing some errors (build errors, not javascript runtime errors) after webpack successfully compiled. This occurred only when linking all these projects so I needed to update the tsconfig with paths to angular and rxjs. Even with these errors everything was working as far as I could tell. There was only a single angular and rxjs install so I don't know why these errors occurred in the first place.


joshuawilson:
Won't this need all the versions of the other UI repos updated too? Can this be merged and work without having those updated?

christianvogt:
@joshuawilson Oops forgot to mention we need the rest of the dependencies fixed before merging this one.

fabric8cd:
@christianvogt fabric8/fabric8-ui:SNAPSHOT-PR-2974-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2974-fabric8-ui.badger.fabric8.io

christianvogt:
Upversioned dependencies to align with those that have their peer dependencies updated.

Removed webpack and tsconfig workarounds introduced by PR https://github.com/fabric8-ui/fabric8-ui/pull/2962 for `fabric8-analytics-dependency-editor`.

Added an alias for `testing` to the test webpack config to allow unit tests to still reference the `testing` dir as an absolute reference.

fabric8cd:
@christianvogt fabric8/fabric8-ui:SNAPSHOT-PR-2974-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2974-fabric8-ui.badger.fabric8.io

joshuawilson:
please resolve merge conflict

config/webpack.common.js
```diff
@@ -100,37 +92,13 @@ module.exports = function (options) {
      */
     resolve: {
 
-      alias: {
```
joshuawilson: why was this removed?

package.json
```diff
@@ -137,14 +137,14 @@
     "mydatepicker": "2.1.0",
     "ng2-completer": "1.6.3",
     "ng2-dragula": "1.5.0",
-    "ng2-restangular": "0.1.30",
     "ng2-truncate": "1.3.11",
     "ngx-base": "2.3.1",
```
joshuawilson: don't you need to update all the libs that you made changes to?

config/webpack.common.js
christianvogt: The alias was removed because now webpack loads nested node_module folders allowing libs to have their own dependencies that fabric8-ui doesn't need to include directly itself. This is one of the major reasons for the change. The alias was put in as a work around until this PR gets in.

package.json
christianvogt: Yup. I didn't realize that all those recent upversion PRs were been deleted instead of merged.

christianvogt:
@joshuawilson should be all up to date now.

PR #2975, opened 2018-05-31
Applications-widget: in-progress message for newly created apps

dlabrecq:
Show in-progress message for newly created applications. Also refactored the loading widget to accommodate additional messages and links.

Fixes: https://github.com/openshiftio/openshift.io/issues/3671

With no applications staged/running:
screen shot 2018-05-29 at 3 11 34 pm

With staged application:
screen shot 2018-05-31 at 10 43 34 am

### DIFF
--- a/src/app/dashboard-widgets/applications-widget/applications-widget.component.html
+++ b/src/app/dashboard-widgets/applications-widget/applications-widget.component.html
@@ -1,13 +1,20 @@
 <f8-feature-toggle featureName="ApplicationsCard">
   <div class="applications-widget card-pf f8-card" user-level>
     <div class="card-pf-heading f8-card-heading">
+      <div class="card-pf-heading-details"
+           *ngIf="applicationsInProgress > 0 && (stageConfigsAvailable || runConfigsAvailable)">
+        <p>
+          Setting up {{applicationsInProgress}} applications.
+          <a [routerLink]="['/', loggedInUser?.attributes?.username, this.currentSpace?.attributes?.name, 'create', 'pipelines']">View progress.</a>
+        </p>
+      </div>
       <h2 class="card-pf-title">
         Applications
       </h2>
     </div>
     <div class="card-pf-body f8-card-body">
-      <ng-container *ngIf="buildConfigsAvailable; then showApplications else showBlankState"></ng-container>
-      <ng-template #showBlankState>
+      <ng-container *ngIf="buildConfigsAvailable; then showApplications else showEmptyState"></ng-container>
+      <ng-template #showEmptyState>
         <fabric8-loading-widget message="Please wait while we load your applications"
                                 *ngIf="loading === true"></fabric8-loading-widget>
         <div class="f8-blank-slate-card" *ngIf="loading !== true">
@@ -23,24 +30,34 @@
         </div>
       </ng-template>
       <ng-template #showApplications>
-        <div id="spacehome-applications-list" class="f8-card-applications-list">
-          <div class="col-md-12 col-lg-6 padding-none">
-            <div class="f8-card-applications-list-stage">
-              <h4 class="f8-card-applications-list-step">Stage</h4>
-              <div class="f8-card-applications-list-name">
-                <fabric8-applications-list [buildConfigs]="stageBuildConfigs"></fabric8-applications-list>
+        <ng-container *ngIf="stageConfigsAvailable || runConfigsAvailable; else showInProgress">
+          <div id="spacehome-applications-list" class="f8-card-applications-list">
+            <div class="col-md-12 col-lg-6 padding-none">
+              <div class="f8-card-applications-list-stage">
+                <h4 class="f8-card-applications-list-step">Stage</h4>
+                <div class="f8-card-applications-list-name">
+                  <fabric8-applications-list [buildConfigs]="stageBuildConfigs"></fabric8-applications-list>
+                </div>
+              </div>
+            </div>
+            <div class="col-md-12 col-lg-6 padding-none">
+              <div class="f8-card-applications-list-run">
+                <h4 class="f8-card-applications-list-step">Run</h4>
+                <div class="f8-card-applications-list-name">
+                  <fabric8-applications-list [buildConfigs]="runBuildConfigs"></fabric8-applications-list>
+                </div>
               </div>
             </div>
           </div>
-        </div>
-        <div class="col-md-12 col-lg-6 padding-none">
-          <div class="f8-card-applications-list-run">
-            <h4 class="f8-card-applications-list-step">Run</h4>
-            <div class="f8-card-applications-list-name">
-              <fabric8-applications-list [buildConfigs]="runBuildConfigs"></fabric8-applications-list>
-            </div>
-          </div>
-        </div>
+        </ng-container>
+        <ng-template #showInProgress>
+          <fabric8-loading-widget
+              [title]="'Application Setup In Progress'"
+              [message]="'This may take a few minutes. Once setup is complete, you\'ll see your deployed applications here.'">
+            <a id="test-applications-pipelines-link"
+               [routerLink]="['/', loggedInUser?.attributes?.username, this.currentSpace?.attributes?.name, 'create', 'pipelines']">View setup progress</a>
+          </fabric8-loading-widget>
+        </ng-template>
       </ng-template>
     </div>
   </div>
--- a/src/app/dashboard-widgets/applications-widget/applications-widget.component.html
+++ b/src/app/dashboard-widgets/applications-widget/applications-widget.component.html
@@ -4,7 +4,7 @@
       <div class="card-pf-heading-details"
            *ngIf="applicationsInProgress > 0 && (stageConfigsAvailable || runConfigsAvailable)">
         <p>
-          Setting up {{applicationsInProgress}} applications.
+          Setting up {{applicationsInProgress}} application{{applicationsInProgress > 1 ? 's' : ''}}.
           <a [routerLink]="['/', loggedInUser?.attributes?.username, this.currentSpace?.attributes?.name, 'create', 'pipelines']">View progress.</a>
         </p>
       </div>
@@ -24,7 +24,7 @@
           </p>
           <div class="f8-blank-slate-main-action">
             <div *ngIf="userOwnsSpace">
-              <button id="spacehome-applications-add-button" class="btn btn-primary btn-lg" (click)="addToSpace.emit()">Create an Application</button>
+              <button id="test-applications-add-button" class="btn btn-primary btn-lg" (click)="addToSpace.emit()">Create an Application</button>
             </div>
           </div>
         </div>
--- a/src/app/dashboard-widgets/applications-widget/applications-widget.component.html
+++ b/src/app/dashboard-widgets/applications-widget/applications-widget.component.html
@@ -24,7 +24,7 @@
           </p>
           <div class="f8-blank-slate-main-action">
             <div *ngIf="userOwnsSpace">
-              <button id="test-applications-add-button" class="btn btn-primary btn-lg" (click)="addToSpace.emit()">Create an Application</button>
+              <button id="spacehome-applications-add-button" class="btn btn-primary btn-lg" (click)="addToSpace.emit()">Create an Application</button>
             </div>
           </div>
         </div>
@@ -54,7 +54,7 @@
           <fabric8-loading-widget
               [title]="'Application Setup In Progress'"
               [message]="'This may take a few minutes. Once setup is complete, you\'ll see your deployed applications here.'">
-            <a id="test-applications-pipelines-link"
+            <a id="spacehome-applications-pipelines-link"
                [routerLink]="['/', loggedInUser?.attributes?.username, this.currentSpace?.attributes?.name, 'create', 'pipelines']">View setup progress</a>
           </fabric8-loading-widget>
         </ng-template>
--- a/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts
+++ b/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts
@@ -211,7 +211,7 @@ describe('ApplicationsWidgetComponent', () => {
     providers: [
       { provide: ActivatedRoute, useValue: jasmine.createSpy('ActivatedRoute') },
       { provide: Contexts, useValue: ({ current: ctxSubj }) },
-      { provide: LocationStrategy, useValue: jasmine.createSpyObj('LocationStrategy', ['prepareExternalUrl']); },
+      { provide: LocationStrategy, useValue: jasmine.createSpyObj('LocationStrategy', ['prepareExternalUrl']) },
       { provide: PipelinesService, useFactory: () => pipelinesService },
       {
         provide: FeatureTogglesService, useFactory: () => {
--- a/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts
+++ b/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts
@@ -295,11 +295,11 @@ describe('ApplicationsWidgetComponent', () => {
       this.testedDirective.runBuildConfigs.length = 0;
       this.testedDirective.stageBuildConfigs.length = 0;
       this.detectChanges();
-      expect(this.fixture.debugElement.query(By.css('#test-applications-pipelines-link'))).toBeNull();
+      expect(this.fixture.debugElement.query(By.css('#test-applications-pipelines-link'))).not.toBeNull();
     });
 
-    it('Stage or run build configs to show empty state', function(this: TestingContext) {
-      expect(this.fixture.debugElement.query(By.css('#test-applications-pipelines-link'))).not.toBeNull();
+    it('Stage or run build configs not to show empty state', function(this: TestingContext) {
+      expect(this.fixture.debugElement.query(By.css('#test-applications-pipelines-link'))).toBeNull();
     });
   });
 
--- a/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts
+++ b/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts
@@ -4,13 +4,16 @@ import {
   Input
 } from '@angular/core';
 import { By } from '@angular/platform-browser';
+import { ActivatedRoute, Router, RouterModule } from '@angular/router';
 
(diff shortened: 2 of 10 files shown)


jiekang:
Can you elaborate on the reason a timeout is required?

dlabrecq:
In the case there are no recent spaces (e.g., I've reset my environment), the widget continues to show it's loading because its boolean state is never changed. The timeout ensures that the empty state can be displayed in that scenario. I assumed that recent pipelines subscription had similar behavior; however, I've removed that one.

dlabrecq:
Other than the pipelines, I'm not receiving an empty response for a clean environment. Only after creating a space does it work as expected.

I can remove that code if this is going to derail the PR? Then, someone can deal with that issue another time?

src/app/dashboard-widgets/applications-widget/applications-widget.component.html
```diff
@@ -18,24 +25,34 @@ <h3>This space has no applications</h3>
...
             </div>
           </div>
-        </div>
+        </ng-container>
+        <ng-template #showEmptyState>
+          <fabric8-loading-widget
+              [title]="'Application Setup In Progress'"
+              [message]="'This may take a few minutes. Once setup is complete, you\'ll see your deployed applications here.'"
+              [hyperlinkText]="'View setup progress'"
+              (onHyperlinkClick)="navigateToPipelines($event)"
```
joshuawilson: ` navigateToPipelines` does NOT take any arguments

jiekang:
Okay. That sounds like the API behavior and loading behavior has some mismatch. If the API only emits when there is content, but it's possible for it to never emit then the widget should have an empty state and as far as I can tell, loading is not needed.

This is separate from the addition of an in-progress flow though. Thanks for opening https://github.com/openshiftio/openshift.io/issues/3705

src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts
```diff
@@ -201,7 +221,28 @@ describe('ApplicationsWidgetComponent', () => {
...
+        }
+      },
+      {
+        provide: UserService, useFactory: () => {
+          let userService = createMock(UserService);
+          userService.getUser.and.returnValue(fakeUser);
+          userService.loggedInUser = fakeUser.publish() as ConnectableObservable<User> & jasmine.Spy;
+          return userService;
+        }
+      },
```
jiekang: Trailing comma; I think the tslint rule is not to have these but I could be wrong.

jiekang:
```
$ npm run lint

> fabric8-ui@0.0.0-development lint /fabric8-ui
> find src -name '*.ts' | xargs tslint -p . -c tslint.json --force


ERROR: fabric8-ui/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts[10, 9]: missing whitespace
ERROR: fabric8-ui/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts[10, 39]: missing whitespace
ERROR: fabric8-ui/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts[245, 8]: Unnecessary trailing comma
ERROR: fabric8-ui/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts[250, 53]: Spaces before function parens are disallowed
ERROR: fabric8-ui/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts[254, 47]: Spaces before function parens are disallowed
ERROR: fabric8-ui/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts[260, 53]: Spaces before function parens are disallowed
ERROR: fabric8-ui/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts[266, 51]: Spaces before function parens are disallowed
ERROR: fabric8-ui/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts[272, 52]: Spaces before function parens are disallowed
ERROR: fabric8-ui/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts[276, 50]: Spaces before function parens are disallowed
ERROR: fabric8-ui/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts[283, 68]: Spaces before function parens are disallowed
ERROR: fabric8-ui/src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts[291, 77]: Spaces before function parens are disallowed
ERROR: fabric8-ui/src/app/feature-flag/resolver/feature-flag.resolver.spec.ts[3, 9]: missing whitespace
ERROR: fabric8-ui/src/app/feature-flag/resolver/feature-flag.resolver.spec.ts[3, 43]: missing whitespace
ERROR: fabric8-ui/src/app/shared/owner-guard.service.spec.ts[5, 1]: Import sources within a group must be alphabetized.
```

While you're in this area, can you fix the tslint issues? The issues in other areas (feature-flags, etc.) can be fixed in another PR (not necessarily written by yourself)

src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts
```diff
@@ -241,17 +282,15 @@ describe('ApplicationsWidgetComponent', () => {
...
     });
 
     it('should disable buttons if the user does not own the space', function (this: TestingContext) {
       this.hostComponent.userOwnsSpace = false;
-      this.testedDirective.runBuildConfigs.length = 0;
-      this.testedDirective.stageBuildConfigs.length = 0;
+      this.testedDirective.buildConfigs.length = 0;
       this.detectChanges();
 
       expect(this.fixture.debugElement.query(By.css('#spacehome-applications-add-button'))).toBeNull();
```
jiekang: The behavior of this component has been altered and I'm surprised there are no test additions or modifications to existing tests. Can tests be added?

src/app/dashboard-widgets/applications-widget/applications-widget.component.ts
```diff
@@ -125,4 +158,10 @@ export class ApplicationsWidgetComponent implements OnDestroy, OnInit {
       this.runBuildConfigs = orderBy(this.runBuildConfigs, ['name']);
     }
   }
+
+  // Navigate to pipelines page
+  private navigateToPipelines(): void {
+    this.router.navigate(['/', this.loggedInUser.attributes.username, this.currentSpace.attributes.name,
```
jiekang: The `loggedInUser` is set via an asynchronous method which happens to involve http calls. Though it would be quite abnormal, there is nothing stopping that http call to be delayed quite a while. And though the path towards calling this function may have a check, it's not immediately obvious that this is safe. This logic applies to `currentSpace` as well.

Best to be safe and null/undefined check these with an appropriate response (maybe notify user something went wrong).

src/app/dashboard-widgets/loading-widget/loading-widget.component.html
```diff
@@ -1,4 +1,6 @@
 <div class="f8-loading f8-blank-slate-card">
   <div class="spinner spinner-lg"></div>
-  <p>{{message}}</p>
+  <h3 *ngIf="title !== undefined">{{title}}</h3>
```
jiekang: These checks are curious. Does it make sense to also check for null?

src/app/dashboard-widgets/applications-widget/applications-widget.component.ts
```diff
@@ -51,8 +68,24 @@ export class ApplicationsWidgetComponent implements OnDestroy, OnInit {
   ngOnInit(): void {
   }
 
+  get applicationsInProgress(): number {
```
jiekang: The logic here should be tweaked, e.g.:

As far as I can tell, the UX path is the extremely similar as the Pipelines page, just with different graphics. Maybe the underlying structure of the Pipelines UX can be used as inspiration.

build-fail

jiekang: While we're here, UX is definitely an evolving thing. Maybe the applications widget should be altered a little. I think it should always show the list of applications that exist in a space, even if they're in the process of building. The information definitely exists and having this widget empty or even having this loading screen that doesn't actually tell you what applications there are (and links to places that do tell you) is just strange.

src/app/dashboard-widgets/applications-widget/applications-widget.component.html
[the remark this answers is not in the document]
dlabrecq: I'll provide the link via ng-content instead.

src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts
dlabrecq: Will do

src/app/dashboard-widgets/loading-widget/loading-widget.component.html
dlabrecq: Just omitting the HTML if the title has not been provided.

src/app/dashboard-widgets/applications-widget/applications-widget.component.ts
dlabrecq: I'm going to refactor this to use a routerLink. However, do you have any examples you can point me to in fabric8-ui? 

I see similar navigation in many places. For example, in the header, profile, app overlay, app launcher, etc. In some cases, context?.user?.attributes?.username is used, but none of that provides an error response.

src/app/dashboard-widgets/applications-widget/applications-widget.component.ts
dlabrecq: I think your concerns are valid, but probably shouldn't be discussed here as the topic is beyond the scope of this PR. It may be best to discuss with the UX designers by creating an issue for the uxd-team to evaluate?

src/app/dashboard-widgets/applications-widget/applications-widget.component.spec.ts
dlabrecq: Fixed them all

src/app/dashboard-widgets/applications-widget/applications-widget.component.ts
jiekang: I think the second comment is out of scope but the first comment, that this new 'Applications in Progress' shows this state even when the builds are 'not' in progress but have failed, is relevant to this PR. I think this needs to be addressed.

jiekang:
Could you reword the commits to be about the changes to the code? The commit message `updated per code review comments` is not helpful at a glance. I recommend multiple commits, each one covering a divisible portion of work. This can be squashed when merged, and the messages will be retained automatically in the squashed commit.

joshuawilson:
Please rebase.

src/app/dashboard-widgets/applications-widget/applications-widget.component.html
```diff
@@ -1,41 +1,58 @@
 <f8-feature-toggle featureName="ApplicationsCard">
   <div class="applications-widget card-pf f8-card" user-level>
     <div class="card-pf-heading f8-card-heading">
+      <div class="card-pf-heading-details"
+           *ngIf="applicationsInProgress > 0 && (stageConfigsAvailable || runConfigsAvailable)">
+        <p>
+          Setting up {{applicationsInProgress}} applications.
```
jiekang: State needs to be checked here for `1` to be `application` instead of `applications`. E.g. `1 application`, not `1 applications`.

src/app/dashboard-widgets/applications-widget/applications-widget.component.html
```diff
@@ -1,41 +1,58 @@
...
+          <a [routerLink]="['/', loggedInUser?.attributes?.username, this.currentSpace?.attributes?.name, 'create', 'pipelines']">View progress.</a>
```
jiekang: Maybe this should be checked with UXD, but I think the period should be dropped.

`View progress` instead of `View progress.` This follows the other widgets and linked text in the UI.

src/app/dashboard-widgets/applications-widget/applications-widget.component.ts
dlabrecq: UXD will need to redesign this to account for the failure state. I suggest adding the failure state in another PR. Or closed this one and wait for a redesign? 

I've added your comments and reopened the UXD issue below:
https://github.com/openshiftio/openshift.io/issues/3249#event-1661303518

jiekang:
Is it expected for the 'Promote' button to appear in both sections, and also after the build has been promoted?

double-promote

I can open an issue to track it if it is indeed an issue. Of cours it's not in the scope of this PR. Please let me know.

src/app/dashboard-widgets/applications-widget/applications-widget.component.ts
jiekang: I would prefer this PR to address the changes coming out of https://github.com/openshiftio/openshift.io/issues/3249 

This failure state is a common occurrence; is there a time constraint for this addition? I think it could wait to be more robust.

src/app/dashboard-widgets/applications-widget/applications-widget.component.html
dlabrecq: Considering the first sentence "Setting up {{applicationsInProgress}} applications." uses a period, then the next sentence should have one as well.

jiekang:
Okay. So there is probably a problem with the pipelines. I requested an update for the previously opened issue https://github.com/openshiftio/openshift.io/issues/3455

For this widget, is it expected for the Promote button appear in the 'Run' section? That seems strange to me.

dlabrecq:
The fabric8-applications-list component renders a generic list of pipelines. It does not necessarily need to account for stage or run environments. 

If the backend API provides the correct data, the promote button would not be displayed under run. In fact, the application should not appear under run at all.

That said, we could add logic specifically for a run environment. However, I believe the underlying issue with the pipelines API should still be corrected. 

The application should have never been under the run environment if it hasn't been promoted, right?

jiekang:
Okay thanks for the explanation. Definitely agree the underlying API is fishy. 

To clarify how I got to the screenshot, I imported an existing codebase with the full pipeline and hit Promote via the widget. It looks to have also triggered a full rebuild of the pipeline as well as the promotion, which then eventually leads to the widget displaying the screenshot.

src/app/dashboard-widgets/applications-widget/applications-widget.component.html
dlabrecq: I'm not going to argue over a period -- reviews shouldn't be painful.

Feel free to obtain clarification from UXD.

src/app/dashboard-widgets/applications-widget/applications-widget.component.ts
dlabrecq: I suspect this won't be addressed until the following sprint. If so, I'll close the PR.

jiekang: I invite third opinions on this decision; maybe @joshuawilson

fabric8cd:
@dlabrecq fabric8/fabric8-ui:SNAPSHOT-PR-2975-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2975-fabric8-ui.badger.fabric8.io

src/app/dashboard-widgets/applications-widget/applications-widget.component.html
```diff
@@ -17,30 +24,40 @@ <h3>This space has no applications</h3>
           </p>
           <div class="f8-blank-slate-main-action">
             <div *ngIf="userOwnsSpace">
-              <button id="spacehome-applications-add-button" class="btn btn-primary btn-lg" (click)="addToSpace.emit()">Create an Application</button>
+              <button id="test-applications-add-button" class="btn btn-primary btn-lg" (click)="addToSpace.emit()">Create an Application</button>
```
joshuawilson: Why was this ID changed and not the one on line 34?

dlabrecq: I just made the IDs consistent for the unit tests (i.e., all prefixed with "test"), so we know where/how they are being used.

joshuawilson: When you change the IDs you need to check with @ldimaggi because they might be used in the e2e. So don't remove it.

dlabrecq: Ok, I'll just revert the ID.

fabric8cd:
@dlabrecq fabric8/fabric8-ui:SNAPSHOT-PR-2975-8 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2975-fabric8-ui.badger.fabric8.io

src/app/dashboard-widgets/applications-widget/applications-widget.component.ts
joshuawilson: I'm fine with an iterative approach to development. @jiekang do you see something that should block releasing this or can the other tasks be in a different PR?

joshuawilson: @dlabrecq can you write up the task[s] for the comments listed above and point UXD toward them?

fabric8cd:
@dlabrecq fabric8/fabric8-ui:SNAPSHOT-PR-2975-9 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2975-fabric8-ui.badger.fabric8.io

src/app/dashboard-widgets/applications-widget/applications-widget.component.ts
jiekang: After this change, if a user has a single pipeline that is in a failed state, the widget will show the  'Application Setup in Progress' screen.

However, before this change, in the same scenario, I believe the widget would remain empty; is that right? @dlabrecq

If before this change, in the same scenario, the widget is able to show the application and it's failed state, then this would be blocking as a regression. Otherwise, I don't see it as blocking as it's a broken state that is replaced by a different broken state that makes improvements elsewhere.

dlabrecq: The application widget never showed a failed state, unless the application was already staged and being promoted, NOT in-progress. The change is that we show an in-progress state, for applications not staged, instead of an empty state.

dlabrecq: @joshuawilson I've already re-opened the task for UXD to include failure states in their design. The team is looking at it...

https://github.com/openshiftio/openshift.io/issues/3249

PR #2978, opened 2018-06-01
fix(profile-update): form validations fixed.

rohitkrai03:
Fixes https://github.com/openshiftio/openshift.io/issues/3708
Fixes https://github.com/openshiftio/openshift.io/issues/3701
Fixes https://github.com/fabric8-services/fabric8-auth/issues/508
Fixes https://github.com/fabric8-services/fabric8-auth/issues/509

Fixes a few validations in profile update page.
- Profile Bio cannot be removed or set to empty string.
- Company name cannot be removed.
- URL cannot be removed.
- URL validation is not working correctly. For example i can update URL to http:// and it gets updated.
- Also, removed imageUrlValidation since we don't take user input for that in new design.

### DIFF
--- a/src/app/profile/update/update.component.spec.ts
+++ b/src/app/profile/update/update.component.spec.ts
@@ -358,26 +358,9 @@ describe('UpdateComponent', () => {
     });
   });
 
-  describe('#validateImageUrl', () => {
-    // Wrongfully fails
-    xit('should verify https://fabric8.io/ to be an invalid image url', () => {
-      let url: string = 'https://fabric8.io/';
-      component.imageUrl = url;
-      component.validateImageUrl();
-      expect(component.imageUrlInvalid).toBeTruthy();
-    });
-
-    it('should verify http://design.jboss.org/fabric8/logo/final/fabric8_logo_600px.png to be a valid image url', () => {
-      let url: string = 'http://design.jboss.org/fabric8/logo/final/fabric8_logo_600px.png';
-      component.imageUrl = url;
-      component.validateImageUrl();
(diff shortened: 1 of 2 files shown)


src/app/profile/update/update.component.spec.ts
```diff
@@ -358,26 +358,9 @@ describe('UpdateComponent', () => {
...
-      component.imageUrl = url;
-      component.validateImageUrl();
-      expect(component.imageUrlInvalid).toBeFalsy();
-    });
-  });
-
   describe('#validateUrl', () => {
     it('should verify not-a-real-url to be an invalid url', () => {
-      let url: string = 'not-a-real-url';
+      let url: string = 'htt:/not-a-real-url.some-thing.';
```
jiekang: Could you add an array of strings containing various invalid urls and then run the inner 'test' over all of them?

jiekang:
For the point 'URL validation is not working correctly. For example i can update URL to http:// and it gets updated.', I'm a little confused. In the regex tester it looks like this would not match http:// so it should return the correct value. Are you saying that even though this validation function works, it still lets you update the URL? Or this validation function doesn't work for this case?

rohitkrai03:
@jiekang It doesn't let you update the URL to `http://`. The description just gives a summary of all the validation problems this PR fixes. That may be the reason for the confusion. I've also added this use case in the test.

fabric8cd:
@rohitkrai03 fabric8/fabric8-ui:SNAPSHOT-PR-2978-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-2978-fabric8-ui.badger.fabric8.io

PR #2982, opened 2018-06-04
fix(deployments): adjust layout of placeholders when loading for smal…

aptmac:
…l screen sizes

This PR addresses [issue 3735](https://github.com/openshiftio/openshift.io/issues/3735) [0], in which the loading page for the deployments page has a distorted layout when viewing with a small screen size. 

Here are some images of the before & after of this PR:

Before:
before

After:
after

[0] https://github.com/openshiftio/openshift.io/issues/3735


andrewazores:
Looks okay to me. @mindreeper2420 any input?

PR #2983, opened 2018-06-04
refactor(deployments): clean up DeploymentsService DI

andrewazores:
Move providers for DeploymentsService parameters (timer, window width) adjacent to provider for the
service itself. Also add provider/dependency for service poll rate. Avoid hardcoded value in
service's calculation for window width.


andrewazores:
@joshuawilson this one's ready to go, looks like you merged master and then didn't get around to merging the PR?

PR #2991, opened 2018-06-06
fix(ngrx): added entity library for data normalization

sudsen:
This is a library that we need for the planner, specifically on ngrx/store.


joshuawilson:
[test]

joshuawilson:
@sudsen this needs the package-lock update too

sudsen:
@joshuawilson added package-lock changes

PR #3001, opened 2018-06-07
fix(http): short term cache for http GET requests

christianvogt:
I've implemented a cache at the fabric8-ui HTTP service level for all GET requests. This cache helps eliminate multiple requests for the same resource that occur within a short period of time. This can happen when separate components on a page request the same data.

The cache key is based on the request URL, query params, and headers.

This is not an ideal implementation with respect to observables because the expectation of an observable returned from the HTTP service is that new subscriptions result in new network requests. Please see my comment in the code and share any suggestions on how to implement it.

cc: @joshuawilson @andrewazores @jiekang @corinnekrych


alien-ike:
### Ike Plugins (test-keeper)

src/app/shared/fabric8-ui-http.service.ts
```diff
@@ -14,9 +15,43 @@ import { AUTH_API_URL, HttpService , SSO_API_URL } from 'ngx-login-client';
...
+  } else {
+    key = req.url;
+    if (req.headers) {
+      key += `_${JSON.stringify(req.headers.toJSON())}`;
+    }
+  }
+  return key;
+}
+
+const TTL = 200; // cache requests for N milliseconds
```
dgutride: This seems like a reasonable timeout

christianvogt: @dgutride have a look at the long comment in the code. Any thoughts on an expiring cache observable?

src/app/shared/fabric8-ui-http.service.ts
```diff
@@ -14,9 +15,43 @@ import { AUTH_API_URL, HttpService , SSO_API_URL } from 'ngx-login-client';
...
+}
+
+// Creates a cache key from headers, query params, and URL
+function createCacheKey(req: string | Request, options?: RequestOptionsArgs): string {
+  let key;
+  if (typeof req === 'string') {
+    key = req;
+    if (options) {
+      if (options.params) {
+        key += `_${options.params.toString()}`;
```
andrewazores: What would you think of introducing a hashing utility for cases like this where we need to construct cache keys? We do something similar in the DeploymentsService (https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/space/create/deployments/services/deployments.service.ts#L320), and although there's nothing particularly *wrong* about constructing the key in this way, there's also no particular reason for the key to be so potentially large or to directly encode state. Anyway, just a thought - I'm fine with this function as-is.

src/app/shared/fabric8-ui-http.service.ts
```diff
@@ -14,9 +15,43 @@ import { AUTH_API_URL, HttpService , SSO_API_URL } from 'ngx-login-client';
...
+    }
+  }
+  return key;
+}
+
+const TTL = 200; // cache requests for N milliseconds
 @Injectable()
 export class Fabric8UIHttpService extends Http {
 
+  private pendingRequests: PendingRequests = {};
```
andrewazores: Can this be a `Map<string, Observable<Response>>`?

src/app/shared/fabric8-ui-http.service.ts
```diff
@@ -29,6 +64,39 @@ export class Fabric8UIHttpService extends Http {
...
+
+    if (isGetMethod(url, options)) {
+
+      const cacheKey = createCacheKey(url, options);
+      if (cacheKey in this.pendingRequests) {
+        // return the cached result
+        return this.pendingRequests[cacheKey];
+      }
+
+      const result = this.makeRequest(urlStr, url, options).do(() => {
```
andrewazores: If the request produces an error, I think this timeout and cache clearing won't happen. Is this intentional? Maybe this should be `finally` rather than `do`?

src/app/shared/fabric8-ui-http.service.ts
```diff
@@ -29,6 +64,39 @@ export class Fabric8UIHttpService extends Http {
...
+        setTimeout(() => {
+          // clear the cache
+          delete this.pendingRequests[cacheKey];
+        }, TTL);
+      })
+      // This caching strategy will always return the cached response.
+      // A new subscriber to the observable will never cause a new network request to be made.
+      // Ideally an observable should be created in such a way that new subscribers to the
+      // observable will cause a new network request to be made if the subscription occurs
+      // after the cached value has expired. But how to do this with rxjs?
```
andrewazores: So the concern is that there may be some elapsed time between when a consumer creates this request and when they actually subscribe to it, resulting in the cached Observable they've received becoming stale in the interim?

If the cache holds ReplaySubjects instead of direct Http cold Observables, those subjects can be constructed with a buffer size of 1 and a window time of TTL. We then wouldn't need the `publishReplay(1)` operator either. This would allow consumers to receive a cached value only if it's still fresh enough. To allow a new request to be made in case the value has gone stale while the consumer held onto the reference to the subject before subscribing, maybe `timeoutWith()` could help us? `subject.timeoutWith(50, this.makeRequest(urlStr, url, options)` would allow the ReplaySubject 50ms to emit a cached value, and if none, fall back to making a new network request. It's not ideal to have to use a timeout to detect when the ReplaySubject's cache has expired, but I don't think there's a better (public) API for that. Although that's not to say there is no better approach to take.

So it would look like:

```
this.pendingRequests[cacheKey] = result;
return result;
```

changes to roughly

```
const subject: ReplaySubject<Response> = new ReplaySubject<Response>(1, TTL).timeoutWith(50, this.makeRequest(urlStr, url, options));
this.pendingRequests[cacheKey] = subject;
this.makeRequest(urlStr, url, options).subscribe(v => subject.next(v), e => subject.error(e), () => subject.complete());
return subject;
```

jiekang: I was thinking of something like... 
Pseudocode:
```
request(url, options?): Observable<Response> {
  return  Observable.of(getCacheItem(url, options))
      .map(cacheItem => {
        if (!cacheItem.isCached)
          return super.request(url, options).do(result => setCache(result))
        }
        return Observable.of(cacheItem.response);
      });
}

getCacheItem(url, options?) : CacheItem {
  if response exists and is not expired, return cacheitem with true and response
  else, remove cached response and return cacheitem with false and null response
}

```

andrewazores: In order to avoid the problem of a consumer holding on to the Observable for long enough that it goes stale before it's subscribed upon, wouldn't `getCacheItem` need to hold some extra state so that the `is not expired` check is relative to the current time of the subscription vs the time that the Observable was returned and the TTL? I think that's where the complexity is getting hidden.

Also... I don't think you can lock on the store...

jiekang: The store holds both the response and the expiry time of the response together so `getCacheItem` check is just `time.Now > cache.expiryTime`.

andrewazores: Hmm. Will `getCacheItem` be invoked at call time or at subscribe time? If it's invoked at subscribe time then it covers the case I'm talking about.

jiekang: It's at subscribe time since the function is returning `Observable.of(getCacheItem(...)...`

andrewazores: I'd like to see a proof of concept for that - if it was `Observable.of(() => getCacheItem())` I would believe that the invocation is delayed until subscribe time, but as it is, I don't see how that call isn't executed immediately on request.

jiekang: I thought `Observable.of(func)` was shorthand for `Observable.of(()=>func)`. My bad, the intention was that.

andrewazores: Okay. That might work. The `.map` needs to be changed as well then though, because the `Observable.of` now has a different type - it's an Observable of functions returning CacheItem, not just an `Observable<CacheItem>`. Now we can delay invocation of `getCacheItem` until subscription time, and if the cache maintains an expiry date for its items, I think this makes some kind of sense.

jiekang: Ah. Looks like there used to be `Observable.start` for this but it doesn't seem to be around anymore. Best would be `Observable.create(observer => observer.next(func()))` I tested in a small unit test:
```
    function getItem(): string {
      console.log('executing');
      return 'hello';
    }

    fit('temp', () => {
      let a = Observable.create((obs) => obs.next(getItem()));
      console.log('pre-subscribe');
      a.subscribe(val => console.log(val));
    });
```

and output was:
```
LOG: 'pre-subscribe'
LOG: 'executing'
LOG: 'hello'
```

jiekang: And you were right, `Observable.of(func())` evaluates the function on the spot. Nice catch!

andrewazores: Yea, I was looking at `create` too. Should it also `obs.complete()`?

jiekang: Good point. I think it should complete. Also the `map` after should probably be a `switchMap`.

jiekang: So result is:

Pseudocode:
```
request(url, options?): Observable<Response> {
  return  Observable
    .create((obs) => {
      obs.next(getCacheItem(url, options));
      obs.complete();
    })
    .switchMap((cacheItem) => {
        if (!cacheItem.isCached)
          return super.request(url, options).do(response => setCache(response, url, options))
        }
        return Observable.of(cacheItem.response);
      });
}

getCacheItem(url, options?) : CacheItem {
  if response exists && is not expired
    return cacheitem with true and response
  else
    remove cached response and return cacheitem with false and null response
}

setCache(response, url, options?) {
  add response to cache with ms expiry time of time.now + expiryTimeInMs
}

```

jiekang: Hmm on second thought, if that inner obs completes, does that affect multiple subscribers? Maybe it should actually remain open... More testing required, hahah...

jiekang: A closer test suggests it's okay :)
```
    let i: number = 1;

    function getItem(): string {
      console.log('executing');
      if (i === 1) {
        i = 2;
        return 'hello';
      }
      i = 1;
      return 'blah';
    }

    fit('temp', () => {
      let a: Observable<string> = Observable
        .create((obs) => {
          obs.next(getItem());
          obs.complete();
        })
        .switchMap((val) => {
           if (val === 'hello') {
             return Observable.of('Bye!');
           } else {
             return Observable.of('No');
           }
        });
      console.log('pre-subscribe');
      a.subscribe(val => console.log(val));
      a.subscribe(val => console.log(val));
      a.subscribe(val => console.log(val));
      a.subscribe(val => console.log(val));
    });
```
outputs
```
LOG: 'pre-subscribe'
LOG: 'executing'
LOG: 'Bye!'
LOG: 'executing'
LOG: 'No'
LOG: 'executing'
LOG: 'Bye!'
LOG: 'executing'
LOG: 'No'
```

christianvogt:
Thanks @jiekang and @andrewazores for all the comments. I'm going to read your suggestions more carefully and try them out.

fabric8cd:
@christianvogt fabric8/fabric8-ui:SNAPSHOT-PR-3001-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-3001-fabric8-ui.badger.fabric8.io

christianvogt:
@jiekang Could you have another look now. I tried your solution but it had one fault. If multiple requests came in before the HTTP response arrived, the 2nd request would still result in another HTTP request. This is because the response was only stored in the cache after receiving the HTTP response.

fabric8cd:
@christianvogt fabric8/fabric8-ui:SNAPSHOT-PR-3001-3 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-3001-fabric8-ui.badger.fabric8.io

andrewazores:
@christianvogt ah, right... did you try the ReplaySubject technique I posted as well? It may still have the short latency added on requests where the cached value has expired between subject acquisition and subscription, but those cases are probably relatively few and it's a relatively short latency as well. 50ms was just an example but I suspect that a ReplaySubject subscription might be fulfilled synchronously if there is a valid cached value (and if not then subscribing with a specific synchronously scheduler should force this), so the `timeoutWith` argument could be 1 or maybe even 0ms. This way of sharing the ReplaySubject and firing off a single request only when there was no subject corresponding to a cache key previously should avoid any issues with multiple requests being fired when there are multiple concurrent subscribers.

jiekang:
Yeah that's exactly the scenario I was describing with "To fully handle the concurrency there needs to be a lock on the store, but I believe the only cost of dropping the locks is the existence of the case where multiple subscribers 'at the same time' will see an empty cache and still make individual http requests."

I think the one way to solve this is to have a blocking mechanism of some sort. Another may be to rework the observable being returned to be cached, so multiple subscribers are subscribing to the same observable.

@andrewazores any thoughts?

andrewazores:
@jiekang I'm not aware of any way to add synchronization locks in JS/TS and explicit blocking doesn't sound great without a proper lock await/notify mechanism. Caching the returned Observable itself and using that to multicast the eventual response is exactly what I'm talking about :) I think that's going to be the way forward.

jiekang:
Nice, we partially have the same answer :)

src/app/shared/fabric8-ui-http.service.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+      httpService = service;
+      mockService = mock;
+    }
+  ));
+
+  afterEach(() => {
+    Date.now = oriDateNow;
+  });
+
+  fit('cache should expire', () => {
```
andrewazores: `fit`s need to be changed back to `it`s before push

src/app/shared/fabric8-ui-http.service.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+    });
+
+    // POST request should nto cache
+    const request: Request = new Request({url: 'test1', method: 'POST'});
+    httpService.request(request).subscribe(() => {});
+    httpService.request(request).subscribe(() => {});
+    expect(httpRequestCount).toBe(2);
+  });
+
+  fit('cache key should include params and headeres', () => {
```
andrewazores: "headeres"

andrewazores:
Yea, I think just switching out the AsyncSubject with a ReplaySubject as I described inside of the `Observable.create` block will have some nice characteristics. It should multicast properly from my part, and should have a better cache eviction strategy due to the CacheItem's explicitly recorded TTL timestamp from your part.

andrewazores:
Actually - to avoid the problem of multiple concurrent subscribers appearing while a request/response is in flight, I think the `timeoutWith` needs to be longer. If it's "immediate" as I was just describing then we'll run into the same problem, where no already-available cached value will result in new requests being fired even if there's already one in progress. Maybe it should be `timeoutWith(TTL)` or `timeoutWith(TTL/2)` or something like that?

christianvogt:
@andrewazores if an HTTP call is currently in flight, subsequent requests will get the `AsyncSubject` which will then return the cached value to all those subscribers when the HTTP call completes. What am I missing?

fabric8cd:
@christianvogt fabric8/fabric8-ui:SNAPSHOT-PR-3001-4 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-3001-fabric8-ui.badger.fabric8.io

andrewazores:
@christianvogt oh right, I don't think I'd read the latest updates carefully enough. I'll take a closer look later today but this seems to make sense.

christianvogt:
I've added another unit tests to test that multiple subscriptions wait on the same async subject.

fabric8cd:
@christianvogt fabric8/fabric8-ui:SNAPSHOT-PR-3001-5 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-3001-fabric8-ui.badger.fabric8.io

jiekang:
So on manual user testing, I think this works quite well, possibly a bit too well. There are situations where the same component is making the same request multiple times, or more generically, making a set of requests that are inefficient and could be easily improved. This hides that behavior too; I wonder if an easy toggle could be introduced for a developer to switch when running locally. Even just having to comment a single line of code would be acceptable to me :)

andrewazores:
@jiekang `if (false && isGetRequest())` and an `xdescribe` would do the trick I guess?

jiekang:
Oh good point; I will have a branch for that now. Thanks!

andrewazores:
Yea, I'm good with it too.

PR #3007, opened 2018-06-08
fix(analyze-overview): correctly hide create app button

chrislessard:
This patch resolves an issue in which the "create an application" button is mistakenly hidden when users visit their own space. It also introduces a lifecycle hook to ensure that angular re-evaluates whether the button should be shown when navigating immediately from one space to another.

Solves: https://github.com/openshiftio/openshift.io/issues/3702

### DIFF
--- a/src/app/space/analyze/analyze-overview/analyze-overview.component.html
+++ b/src/app/space/analyze/analyze-overview/analyze-overview.component.html
@@ -2,10 +2,10 @@
   <div id="analyze-overview-dashboard" class="container-fluid analyze-overview-wrapper" user-level>
     <div class="row f8-dashboard-masthead">
       <div class="col-sm-10">
-        <fabric8-edit-space-description-widget [userOwnsSpace]='userOwnsSpace()'></fabric8-edit-space-description-widget>
+        <fabric8-edit-space-description-widget [userOwnsSpace]='userOwnsSpace'></fabric8-edit-space-description-widget>
       </div>
       <div class="col-sm-2">
-        <div *ngIf="userOwnsSpace()">
+        <div *ngIf="userOwnsSpace">
           <button *ngIf="this.authentication.isLoggedIn()" id="user-level-analyze-overview-dashboard-create-space-button" class="btn btn-primary btn-lg f8-dashboard-masthead-create-application pull-right"
             (click)="showAddAppOverlay()"> Create an Application </button>
         </div>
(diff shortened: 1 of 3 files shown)


src/app/space/analyze/analyze-overview/analyze-overview.component.ts
```diff
@@ -43,6 +44,13 @@ export class AnalyzeOverviewComponent implements OnInit, OnDestroy {
         this._myWorkItemsCard = true;
       }
     }));
+
+    this.userOwnsSpaceBool = this.userOwnsSpace();
+  }
+
+  ngDoCheck() {
+    console.log('logged');
```
aptmac: Did you mean to leave a console.log in here?

src/app/space/analyze/analyze-overview/analyze-overview.component.ts
```diff
@@ -43,6 +44,13 @@ export class AnalyzeOverviewComponent implements OnInit, OnDestroy {
         this._myWorkItemsCard = true;
       }
     }));
+
+    this.userOwnsSpaceBool = this.userOwnsSpace();
```
aptmac: I'm wondering if it'd be cleaner to call the function here to set the `userOwnsSpace` boolean, and create a get method for retrieving the value (similar to how the `myWorkItemsCard` boolean is used). 

Something like:
`get userOwnsSpace: boolean () { return this._userOwnsSpace; }`

src/app/space/analyze/analyze-overview/analyze-overview.component.spec.ts
```diff
@@ -156,4 +155,52 @@ describe('AnalyzeOverviewComponent', () => {
 
     expect(this.testedDirective.userOwnsSpace()).toBe(false);
   });
+
+  it('should hide the Create an Application button if the user owns the space', function(this: TestingContext) {
```
aptmac: this should be `it('should not hide...)`?

src/app/space/analyze/analyze-overview/analyze-overview.component.html
```diff
@@ -5,7 +5,7 @@
         <fabric8-edit-space-description-widget [userOwnsSpace]='userOwnsSpace()'></fabric8-edit-space-description-widget>
       </div>
       <div class="col-sm-2">
-        <div *ngIf="userOwnsSpace()">
+        <div *ngIf="userOwnsSpaceBool">
```
aptmac: See comments below about using a `get` method for the `userOwnsSpace` boolean, and then this could read something like: `*ngIf="userOwnsSpace"` etc.

chrislessard:
@aptmac Good catches, I've forced pushed changes that follow up on your comments. Let me know what you think !

fabric8cd:
@chrislessard fabric8/fabric8-ui:SNAPSHOT-PR-3007-2 fabric8-ui is deployed and available for testing at https://fabric8-ui-fabric8-ui-pr-3007-fabric8-ui.badger.fabric8.io

chrislessard:
@aptmac done !

PR #3026, opened 2018-06-13
fix(pipelines): do not emit error if 401 recieved when not logged in

jiekang:
This addresses the error notification that appears if you are not logged in and visit a space. It will not appear if you are not logged in. This makes an assumption that 'logged in' means there is an auth token in local storage (looking at how authentication service actually works). I think this is reasonable. If the auth token is invalid, it will still notify, which is nice.

This fixes https://github.com/openshiftio/openshift.io/issues/3347

Interestingly the auth token service implementation returns the value in local storage or the empty string so in theory it will never be null. I think there's a few http services that do a null check on the getToken function; they should also probably not add the header if the return value is an empty string. Alternatively, the auth service should return null as the local storage could contain a key with an empty value.

### DIFF
--- a/src/app/space/create/pipelines/services/pipelines.service.spec.ts
+++ b/src/app/space/create/pipelines/services/pipelines.service.spec.ts
@@ -51,18 +51,15 @@ describe('Pipelines Service', () => {
   let mockNotificationsService: jasmine.SpyObj<NotificationsService>;
   let svc: ActualPipelinesService;
 
-  beforeEach(() => {
-    const mockAuthService: jasmine.SpyObj<AuthenticationService> = createMock(AuthenticationService);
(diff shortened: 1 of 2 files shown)


alien-ike:
### Ike Plugins (test-keeper)

jiekang:
Oh I should add tests. Thanks test-keeper :)

jiekang:
I have added tests, for what it's worth.

andrewazores:
The verb can be changed from "fix" to "test" on the second commit.

Seems fine otherwise. There are probably other places where we would need to make a similar change to avoid the same bug report symptoms from different views. Not sure at this point how we would go about generalizing this solution - I guess we would need to look to the application-wide notification system for that.

PR #3034, opened 2018-06-15
fix(add-collaborators): replaced angular-2-dropdown-multiselect component with ng-select

rohitkrai03:
Fixes https://github.com/openshiftio/openshift.io/issues/3122
Fixes https://github.com/openshiftio/openshift.io/issues/3622
Fixes https://github.com/fabric8-services/fabric8-auth/issues/510

Previously used `angular-2-dropdown-component` was not suitable for customizing options and labels in the dropdown. Also had a very limited sets of features which will be used in future when components for Teams, Spaces and Permissions will be written.
This PR -
- Replaces the old multiselect component with `ng-select` https://github.com/ng-select/ng-select
- Used 1.x version as angular 6 has breaking changes which is supported in 2.x version of ng-select.
- Adds user avatar and username to search result in dropdown.
- Adds a typeahead feature for user search with debounce.
- Adds default placeholder options showing start typing, nothing found and fetching data states.


src/app/space/settings/collaborators/add-collaborators-dialog/add-collaborators-dialog.component.ts
```diff
@@ -22,66 +22,48 @@ export class AddCollaboratorsDialogComponent implements OnInit {
...
-      buttonClasses: 'btn btn-default',
-      selectionLimit: 0,
-      closeOnSelect: false,
-      showCheckAll: false,
-      showUncheckAll: false,
-      dynamicTitleMaxItems: 3,
-      maxHeight: '300px'
-    };
+    this.searchTerm.pipe(
+      tap(() => console.log('Something')),
```
christianvogt: remove `console.log`

src/app/space/settings/collaborators/add-collaborators-dialog/add-collaborators-dialog.component.ts
```diff
@@ -22,66 +22,48 @@ export class AddCollaboratorsDialogComponent implements OnInit {
...
+      distinctUntilChanged(),
+      debounceTime(200),
+      switchMap(term => this.userService.getUsersBySearchString(term))
+    ).subscribe(users => {
+      this.collaborators = users;
+      this.loading = false;
```
christianvogt: `this.loading` is never set to `true`

package.json
```diff
@@ -112,12 +112,12 @@
     "@angular/upgrade": "4.4.6",
     "@angularclass/hmr": "2.1.3",
     "@angularclass/hmr-loader": "3.0.4",
+    "@ng-select/ng-select": "^1.5.2",
```
christianvogt: remove `^`

src/app/space/settings/collaborators/add-collaborators-dialog/add-collaborators-dialog.component.less
```diff
@@ -1,4 +1,5 @@
 @import (reference) '../../../../../assets/stylesheets/shared/osio.less';
+@import '../../../../../../node_modules/@ng-select/ng-select/themes/default.theme.css';
```
christianvogt: Does less duplicate the rules if another less file imports the same css?

src/app/space/settings/collaborators/add-collaborators-dialog/add-collaborators-dialog.component.ts
rohitkrai03: this.loading was supposed to be set true in the tap() but by mistake I left console.log statement after testing something there. Updated.

src/app/space/settings/collaborators/add-collaborators-dialog/add-collaborators-dialog.component.less
rohitkrai03: Webpack should remove duplicate rules if another file imports the same css i think.

src/app/space/settings/collaborators/add-collaborators-dialog/add-collaborators-dialog.component.ts
christianvogt: I think the order of your pipe should be `debounceTime` => `distinctUntilChanged` => `tap`.
* `debounceTime` first to ensure efficient event processing by waiting 200ms before doing any work.
* `distinctUntilChanged` only after debounced so that comparisons to previous value is minimized.
* `tap` only after we know we're about to search. So that we aren't showing `loading` while the user simply types. Otherwise loading can go on forever if they return back to their previous value due to `distinctUntilChanged`.

```
debounceTime(200),
distinctUntilChanged(),
tap(() => this.loading = true),
```

src/app/space/settings/collaborators/add-collaborators-dialog/add-collaborators-dialog.component.less
christianvogt: I feel like that's not the case because CSS specificity is dependent on order of inclusion.

Not a big concern right now since this is the only import. I'd be interested in confirming though. Because if it would be duplicated, then this is not a good place to include the theme. We'd be better off creating a wrapper component for ng-select that includes the css.

My first thought still is that the rules would be duplicated.

rohitkrai03: Since this is the only component where ng-select is being used right now I think it's ok. We'll probably use ng-select more in future and then we can move it to a common CSS file or wrap it in a custom component as you said.

joshuawilson:
please rebase, you need to clean up package-lock

rohitkrai03:
@joshuawilson Done. Please check.

PR #3037, opened 2018-06-15
fix(header): correct gradient and underline in navbars

chrislessard:
This small patch corrects the gradient direction and removes a white underline from the selected link of the principle navbar when viewing a space.

Preview: 
2018-06-15-123447_1920x1080_scrot

### DIFF
--- a/src/app/layout/header/header.component.less
+++ b/src/app/layout/header/header.component.less
@@ -71,13 +71,11 @@
   .navbar-primary {
     > .active > a {
       background-color: @color-pf-black-900;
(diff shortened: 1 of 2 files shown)


joshuawilson:
What do we do about the nav in Planner?
@mindreeper2420

mindreeper2420:
@chrislessard can you fix the solid black background that is happening behind 'Analyze/Plan/Create'?

mindreeper2420:
@joshuawilson nothing - it looks the same as it does in the Create/Deployments screenshot, which is fine.

chrislessard:
@mindreeper2420 done !

PR #3043, opened 2018-06-18
patternfly-ng: remove deprecations

dlabrecq:
Removes deprecated and unused patternfly-ng module imports.

Resolves https://github.com/openshiftio/openshift.io/issues/2704

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "resolved": "https://registry.npmjs.org/@angular/animations/-/animations-4.4.6.tgz",
       "integrity": "sha1-+mYYmaik44y3xYPHpcl85l1ZKjU=",
       "requires": {
-        "tslib": "^1.7.1"
+        "tslib": "1.9.2"
       }
     },
     "@angular/common": {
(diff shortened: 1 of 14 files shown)


jiekang:
The change to the package-lock.json file is massive relative to the rest of this PR. Could you clarify where it's coming from? Just taking a look it seems to have a lot more exact versions instead of caret ranges than which sounds good for stability to me.

dlabrecq:
That's what `npm install` generated (npm 5.3.0/node 8.3.0). I assumed we should be updating the package-lock when dependencies are updated. Should I remove it?

jiekang:
I like the changes there but I don't know when package-lock.json updates are supposed to be made. If someone else could chime in I'd appreciate it. @joshuawilson Is it okay?

andrewazores:
Same, actual changes look fine but I don't know how we want to handle the package lock file.

joshuawilson:
The package-lock should (must) be updated when the package is changed. 

Please rebase.

dlabrecq:
Rebased with latest changes

PR #3053, opened 2018-06-20
fix(stylelint): prevent webpack watching from stopping when stylelint…

aptmac:
… errors occur

At the moment, when the webpack stylelint plugin picks up an error from a `.less` file it stops the build. However, when the error is resolved it does not resume building due to `failOnError: true` [[0]](https://github.com/fabric8-ui/fabric8-ui/blob/master/config/webpack.common.js#L469) in our webpack common config. When this occurs I usually have to re-run `npm start`, which is potentially tedious when testing or changing numerous `.less` values. FWIW, the stylelint plugin ships with this value set to `false` by default [[1]](https://github.com/webpack-contrib/stylelint-webpack-plugin#failonerror). I thought this might be a nice QoL improvement based on a conversation I had with @chrislessard the other day, @mindreeper2420 what are your thoughts on this?

I've prepared a couple of screenshots and gifs below that illustrate the changes.

**Gifs:**
In these gifs, I'm removing a line from a `.less` file and leaving an empty line (which will cause stylelint to flag an error), and then restoring deleted code in hopes that webpack will continue. 

Current (build stops after stylelint error):
before

With this PR (build resumes after stylelint error):
after

**Screenshots of the displayed error messages:**
Current:
before

With this PR:
after

[0] https://github.com/fabric8-ui/fabric8-ui/blob/master/config/webpack.common.js#L469
[1] https://github.com/webpack-contrib/stylelint-webpack-plugin#failonerror


alien-ike:
### Ike Plugins (test-keeper)

chrislessard:
+1 this is amazing work

mindreeper2420:
@aptmac as long as the build still fails on stylelint errors, I'm good with this (as it looks like it does). The change was made to stop the build as stylelint errors were still making it into production, causing issues with renders.

aptmac:
@mindreeper2420 Yeah it fails to compile when there are stylelint errors, but will resume compilation when changes occur.

joshuawilson:
@mindreeper2420 do you approve?

christianvogt:
+1 awesome. This was so annoying

PR #3055, opened 2018-06-20
fix(pipelines): add padding-top to pipelines history page

christianvogt:
fixes https://github.com/openshiftio/openshift.io/issues/3722

Added 36px padding to the pipelines history page to align with what other pages are doing.

IMO this is still an ugly hack. The content area should be sized and positioned correctly so that the content only needs to fill the available space. Right now the content needs to be aware of the size of the navigation. Unfortunately such a change would affect other components which already do the same thing done in this fix.


alien-ike:
### Ike Plugins (test-keeper)

PR #3061, opened 2018-06-21
Deployments service tests

andrewazores:
This PR contains a large number of commits targeting only the file `deployments.service.spec.ts`. As one of (if not the very) oldest spec files in the Deployments area, a lot of cruft had accumulated over time. This patchset cleans it out and brings the code style in line with the rest of the specs in the Deployments area.

There are some minor semantic changes limited to the form of test removals. These tests were written at a time when the `DeploymentsService` itself directly made HTTP calls and the `DeploymentApiService` did not exist, and verified things like proper URL encoding of request URLs. Since the `DeploymentApiService` handles the direct HTTP connections now, these tests are redundant, and were removed after "porting" the tests to using a mock DeploymentApiService instance rather than a "live" one with mocked HTTP backend. Otherwise, test semantics remain unchanged and the changes are simply various refactoring steps.


aptmac:
There is also a nice decrease in the amount of time it takes for the tests to run, see
Before:
before
With this PR:
after

PR #3065, opened 2018-06-21
fix(work-items): fix TypeError regarding undefined values viewing ano…

aptmac:
…ther user's space

This PR addresses [issue 3837](https://github.com/openshiftio/openshift.io/issues/3837) [0], in which an error is thrown regarding an undefined value for `loggedInUser` in the profile work-item component.

Currently, the profile work-items component uses the `loggedInUser` value to retrieve work items and create links to work items, however, this does not make any sense in the context of viewing another user's space. If I'm viewing a different users profile, the component starts loading my work items because I'm logged in, which is what causes this error. Furthermore, the work item related variables aren't reset in the subscription to `broadcaster.on('contextChanged')`, so if I view a subsequent profile page then the original values still linger.

[0] https://github.com/openshiftio/openshift.io/issues/3837

### DIFF
--- a/src/app/profile/overview/work-items/work-items.component.ts
+++ b/src/app/profile/overview/work-items/work-items.component.ts
@@ -22,11 +22,11 @@ export class WorkItemsComponent implements OnDestroy, OnInit  {
   context: Context;
   currentSpace: Space;
   currentSpaceId: string = 'default';
-  loggedInUser: User;
   recentSpaces: Space[] = [];
   recentSpaceIndex: number = 0;
   subscriptions: Subscription[] = [];
   spaces: Space[] = [];
+  user: User;
   workItems: WorkItem[] = [];
   viewingOwnAccount: Boolean;
 
@@ -38,7 +38,6 @@ export class WorkItemsComponent implements OnDestroy, OnInit  {
       private broadcaster: Broadcaster,
       private userService: UserService,
       private contextService: ContextService) {
-    this.viewingOwnAccount = this.contextService.viewingOwnContext();
     this.subscriptions.push(contexts.current.subscribe(val => this.context = val));
     if (this.context.user.attributes) {
       this.subscriptions.push(spaceService.getSpacesByUser(this.context.user.attributes.username, 10).subscribe(spaces => {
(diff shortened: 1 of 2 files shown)


alien-ike:
### Ike Plugins (test-keeper)

src/app/profile/overview/work-items/work-items.component.ts
```diff
@@ -38,7 +38,6 @@ export class WorkItemsComponent implements OnDestroy, OnInit  {
       private broadcaster: Broadcaster,
       private userService: UserService,
       private contextService: ContextService) {
-    this.viewingOwnAccount = this.contextService.viewingOwnContext();
```
andrewazores: `viewingOwnAccount` looks unused now

rohitkrai03:
@aptmac Seems like a duplicate issue - https://github.com/openshiftio/openshift.io/issues/3747 for which i raised a PR - https://github.com/fabric8-ui/fabric8-ui/pull/3036.

aptmac:
@rohitkrai03 ah I see, thanks for referencing that issue (I knew this error was familiar ..). PR #3036 would indeed fix the undefined error issue, and I suppose that is enough to close the issue. However, still using the `loggedInUser` value in the other parts of the component would cause the logged in user to see their own work items in the components of other user's profiles. For example, with #3036 applied if I view another user's profile I still see my own work items:
my work items

Additionally, the work item variables won't get reset after the broadcast of `contextChanged`, so if I view someone's profile then traverse to my own profile page, my work items will not be displayed. I think it'd be better to wait for reply from UX as to what information should be displayed when viewing another user's profile page.

jiekang:
This looks like a duplicate of https://github.com/fabric8-ui/fabric8-ui/pull/3036 but with more changes. Can 3036 be dropped? Please inform @rohitkrai03 if so.

aptmac:
@jiekang I think https://github.com/fabric8-ui/fabric8-ui/pull/3036 could go in to fix the original issue, and this PR could be reworked into something later pending reply from UX. I had tried to fix side-effects as a result of the original issue, but this may have lost the original scope.

The problem is that once the URL is fixed then the logged in user will see their work items in the work-item-components when viewing other profile pages. The wording on the profile pages of 'X's work items' indicates that I should be able to see the work items assigned to them, but instead I'd see my own. But this is behaviour hasn't been confirmed, so it should be okay to merge #3036 in the meantime.

andrewazores:
Using the context user rather than the currently logged in user sounds like it would cause the widget to display the context user's assigned workitems, no? Assuming there isn't some access restriction on the workitems query API.

IMO #3036 should go in as a minimal fix for the breaking navigation problem, and the question of whether the widget should really display the other user's workitems or do something else would be a separate issue with a separate PR.

jiekang:
Yeah it will have the same behavior; switching to context is only to avoid the race condition that occurs when using the loggedInUser. Sounds like everyone is for #3036 going in. Thanks!

aptmac:
The issue was fixed in https://github.com/fabric8-ui/fabric8-ui/pull/3036, I'll await further discussion on the topic of what the work-items component should look like in another user's profile and work towards a new PR

PR #3070, opened 2018-06-22
fix(profile): make user profile tab animations consistent with others

andrewazores:
Make "Work Items" and "Spaces" sub-view nav tab animations and styling consistent with sub-view nav
tabs elsewhere in the application, ex. Settings and Create areas. This also replaces a hand-rolled
tab selection system with routing and a router outlet for better extensibility and simplicity.

fixes https://github.com/openshiftio/openshift.io/issues/3724

profile-tabs


andrewazores:
I should note that replacing the hand-rolled "tab index" system with router links and a router outlet does mean that the URL is updated when switching between the subviews as well. This seems fine to me, but I'm not sure if this is undesirable for some reason. IMO it's the right way to go.

src/app/profile/overview/overview-routing.module.ts
```diff
@@ -2,11 +2,27 @@ import { NgModule }  from '@angular/core';
...
+      {
+        path: '',
+        redirectTo: '_workitems'
+      },
+      {
+        path: '_workitems',
+        component: WorkItemsComponent
+      },
+      {
+        path: '_spaces',
```
aptmac: @andrewazores This path was already defined for the spaces.component [0], so clicking "View all spaces" from the header [1] will now route to the profile page my-spaces widget, I'm not sure if this was your intent or not but it renders the "View all spaces" page unreachable (which I don't think is the biggest deal because it's supposed to be removed anyways [2]) but thought I'd just confirm.

[0] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/profile/profile-routing.module.ts#L21
[1] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/layout/header/header.component.html#L101
[2] https://github.com/openshiftio/openshift.io/issues/3536#issuecomment-390823742

andrewazores: Ah, no, that was not intentional. I intended for that to be a unique path. I'll raise a PR momentarily to fix this path collision so the old "View All Spaces" page is reachable, until UXD gives guidance on what should be done with it instead.

PR #3071, opened 2018-06-22
fix(recent-pipelines-widget): retrieve pipeline namespace from pipeli…

aptmac:
…nesService instead of Contexts

This PR addresses [issue 3841](https://github.com/openshiftio/openshift.io/issues/3841) [0], in which the URL to a recent active pipeline will contain an undefined value - resulting in an invalid URL.

The [line of code](https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/dashboard-widgets/recent-pipelines-widget/recent-pipelines-widget.component.ts#L37) [1] that sets the `contextPath` is not being hit, and as a result the value is undefined when placed in the `routerLink`. This PR instead uses the value from the `config.namespace` retrieved from the `PipelinesService.getRecentPipelines()` to build the URL to the pipeline [2].  

[0] https://github.com/openshiftio/openshift.io/issues/3841
[1] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/dashboard-widgets/recent-pipelines-widget/recent-pipelines-widget.component.ts#L37
[2] https://github.com/aptmac/fabric8-ui/commit/dfd49a59056e58ea97a5aa9be2c7cb61dd094406#diff-0f9f68d891087e9b7384f274b6eb0bd7R59

### DIFF
--- a/src/app/dashboard-widgets/recent-pipelines-widget/recent-pipelines-widget.component.html
+++ b/src/app/dashboard-widgets/recent-pipelines-widget/recent-pipelines-widget.component.html
@@ -56,7 +56,7 @@
           <div class="f8-card__pipeline-column">
             <span class="{{buildconfig.iconStyle}} fa-spin" title="{{buildconfig.statusPhase}}" *ngIf="buildconfig.iconStyle === 'pficon-running'"></span>
             <span class="{{buildconfig.iconStyle}}" title="{{buildconfig.statusPhase}}" *ngIf="buildconfig.iconStyle !== 'pficon-running'"></span>
-            <a id="spacehome-pipelines-title" [routerLink]="[contextPath, buildconfig.labels['space'], 'create', 'pipelines']"
+            <a id="spacehome-pipelines-title" [routerLink]="['/', contextPath, buildconfig.labels['space'], 'create', 'pipelines']"
               class="f8-card__pipeline-column-name">
               {{buildconfig.name}}
             </a>
--- a/src/app/dashboard-widgets/recent-pipelines-widget/recent-pipelines-widget.component.ts
+++ b/src/app/dashboard-widgets/recent-pipelines-widget/recent-pipelines-widget.component.ts
@@ -1,8 +1,7 @@
 import { Component, OnDestroy, OnInit, ViewEncapsulation } from '@angular/core';
-import { Context, Contexts } from 'ngx-fabric8-wit';
-import { Observable, Subscription } from 'rxjs/Rx';
+import { Subscription } from 'rxjs/Rx';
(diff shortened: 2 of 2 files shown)


alien-ike:
### Ike Plugins (test-keeper)

src/app/dashboard-widgets/recent-pipelines-widget/recent-pipelines-widget.component.html
```diff
@@ -56,7 +56,7 @@ <h3>Pipelines reflect integration and deployment stages that code goes through</
           <div class="f8-card__pipeline-column">
             <span class="{{buildconfig.iconStyle}} fa-spin" title="{{buildconfig.statusPhase}}" *ngIf="buildconfig.iconStyle === 'pficon-running'"></span>
             <span class="{{buildconfig.iconStyle}}" title="{{buildconfig.statusPhase}}" *ngIf="buildconfig.iconStyle !== 'pficon-running'"></span>
-            <a id="spacehome-pipelines-title" [routerLink]="[contextPath, buildconfig.labels['space'], 'create', 'pipelines']"
+            <a id="spacehome-pipelines-title" [routerLink]="['/', buildconfig.namespace, buildconfig.labels['space'], 'create', 'pipelines']"
```
jiekang: The namespace isn't always the correct choice. For example, when the user is 'jkang@redhat.com' and the namespace is 'jkang', the link will not work.

andrewazores:
If this is a fix for https://github.com/openshiftio/openshift.io/issues/3841 , please indicate this in the commit message body by adding `fixes https://github.com/openshiftio/openshift.io/issues/3841` in commitizen format.

src/app/dashboard-widgets/recent-pipelines-widget/recent-pipelines-widget.component.html
jiekang: Best case is there's a value we can use from the buildconfig, preferably the full url, otherwise the Contexts service should be able to give us a username.

aptmac: Hmm good call. Since it's a user's own pipelines, would it be acceptable here to subscribe to `UserService.loggedInUser` and use the username in the link here?

jiekang: That sounds like it would work. The Contexts service also provides the current user. I wonder which should be used, hah...

aptmac: Ah that sounds good too.

PR #3077, opened 2018-06-26
fix(my-spaces): introduce case insensitive filtering

aptmac:
…e filtering

fixes https://github.com/openshiftio/openshift.io/issues/3857

This PR tweaks the filtering of spaces in the `my-spaces.component`with regards to combination of filters and case insensitivity. 

The `my-spaces.component` uses `name` as a filter attribute, and based on the PatternFly design explanations [[0]](http://www.patternfly.org/pattern-library/forms-and-controls/filter/#design) should be combined with a logical OR. Instead, patternfly-ng [[1]](http://www.patternfly.org/patternfly-ng/#/filters) seems to use logical AND, but this is an implementation detail and has been changed here. I have spaces numbered 1 through 13, and using the the following filters I would expect the following results:

-  filter '1' = [1, 10, 11, 12, 13]
-  filter '2' = [2, 12]
-  filter '1' with '2' = [1, 2, 10, 11, 12, 13]

However, the current result is [12], because of the way the filters are being aggregated.

Before:
before-1-2

After:
after-1-2

Previously, the case had to match exactly what you were looking for. When trying to find my spaces that contain 'em' in them, I expected to have two results show up but instead here are the results:
Before:
before-em

After:
after-em

[0] http://www.patternfly.org/pattern-library/forms-and-controls/filter/#design
[1] http://www.patternfly.org/patternfly-ng/#/filters

### DIFF
--- a/src/app/profile/my-spaces/my-spaces.component.ts
+++ b/src/app/profile/my-spaces/my-spaces.component.ts
@@ -145,7 +145,8 @@ export class MySpacesComponent implements OnDestroy, OnInit {
   matchesFilter(space: Space, filter: Filter): boolean {
     let match = true;
     if (filter.field.id === 'name') {
-      match = space.attributes.name.match(filter.value) !== null;
+      let re = new RegExp(filter.value, 'i');
+      match = space.attributes.name.match(re) !== null;
     }
     return match;
   }
--- a/src/app/profile/my-spaces/my-spaces.component.spec.ts
+++ b/src/app/profile/my-spaces/my-spaces.component.spec.ts
@@ -15,9 +15,7 @@ import { spaceMock } from '../../shared/context.service.mock';
 import { EventService } from '../../shared/event.service';
 import { MySpacesComponent } from './my-spaces.component';
 
-
 describe('MySpacesComponent', () => {
-
   let fixture: ComponentFixture<MySpacesComponent>;
   let component: DebugNode['componentInstance'];
   let mockContexts: any = jasmine.createSpy('Contexts');
@@ -98,7 +96,7 @@ describe('MySpacesComponent', () => {
     spaceMock1 = cloneDeep(spaceMock);
     spaceMock2 = cloneDeep(spaceMock);
     spaceMock2.id = '2';
-    spaceMock2.attributes.name = 'spaceMock2-id';
+    spaceMock2.attributes.name = 'spaceMock2-name';
     mockContexts.current = Observable.of(mockContext);
     mockUserService.loggedInUser = Observable.of(mockUser);
     mockGettingStartedService.createTransientProfile.and.returnValue(mockExtProfile);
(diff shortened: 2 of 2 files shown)


jiekang:
Hmm. I think all of the filtering toolbars are implemented with && behaviour. If we want to change this, I'd be inclined to change all of them prior to any of the changes hitting production, as the inconsistent behaviour would be very confusing.

Is there any documentation on case sensitivity? I think case insensitive by default works better if there is an option to enable case sensitivity via some special character prefix like `^`.

dlabrecq:
Regarding the logical OR comment, I believe the patternfly guidelines are specifically speaking about filter groups. That is, how to apply more than one filter, not case sensitivity. 

The patternfly-ng example is case sensitive, but you're free to implement case insensitive filtering. I don't believe there is a requirement one way or the other.

src/app/profile/my-spaces/my-spaces.component.ts
```diff
@@ -146,18 +146,16 @@ export class MySpacesComponent implements OnDestroy, OnInit {
   matchesFilter(space: Space, filter: Filter): boolean {
     let match = true;
     if (filter.field.id === 'name') {
-      match = space.attributes.name.match(filter.value) !== null;
+      match = space.attributes.name.toLocaleLowerCase().includes(filter.value.toLocaleLowerCase());
```
dlabrecq: Curious what results you get with "includes" Vs "match" when multiple filters are applied Just want to make sure the results still make sense. Can you provide a couple screenshots?

dlabrecq: I believe using "includes" Vs "match" breaks the patternfly pattern.

aptmac: Ah sure, I can revert the changes in that case.

aptmac:
`Regarding the logical OR, I believe the patternfly guidelines are specifically speaking about filters within a category (i.e., the checkbox filtering described by #2 below. We're not doing that here, but working with the behavior described by #1 below.`

Ah okay, I think this is where I was getting confused - I had thought that because our only category in the my-spaces filter is 'name' that they should be treated like #2. In which case, the original behaviour logged in issue https://github.com/openshiftio/openshift.io/issues/3857 was intentional, and this PR (and issue) should be closed?

dlabrecq:
Yes, this behavior is intentional. I believe making it case insensitive would still be ok.

joshuawilson:
Is this something we are trying to get done this sprint/train?

aptmac:
@dlabrecq that sounds good to me, I've changed the matchesFilter to be like your example; case-insensitive filtering works now.

@joshuawilson yes it should be doable by the end of the sprint/train, I had some confusion with the original issue and misinterpreted the intended functionality of the filter - this PR is now just a small QoL improvement.

PR #3085, opened 2018-06-27
home-dashboard: refactor widget loading

dlabrecq:
This change ensures the widgets are not shown in the loading state when the environment is reset and no spaces exist. The problem is that the spaces subscription does not return unless a space exists; thus, the loading state wasn't updated.

For the recent spaces widget, the loading widget was removed because there is nothing to load after the spaces have been retrieved.

Fixes https://github.com/openshiftio/openshift.io/issues/3705


alien-ike:
### Ike Plugins (test-keeper)

Thank you @dlabrecq for this contribution!



src/app/dashboard-widgets/recent-workspaces-widget/recent-workspaces-widget.component.ts
```diff
@@ -137,9 +136,11 @@ export class RecentWorkspacesWidgetComponent implements OnDestroy, OnInit {
         return Observable.of([]);
       }
       return Observable.forkJoin(val.map((space: ExtSpace) => {
+        this.loading = true;
         return this.fetchCodebases(space.id)
           .map(codebases => {
             space.codebases = codebases;
+            this.loading = false;
```
aptmac: The loading state only turns false if the Observable can be successfully subscribed to, what happens if the error block is triggered? Or if for whatever reason the `workItemService.getWorkspaces()` [0] throws an error? It looks like the widget will continue to display a loading state? Could there be tests that verify the behaviour of a successful/unsuccessful return from the service? Does this same behaviour apply to the `work-item-widget.component.ts` changes as well?

[0] https://github.com/dlabrecq/fabric8-ui/blob/2a8630f1f60537f9cf97bbf0073aca19a47d81d1/src/app/dashboard-widgets/recent-workspaces-widget/recent-workspaces-widget.component.ts#L114

dlabrecq: Changed the code to set the loading state after the catch.

This PR is limited to the space subscriptions used with the home dashboard. The same behavior does not need to be applied to other widgets used on the space dashboard because that code runs within the context of a space and the loading state is set properly.

aptmac:
Ah, with regards to my comment about the work-items-widget I meant more along the lines of if the loading boolean will be toggled if there is an error with the observable; if [[1]](https://github.com/dlabrecq/fabric8-ui/blob/bdff260741dd90607fda872de1c98293bb4f37ba/src/app/home/work-item-widget/work-item-widget.component.ts#L97) will be reached if the workItemsService.resolveCreator() fails or something else in the previous observable related calls. I guess the same applies to [[2]](https://github.com/dlabrecq/fabric8-ui/blob/bdff260741dd90607fda872de1c98293bb4f37ba/src/app/dashboard-widgets/recent-pipelines-widget/recent-pipelines-widget.component.ts#L39)? The recent-workspace-widget change you made worked for me when I made it throw an error so that's nice.

With regards to the testing I think it'd be nice to have something along the lines of ..
`it('should set loading to false if an error occurs with X-service', () => { ... }` and test when the service returns a valid mocked observable vs. something like Observable.throw('something')

[1] https://github.com/dlabrecq/fabric8-ui/blob/bdff260741dd90607fda872de1c98293bb4f37ba/src/app/home/work-item-widget/work-item-widget.component.ts#L97
[2] https://github.com/dlabrecq/fabric8-ui/blob/bdff260741dd90607fda872de1c98293bb4f37ba/src/app/dashboard-widgets/recent-pipelines-widget/recent-pipelines-widget.component.ts#L39

dlabrecq:
Added a recent-workspace-widget unit test for when an error is thrown.

dlabrecq:
Had trouble applying a similar test for work-item-widget.component.spec.ts -- setting up the mocks to step into that part of the code is overly complicated. I'll try returning to that when time frees up.

joshuawilson:
Is this something you are trying to get done for this sprint/train?

aptmac:
This looks like it should go in, I was initially skeptical of the test coverage we have but it looks pretty good as far as the work-items-widget is concerned:
2018-07-17-091654_809x60_scrot

As far as the other two files are concerned, the recent-workspaces spec coverage seems to only hit the happy paths (non-erroring) which is outside the scope of this PR (and also has good coverage), and the recent-pipelines spec doesn't even exist (which is a different issue itself).

PR #3103, opened 2018-07-02
Consuming one PlannerModule and got the /board back

sanbornsen:
## What does this PR do?

The new version (0.50.12) of planner exports a **PlannerModule** along with the PlannerList, PlannerBoard, and WorkItemDetailModule. It has routes configured for /list, /detail and /board. 

In this PR `fabric8-ui` consumes only `PlannerModule` on `/plan` path and rest is taken care of from planner.


src/app/layout/header/menus.service.ts
```diff
@@ -45,10 +45,10 @@ export class MenusService {
                 name: 'Backlog',
                 path: ''
               }
-              // , {
-              //   name: 'Board',
-              //   path: 'board'
-              // }
+              , {
```
joshuawilson: this is just a knit-pick but the comma should be on the line above

joshuawilson:
This should be easier for you to maintain but you will need to be aware/sensitive to coming navigation changes.

joshuawilson:
the text for the board view is behind the nav bar, but I'll merge this as that is not related to this PR

PR #3122, opened 2018-07-03
Secondary tab styling

andrewazores:
Removes custom, non-standard styling of secondary tab navigation elements in the Settings and Profile pages, and applies Patternfly standard styling instead.

https://github.com/openshiftio/openshift.io/issues/3889


andrewazores:
@joshuawilson rebased

PR #3124, opened 2018-07-04
fix(profile):  404 error while opening My Work Items from Profile page 

rohitkrai03:
Fixes https://github.com/openshiftio/openshift.io/issues/3868
Fixes https://github.com/fabric8-services/fabric8-auth/issues/537
- Using username of the current space owner instead of current logged in user.


sbose78:
@christianvogt @joshuawilson could we merge this please ?

PR #3154, opened 2018-07-16
fix(cors): remove `api` subdomain from wit api url

christianvogt:
In today's platform meeting, @aslakknutsen requested we change API calls from `https://api.openshift.io/api/` to `https://openshift.io/api/`.

Tested dev with prod and prod-preview environments.

- updated environment scripts
- updated webpack dev fallback
- removed wit prefix from `api-locator.service.ts`

### DIFF
--- a/config/webpack.dev.js
+++ b/config/webpack.dev.js
@@ -26,7 +26,7 @@ const HMR = helpers.hasProcessFlag('hot');
 // if env is 'inmemory', the inmemory debug resource is used
 const FABRIC8_FORGE_API_URL = process.env.FABRIC8_FORGE_API_URL || 'https://forge.api.prod-preview.openshift.io';
 const FABRIC8_FEATURE_TOGGLES_API_URL = process.env.FABRIC8_FEATURE_TOGGLES_API_URL;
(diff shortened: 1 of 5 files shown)


alien-ike:
### Ike Plugins (test-keeper)

christianvogt:
@jiekang Joshua and I looked for a deployment config that has the wit url and couldn't find one. I think the locator service constructs the urls for prod.

jiekang:
@christianvogt Okay, sounds good. A quick search shows the following two also might need changes. What do you think?
https://github.com/fabric8-ui/fabric8-planner/blob/15bcc7f5ae9e6551e37c46120bd8319d9f21152e/cico_test_prod.sh
https://github.com/fabric8io/fabric8-pipeline-library/blob/b1478e95f0ce49d032007c026b2b2b5696c84c7e/vars/deployOpenShiftSnapshot.groovy

It looks like those are the only two that have recent commits. There are a lot of repositories that haven't had commits since 2017

christianvogt:
Thanks @jiekang. I knew about the planner one but not the other.

aslakknutsen:
>  A quick search shows the following two also might need changes. What do you think?

Just to note, both URLs will still work. It's just a way to avoid in-flight checks in prod. The UI would need to be hosted from 'openshift.io' to work, which is not the cause on PR deployments so technically it doesn't matter which URL it use.

PR #3166, opened 2018-07-17
Recent spaces widget

andrewazores:
This PR includes several refactoring commits, a test addition commit, and one commit that should address a simple recent bug report [0].

[0] https://github.com/openshiftio/openshift.io/issues/4003

### DIFF
--- a/src/app/home/recent-spaces-widget/recent-spaces-widget.component.html
+++ b/src/app/home/recent-spaces-widget/recent-spaces-widget.component.html
@@ -1,0 +1,44 @@
+<div class="card-pf {{cardSizeClass}}">
+  <div class="card-pf-heading f8-card-heading">
+    <div class="card-pf-heading-details f8-card-heading-details">
+      <a class="btn btn-link f8-card-heading-btn-link" (click)="showAddSpaceOverlay()">
+        <i class="pficon pficon-add-circle-o"></i>
+      </a>
+    </div>
+    <h2 class="card-pf-title">
+      Recent Spaces
+    </h2>
+  </div>
+  <div class="card-pf-body f8-card-body">
+    <div class="f8-blank-slate-card" *ngIf="_spaces.length === 0">
+      <h3>Create a space</h3>
+      <p>Learn how to get started
+        <a href="https://docs.openshift.io/user-guide.html#getting_started_with_openshift_io" target="top">
+          in the documentation.
+        </a>
+      </p>
+      <div class="blank-slate-pf-main-action">
+        <button class="btn btn-primary btn-lg" (click)="showAddSpaceOverlay()"> Create a Space </button>
+      </div>
+    </div>
+    <!-- A list of the 5 most recent Spaces will be displayed in place of the Blank Slate. -->
+    <div class="f8-home-space-list-result" *ngIf="_spaces.length > 0">
+      <ul class="list-group list-view-pf list-view-pf-view list-view-pf-striped">
+        <li class="list-group-item" *ngFor="let space of recent">
+          <div class="list-view-pf-main-info">
+            <div class="list-view-pf-body">
+              <div class="list-view-pf-description">
+                <div class="list-group-item-text">
+                  <a [routerLink]="['/', space.relationalData.creator.attributes.username, space.attributes.name]" tooltip="{{space.attributes.name | spaceName}}"
+                    placement="bottom">
+                    {{space.attributes.name | spaceName}}
+                  </a>
+                </div>
+              </div>
+            </div>
+          </div>
+        </li>
+      </ul>
+    </div>
+  </div>
+</div>
--- a/src/app/home/recent-spaces-widget/recent-spaces-widget.component.html
+++ b/src/app/home/recent-spaces-widget/recent-spaces-widget.component.html
@@ -9,7 +9,7 @@
       Recent Spaces
     </h2>
   </div>
-  <div class="card-pf-body f8-card-body">
+  <div class="card-pf-body {{cardBodySizeClass}}">
 
     <div *ngIf="(userHasSpaces | async) then spaceList; else emptyState"></div>
     <ng-template #emptyState>
--- a/src/app/home/recent-spaces-widget/recent-spaces-widget.component.html
+++ b/src/app/home/recent-spaces-widget/recent-spaces-widget.component.html
@@ -28,7 +28,7 @@
     <ng-template #spaceList>
       <div class="f8-home-space-list-result">
         <ul class="list-group list-view-pf list-view-pf-view list-view-pf-striped">
-          <li class="list-group-item" *ngFor="let space of recent">
+          <li class="list-group-item" *ngFor="let space of (recentSpaces | async)">
             <div class="list-view-pf-main-info">
               <div class="list-view-pf-body">
                 <div class="list-view-pf-description">
--- a/src/app/home/recent-spaces-widget/recent-spaces-widget.component.html
+++ b/src/app/home/recent-spaces-widget/recent-spaces-widget.component.html
@@ -10,35 +10,41 @@
     </h2>
   </div>
   <div class="card-pf-body f8-card-body">
-    <div class="f8-blank-slate-card" *ngIf="_spaces.length === 0">
-      <h3>Create a space</h3>
-      <p>Learn how to get started
-        <a href="https://docs.openshift.io/user-guide.html#getting_started_with_openshift_io" target="top">
-          in the documentation.
-        </a>
-      </p>
-      <div class="blank-slate-pf-main-action">
-        <button class="btn btn-primary btn-lg" (click)="showAddSpaceOverlay()"> Create a Space </button>
+
+    <div *ngIf="_spaces.length === 0 then emptyState; else spaceList"></div>
+    <ng-template #emptyState>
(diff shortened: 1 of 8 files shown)


aptmac:
Nice, these changes look good, great test coverage as well!

2018-07-18-095038_757x62_scrot

src/app/home/recent-spaces-widget/recent-spaces-widget.component.ts
```diff
@@ -0,0 +1,82 @@
...
+      .subscribe(this.userHasSpaces);
+  }
+
+  showAddSpaceOverlay(): void {
+    this.broadcaster.broadcast('showAddSpaceOverlay', true);
+  }
+
+  private static compareSpaces(a: Space, b: Space): boolean {
+    return (a.attributes.name === b.attributes.name)
+    && (a.relationalData.creator.attributes.username === b.relationalData.creator.attributes.username);
```
dlabrecq: I'm thinking that we should we fix `Context` so that it returns unique recent spaces. That is, instead of each component doing this separately.

src/app/home/recent-spaces-widget/recent-spaces-widget.component.ts
```diff
@@ -0,0 +1,82 @@
...
+      .flatMap((context: Context): Observable<Context> => {
+        if (context && context.user) {
+          return Observable.of(context);
+        } else {
+          return Observable.throw('Failed to retrieve list of spaces owned by user');
+        }
+      })
+      .flatMap((context: Context): Observable<boolean> => this.spaceService
+        .getSpacesByUser(context.user.attributes.username)
+        .map((spaces: Space[]): boolean => spaces.length > 0)
```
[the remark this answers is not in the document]
andrewazores: The `uniqWith` is the actual fix for the duplicate spaces issue, I tried to mark this by using the "fix" verb on that specific commit only but this definitely isn't very visible.

I can certainly add a comment here, but you're right that fixing the `Spaces` to not emit duplicates to begin with would be better, and would remove that logic from this component so no need for a comment. I was avoiding touching the `Spaces` (really the `SpacesService` implementation) because it does touch so many places and is consumed by so many other spaces and components, and I have no idea if any of those might for some reason rely on the fact that it does emit duplicates.

src/app/home/recent-spaces-widget/recent-spaces-widget.component.ts
```diff
@@ -0,0 +1,82 @@
...
+    private errorHandler: ErrorHandler,
+    private broadcaster: Broadcaster,
+    private logger: Logger
+  ) {
+    this.recentSpaces = this.spaces.recent
+      .map((spaces: Space[]): Space[] => uniqWith(spaces, RecentSpacesWidget.compareSpaces));
+  }
+
+  ngOnInit(): void {
+    (this.router.url.startsWith('/_home') ? this.contexts.default : this.contexts.current)
```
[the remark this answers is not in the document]
andrewazores: `git blame` points to this commit [0] as the origin of this bit of logic, with the last small revision to it being [1].

I'm not exactly sure what "menus" are being referenced here. The `Context` is used just to grab the owning user's username and use that for the `SpaceService` query. I'm not sure why this widget would need to be able to display recent spaces according to various different contexts that a user could be looking at. Since we are currently only displaying it on the home page, I could eliminate this logic and just make the assumption that we're on the homepage and use the default context.

[0] https://github.com/fabric8-ui/fabric8-ui/commit/698ce51c3f193e5c30bffbe949dd779f5a3df4cb

[1] https://github.com/fabric8-ui/fabric8-ui/commit/493516db01a0f68836aa7968643311a4c11e79b2

andrewazores:
@dlabrecq I've updated this, removing the dead code for checking the router URL and adding a comment about the `uniqWith`. I haven't gone through with modifying the `Spaces` object yet though I agree that sounds like it makes more sense, but I'm just reluctant to change the behaviour of something used so widely.

joshuawilson:
@dlabrecq can you please review again

src/app/home/recent-spaces-widget/recent-spaces-widget.component.html
```diff
@@ -0,0 +1,50 @@
...
+    <div class="card-pf-heading-details f8-card-heading-details">
+      <a class="btn btn-link f8-card-heading-btn-link" (click)="showAddSpaceOverlay()">
+        <i class="pficon pficon-add-circle-o"></i>
+      </a>
+    </div>
+    <h2 class="card-pf-title">
+      Recent Spaces
+    </h2>
+  </div>
+  <div class="card-pf-body f8-card-body">
```
aptmac: While you're here, the widget seems to have the f8-card-body height of 375px [[0]](https://github.com/fabric8-ui/fabric8-ui/blob/master/src/assets/stylesheets/shared/_cards.less#L17) regardless of if it's large or not. 

Example: here's a exp/internal flagged large f8-card, but since the height is maxed out at 375px it doesn't fill the component
recent-spaces-cutoff

[0] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/assets/stylesheets/shared/_cards.less#L17

src/app/home/recent-spaces-widget/recent-spaces-widget.component.ts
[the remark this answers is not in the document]
dlabrecq: I'm fine with not touching SpacesService, but could you please log an issue against the REST API instead? The API should provide unique spaces, not duplicates -- the UI shouldn't be forced to prune them out.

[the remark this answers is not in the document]
andrewazores: It's actually not backed by a REST API:

https://github.com/fabric8-ui/fabric8-ui/blob/5e2074265445165adf27759fb4d42a7296b01c50/src/app/shared/spaces.service.ts#L35

There is a bit of logic there that's meant to eliminate duplicates, but clearly it's not working as intended since there's a bug report showing exactly that happening. I'll open a separate PR to fix the duplication logic there and remove the uniqWith from this one.

src/app/home/recent-spaces-widget/recent-spaces-widget.component.html
aptmac: I think the large f8-card needs a different (and larger) height dimension. Or potentially just fill the component, the spacesservice.recent only returns at max 8 spaces which isn't enough to completely fill the large card and cause the scrollbar anyways.

andrewazores: Ah, so it's the combination of f8-card-lg and f8-card-body that isn't working together well? Is this something that should be fixed in the shared CSS then?

src/app/home/recent-spaces-widget/recent-spaces-widget.component.ts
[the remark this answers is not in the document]
dlabrecq: Sounds good to me

src/app/home/recent-spaces-widget/recent-spaces-widget.component.html
aptmac: Yeah, so for the exp/internal you're passing the f8-card-lg to the recent-spaces component, so that the entire component becomes a large card [0]. However, the body of the card is still a f8-card-body [1], and there isn't a distinct CSS rule for displaying large cards. I'm thinking you could probably nest a `&-body` to match `f8-card-lg-body` under the f8-card rule [2] to adjust for the size, and then have the card body [1] be `{{cardSizeClass}}`.

[0] https://github.com/andrewazores/fabric8-ui/blob/d85f39ddc234e94aab8365d5124eb166c004722c/src/app/home/recent-spaces-widget/recent-spaces-widget.component.html#L1
[1] https://github.com/andrewazores/fabric8-ui/blob/d85f39ddc234e94aab8365d5124eb166c004722c/src/app/home/recent-spaces-widget/recent-spaces-widget.component.html#L12
[2] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/assets/stylesheets/shared/_cards.less#L10

andrewazores: I see what you mean, thanks. Does nesting it as `f8-card-lg-body` make sense though? Seems like it goes against the naming convention, which would have it as `f8-card-body-lg`. This makes it a little messier to specify the sizing of the component since now there are two inputs that are passed in, but that's only a temporary workaround to deal with feature flagging anyway.

src/assets/stylesheets/shared/_cards.less
```diff
@@ -20,6 +20,9 @@
       overflow-x: hidden;
       .list-group { padding-bottom: 0; }
       .list-view-pf-view { margin-top: -1px; }
+      &-lg {
+        max-height: 585px;
```
aptmac: The component looks better to me, thanks, but why in particular is it 585 px here?

andrewazores: f8-card has a standard height of 440px and a -lg height of 650px, so given f8-card-body height of 375px I gave f8-card-body-lg 585px. +210px either way. *shrug*

aptmac: Ah makes sense, just wanted to confirm where the numbers came from, looks good to me.

PR #3179, opened 2018-07-19
fix(collaborators): filter out already added collaborators from search results

rohitkrai03:
Fixes https://github.com/openshiftio/openshift.io/issues/4011
Fixes https://github.com/fabric8-services/fabric8-auth/issues/555

- Filters out already added collaborators from the search result in add collaborators search box.


joshuawilson:
based on the discussion on openshiftio/openshift.io#4011 I think we should put this on hold

rohitkrai03:
@joshuawilson Sure.

PR #3180, opened 2018-07-19
fix(collaborators): fixed component height and vertical scroll enabled for infinite scroll

rohitkrai03:
Fixes https://github.com/openshiftio/openshift.io/issues/4010
Fixes https://github.com/fabric8-services/fabric8-auth/issues/554

- Set `max-height: 680px` and `overflow-y:scroll` to enable vertical scroll for collaborators component which will trigger scroll event for infiniteScroll directive.

### DIFF
--- a/src/app/space/settings/collaborators/collaborators.component.less
+++ b/src/app/space/settings/collaborators/collaborators.component.less
@@ -28,3 +28,8 @@
 .collab-username {
   font-weight: normal;
 }
+
+.collaborator-list-scroll {
+  max-height: 680px;
+  overflow-y: scroll;
+}
--- a/src/app/space/settings/collaborators/collaborators.component.less
+++ b/src/app/space/settings/collaborators/collaborators.component.less
@@ -31,5 +31,5 @@
 
 .collaborator-list-scroll {
   max-height: 680px;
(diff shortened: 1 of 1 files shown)


christianvogt:
Won't this create a scroll area within the body scroll that will be a weird experience when the vertical space available on screen is less than `680px`?
Also I think you want `overflow-y: auto` instead so that the scrollbar isn't always present on some browsers.

rohitkrai03:
> Won't this create a scroll area within the body scroll that will be a weird experience when the vertical space available on screen is less than 680px?

This is how it has been handled in other components as well. I think there can be better way to handle this like having infinite scroll directly on body and no `max-height` for the component which will lead to having a single scroll experience but currently there are limitations in infinite scroll directive. See comment - https://github.com/openshiftio/openshift.io/issues/4010#issuecomment-406012293 Or maybe UX team can come up will a better interaction here.

> Also I think you want overflow-y: auto instead so that the scrollbar isn't always present on some browsers.
Updated it.

christianvogt:
It's an improvement so I'm ok with the fixed height until we have a better solution.

rohitkrai03:
@joshuawilson This one can be merged.

PR #3191, opened 2018-07-24
Fix Analyze overview feature flags

jiekang:
This addresses
https://github.com/openshiftio/openshift.io/issues/2908

for the Analyze Overview component, which is the 'home' page for a Space. The directives are no longer instantiated when hidden via feature flags. This prevents useless code from being executed, especially the http requests.

The second commit adds a test for this that fails without and passes with, but it is a little bit ugly. If anyone has suggestions on how to write a better test, I would appreciate it. I don't mind including or excluding it, as it is subject to change if the directives within a feature flag level changes, or the feature flags are removed for this component.

Thoughts?

### DIFF
--- a/src/app/space/analyze/analyze-overview/analyze-overview.component.ts
+++ b/src/app/space/analyze/analyze-overview/analyze-overview.component.ts
@@ -15,7 +15,7 @@ import { AuthenticationService, User, UserService } from 'ngx-login-client';
 })
 export class AnalyzeOverviewComponent implements OnInit, OnDestroy {
   context: Context;
-  isUserEnabled: boolean = false;
+  _isUserEnabled: boolean;
   isLoggedIn: boolean;
 
   private loggedInUser: User;
@@ -48,9 +48,15 @@ export class AnalyzeOverviewComponent implements OnInit, OnDestroy {
     this.subscriptions.push(
       this.featureTogglesService
         .getFeature('Analyze.newSpaceDashboard')
-        .subscribe((feature: Feature) => {
-          this.isUserEnabled = feature.attributes.enabled && feature.attributes['user-enabled'];
-        })
+        .timeout(2000)
+        .subscribe(
+          (feature: Feature) => {
+            this._isUserEnabled = feature.attributes.enabled && feature.attributes['user-enabled'];
+          },
+          () => {
+            this._isUserEnabled = false;
+          }
+        )
     );
 
     this.subscriptions.push(
@@ -75,6 +81,14 @@ export class AnalyzeOverviewComponent implements OnInit, OnDestroy {
     this.subscriptions.forEach((sub: Subscription) => {
       sub.unsubscribe();
     });
+  }
+
+  isUserEnabled(): boolean {
+    return this._isUserEnabled !== undefined && this._isUserEnabled;
+  }
+
+  isDefaultEnabled(): boolean {
+    return this._isUserEnabled !== undefined && !this._isUserEnabled;
   }
 
   showAddAppOverlay(): void {
--- a/src/app/space/analyze/analyze-overview/analyze-overview.component.ts
+++ b/src/app/space/analyze/analyze-overview/analyze-overview.component.ts
@@ -1,12 +1,11 @@
 import { Component, OnDestroy, OnInit, ViewEncapsulation } from '@angular/core';
 
-import { Broadcaster } from 'ngx-base';
-import { Context, Contexts, Space } from 'ngx-fabric8-wit';
-import { AuthenticationService, User, UserService } from 'ngx-login-client';
 import { Subscription } from 'rxjs';
 
-import { FeatureTogglesService } from 'ngx-feature-flag';
-
+import { Broadcaster } from 'ngx-base';
+import { Context, Contexts } from 'ngx-fabric8-wit';
+import { Feature, FeatureTogglesService  } from 'ngx-feature-flag';
+import { AuthenticationService, User, UserService } from 'ngx-login-client';
 
 @Component({
   encapsulation: ViewEncapsulation.None,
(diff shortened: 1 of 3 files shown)


andrewazores:
`fix(analyze-overview)` in the commit message of the second commit should be `test(analyze-overview)`

andrewazores:
If this is using `*ngIf` on a variable that defaults to `false` at first and is then set to `true` later asynchronously, isn't it still going to initially load the `false` branch, including initializing the child views (which fires off the unwanted HTTP requests)?

jiekang:
Hmm good point. In my user testing the components that were not visible were not being initialized, but I only looked at one case. I will look at the other too. I'm not sure if we can mitigate this. I think this will help the components be destroyed, at least.

jiekang:
A better long term fix for this and the home page could be to:

1. extract all widgets as separate components
2. create parent components for each feature level that arrange widgets as they please
3. add dynamic routing to the parent component based on feature level, for example, if visiting _home, route to A if >= beta, otherwise route to B

andrewazores:
Yea, that would definitely sidestep the problem of having toggled-off features still loaded into the DOM and initialized, although it's definitely more work to implement (IMO worth it). It would also avoid issues like https://github.com/openshiftio/openshift.io/issues/4052 . My only concern would be that this would lead to users of differing feature levels to potentially be unable to share links to specific application views with each other. This means it would also break user bookmarks if they change feature levels and a bookmarked page has a changed feature level route.

jiekang:
@andrewazores You're right. When adding a delay of 1000 ms to the feature toggle service, the default level components are constructed. However, we can't add a bottle-neck to this page behind the feature toggles service. If feature toggle service is unavailable, the page should load the default level. We could incorporate a timeout and loading state. Thoughts?

andrewazores:
Yea that makes sense. It sounds like something that should be possible to incorporate in the upstream fabric8-ui/ngx-feature-flag library, probably in the feature-toggle.component . Then all consumers get this behaviour for free.

christianvogt:
Feature toggle service should not load the child components until it has fetched the appropriate data from the server in order to prevent unnecessary loading of components and other potential data fetches. Do not show the default.... By default :)

While it could also introduce a loading state, the typical response time of feature toggles should be quick and therefore it would be more appropriate to simply show a blank area vs flashing a loading indicator. So I'm in favor of no loading state.

This should be solved at the feature toggle level to affect all downstream consumers. And not as part of this pr.

jiekang:
Yeah, I wasn't sure what could be done with routes and you're right, a async decision there would be pretty bad. I'll look into the general idea more I guess. The README for https://github.com/fabric8-ui/ngx-feature-flag talked about dynamic components that I wanted to look at.

jiekang:
@christianvogt if feature toggles service is unavailable, then the application becomes unusable?

andrewazores:
An error response or a long timeout can be considered as a form of "appropriate response" to the deferred decision to load the default level ;)

jiekang:
The default 'production' level always exists so why shouldn't it be loaded if the toggles service is unavailable?

christianvogt:
I didn't know what you meant by unavailable. If the service is truly unavailable then yes load the default (when does this happen?). I thought you meant that while the toggle level is unavailable due to the service still loading the response, in which case do not load the default but continue to wait.

Yes an error or timeout should result in default :)

src/app/space/analyze/analyze-overview/analyze-overview.component.ts
```diff
@@ -15,7 +15,7 @@ import { AuthenticationService, User, UserService } from 'ngx-login-client';
 })
 export class AnalyzeOverviewComponent implements OnInit, OnDestroy {
   context: Context;
-  isUserEnabled: boolean = false;
+  _isUserEnabled: boolean;
```
andrewazores: Does leaving this `undefined` have any functional effect on the `*ngIf`? I'd expect it to still evaluate falsy, same as if it were initialized to `false`.

jiekang: I altered the `*ngIf` to use a function that was added as well. Or at least I should have; I seem to be making random mistakes with merging commits and what not today.

andrewazores: Right, but as far as the `*ngIf` conditional goes, `false`, `null`, and `undefined` all evaluate falsy and produce the same result. So the expression used in the new function is going to produce the same result in the `undefined` case as the old case of just initializing to `false` off the bat.

jiekang: Oh... yeah good point.

jiekang: Right, they need to actually be different for the two divs. Bleh.

jiekang: Bleh; I was looking for a small, neat fix to cover the three states.

jiekang:
Maybe with some significant changes it might be usable. The feature flag component asks users to embed their user and default level DOM elements within the DOM element of the feature flag like:

```
foo.html:
<f8-feature-toggle featureName="bar">
  <one user-level></one>
  <two default-level></two>
<f8-feature-toggle>
```

The feature flag component then accesses these to hide/show them via it's template:
```
template: `<ng-content *ngIf="isEnabled" select="[user-level]"></ng-content><ng-content *ngIf="!isEnabled" select="[default-level]"></ng-content>`
```

The issue here is that Angular will evaluate foo.html and construct the feature flag, one and two components, and when it initializes the feature flag component, it will then re-evaluate the DOM to hide the appropriate child component. This behaviour is not desirable. In this PR, the use of the feature flag component is dropped for this reason.

I've been brainstorming some alternatives, like possibly supplying the component selector name as input to be used in the feature flag template so the directives don't appear in foo.html, but I haven't tried anything yet. I'll be tinkering around now.

andrewazores:
Hmm, that makes sense. I'll tinker around too, I'm sure we can come up with something widely and easily reusable with this idea.

jiekang:
Oh nice. That looks pretty good!

corinnekrych:
@andrewazores I like the proposal!
Pls do a PR on ngx-feature-flag and I'll merge it today and release 0.5.0. So that @jiekang can update/redo this PR.

joshuawilson:
please rebase

jiekang:
This is no longer relevant after [1]. Closing.

[1] https://github.com/fabric8-ui/fabric8-ui/pull/3202

PR #3194, opened 2018-07-24
Deployments upgrade to HttpClient

andrewazores:
This PR upgrades the DeploymentApiService from the old Angular Http to the newer HttpClient, and does some light refactoring on top.

### DIFF
--- a/src/app/space/create/deployments/services/deployment-api.service.ts
+++ b/src/app/space/create/deployments/services/deployment-api.service.ts
@@ -183,23 +183,25 @@ export class DeploymentApiService {
       .map((response: TimeseriesResponse) => response.data.attributes);
   }
 
-  deleteDeployment(spaceId: string, environmentName: string, applicationId: string): Observable<HttpResponse<any>> {
+  deleteDeployment(spaceId: string, environmentName: string, applicationId: string): Observable<void> {
     const encSpaceId: string = encodeURIComponent(spaceId);
     const encEnvironmentName: string = encodeURIComponent(environmentName);
     const encApplicationId: string = encodeURIComponent(applicationId);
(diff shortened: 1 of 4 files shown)


joshuawilson:
I think you need to update the `src/app/app.module.ts` too.

andrewazores:
@joshuawilson you mean to add the module import for the HttpClientModule, right? Or something else?

Should that module be added to the top-level app.module.ts now, or restricted to the DeploymentsModule?

joshuawilson:
I would think it best to add it to the top level so that we don't end up with it left behind in deployments. As long as you can have both old and new `http` apis co-exist. The whole app needs to move to that anyway.

src/app/space/create/deployments/services/deployment-api.service.ts
```diff
@@ -134,79 +136,83 @@ export interface SeriesData {
...
-    return this.http.delete(url, { headers: this.headers })
-      .catch((err: Response) => this.handleHttpError(err));
+  deleteDeployment(spaceId: string, environmentName: string, applicationId: string): Observable<void> {
+    const encSpaceId: string = encodeURIComponent(spaceId);
+    const encEnvironmentName: string = encodeURIComponent(environmentName);
+    const encApplicationId: string = encodeURIComponent(applicationId);
+    const url: string = `${this.apiUrl}${encSpaceId}/applications/${encApplicationId}/deployments/${encEnvironmentName}`;
+    return this.http.delete(url, { headers: this.headers, responseType: 'text' })
+      .catch((err: HttpErrorResponse) => this.handleHttpError(err))
+      .map(() => null);
```
rohitkrai03: @andrewazores Why do we need to map the value to `null` here?

rohitkrai03: Got it.

Issue #2296, opened 2017-12-06
Add unit tests for the space service


sbose78:
We use
https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/shared/spaces.service.ts
everywhere, hence we need to add unit tests for this.

aptmac:
fixed in https://github.com/fabric8-ui/fabric8-ui/pull/3201

PR #3201, opened 2018-07-26
test(SpacesService): add unit tests for SpacesService

aptmac:
This PR introduces some unit tests for the SpacesService, and is related to issues [[0]](https://github.com/openshiftio/openshift.io/issues/923) and [[1]](https://github.com/fabric8-ui/fabric8-ui/issues/2296).

SpacesService is widely used across fabric8-ui, and it's too bad we don't currently have any unit tests for it. The unit tests here aren't as comprehensive as they should be (and serve as a good starting point for additional tests to be added), but cover the majority of the lines/branches of code. I figured it'd be good to have some base level unit tests in place for when SpacesService gets tweaked to reduce the number of calls to /api/users [[2]](https://openshift.io/openshiftio/Openshift_io/plan/detail/157), that I have a WIP [[3]](https://github.com/aptmac/fabric8-ui/commit/c28aa0d580c80f3841d75164c27cb9c82132a9c3) for at the moment.

coverage

[0] https://github.com/openshiftio/openshift.io/issues/923
[1] https://github.com/fabric8-ui/fabric8-ui/issues/2296
[2] https://openshift.io/openshiftio/Openshift_io/plan/detail/157
[3] https://github.com/aptmac/fabric8-ui/commit/c28aa0d580c80f3841d75164c27cb9c82132a9c3


src/app/shared/spaces.service.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+      if (key === 'spaceUpdated') {
+        return Observable.of(mockSpace);
+      }
+    });
+
+    mockProfileService = jasmine.createSpyObj('ProfileService', ['silentSave']);
+    mockProfileService.current = Observable.of(mockProfile);
+
+    TestBed.configureTestingModule({
+      imports: [],
```
andrewazores: This can be omitted since it's empty

src/app/shared/spaces.service.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+import { Context, Contexts, Space, SpaceService } from 'ngx-fabric8-wit';
+import { User } from 'ngx-login-client';
+import { ConnectableObservable, Observable } from 'rxjs';
+import { createMock } from 'testing/mock';
+import { ExtProfile, ProfileService } from '../profile/profile.service';
+import { SpacesService } from './spaces.service';
+
+describe('SpacesService', () => {
+
+  let mockSpace: Space;
```
andrewazores: My preference is to use `TestBed.get` to access these mock instances from within individual specs, rather than holding onto top-level references. This goes nicely with the `useFactory` trick you're already using below for the `providers`.

Actually I see you do use `TestBed.get` in the specs below some of the time, so I'd prefer to use it consistently and avoid these top-level references.

src/app/shared/spaces.service.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+      ]
+    });
+  });
+
+  describe('#get current', () => {
+    it('should return the space from contexts.current', (done: DoneFn) => {
+      mockBroadcaster.on.and.returnValue(Observable.never());
+      const spacesService: SpacesService = TestBed.get(SpacesService);
+      let result: Observable<Space> = spacesService.current;
+      result.subscribe(r => {
```
andrewazores: `r` should have a type

src/app/shared/spaces.service.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+  describe('#saveRecent', () => {
+    it('should silentSave after a spaceChanged has been broadcasted', () => {
+      mockProfileService.silentSave.and.returnValue(Observable.of(mockUser));
+      mockBroadcaster.broadcast('spaceChanged', mockSpace);
+      let expectedPatch = {
+        store: {
+          recentSpaces: [mockSpace.id]
+        }
+      };
+      console.log = jasmine.createSpy('log');
```
andrewazores: I think it's preferable to do `spyOn(console, 'log')`

src/app/shared/spaces.service.spec.ts
joshuawilson: @aptmac is this something you can address or explain why not?

aptmac: Ah I should have commented when I fixed it, I already made all the changes to better used `TestBed.get()` in a follow up commit, see below: https://github.com/fabric8-ui/fabric8-ui/pull/3201/commits/2528d95a6b8f3d9169ab7dba78c77c667b4b22ee

andrewazores: Yea I reviewed that commit and approved the change. The remaining top-level references are just for mock data, not service references, which I think is fine.

joshuawilson: thanks

PR #3202, opened 2018-07-26
feat(feature-flag): update for feature-flag v0.5 compat

andrewazores:
This PR should be merged at the same time as the PR to upgrade ngx-feature-flag to v0.5. That update changes the feature-flag component syntax in a backward *in*compatible way, and this commit restores compatibility. The new feature flag component will also by default cause Angular to render *no* toggled components rather than *both* as in the old behaviour if the tests do not override the feature flag component/service, so a mock feature toggle component is included which simply renders the user-level component immediately and unconditionally.


rohitkrai03:
@andrewazores From what i could understand, the new feature-flag component syntax changed such that we need to define `<ng-template>` for both `userLevel` and `defaultLevel` explicitly. But i am a bit confused about the mock-feature-toggle component, that is just for unit tests?

andrewazores:
@rohitkrai03 correct, as per the description I gave in the original message: mock-feature-toggle component is there for unit tests, because without it, *neither* of the user-level or default-level templates will be rendered with the new feature-flag component while under testing. This would mean that feature-flagged components are untestable. The mock-feature-toggle component allows feature-flagged components to be rendered and tested.

The previous behaviour of the feature-toggle component still required devs to mark the user-level and default-level explicitly, but this was done using content projection, and selecting the projected content based on the user-level and default-level attributes. This causes both contents to be initialized, which is the problem we're trying to solve. The solution is to use template outlets, which require the toggled components to be placed within ng-templates, which are marked using template variables rather than attributes. Similar end syntax but (intentionally and purposefully) pretty different semantics and performance implications.

rohitkrai03:
@andrewazores Understood. Thanks for all the explanation.

PR #3209, opened 2018-07-30
fix(sentry): group errors with stack traces based on first frame

christianvogt:
This change affects the grouping of exceptions sent to Sentry. Errors with a stack trace now provide a custom `fingerprint` to sentry to group like errors based on the first frame of the stack trace. The assumption to agree on here is that where the error occurred is more important for grouping than how we got to the error.


alien-ike:
### Ike Plugins (test-keeper)

src/app/shared/exception.handler.ts
```diff
@@ -32,8 +32,42 @@ export class RavenExceptionHandler extends ErrorHandler {
         id: userService.currentLoggedInUser.id
       });
     }
+
     if (environment === 'prod-preview' || environment === 'production') {
-      Raven.captureException(err.originalException || err);
+      let ex = err.originalException || err;
```
jiekang: Can the fingerprint creation be moved to a separate function; it is more intuitive to modify later.

christianvogt:
@jiekang I cleaned up the implementation.
@joshuawilson added unit tests

Issue #157, opened 2017-02-17
Refresh token to avoid access token experation


alexeykazakov:
We need to refresh token if user is active. Otherwise the token will be expired and the user will be redirected to the login page.
Backend has all needed support for that:
https://github.com/almighty/almighty-core/pull/776
https://github.com/almighty/almighty-core/pull/780

I know that @dgutride is working on this but I didn't find any open UI issue for that. Sorry if it's a duplicate.

pmuir:
@dgutride please self assign if you are working on this.

dgutride:
@pmuir - I'm not able to assign myself, can you add me to the fabric8io org?

dgutride:
https://github.com/fabric8io/fabric8-ui/pull/194 - ngx-login-client was updated to auto refresh the token prior to expiration.  Closing...

PR #3211, opened 2018-07-31
fix(SpacesService): reduce number of calls to /api/users regarding recent spaces

aptmac:
Note: the last commit (regarding tests) in this PR will need to be refactored when https://github.com/fabric8-ui/fabric8-ui/pull/3201 is merged

This PR addresses OSIO defect #157 [0], in which the updating of 'recent space' causes a large number of PATCH calls to /api/users. Note that refactoring the SpacesService isn't enough to remove all the redundant calls, it appears that the ContextService is operating in a similar fashion and is responsible for most of the PATCH calls.

Currently, the _recent variable is a merge of the broadcaster 'spaceChanged' and addRecent (which gets updated for each space returned from `loadRecent()`) [1]. The problem lies with the `scan()` [2], which gets triggered for each space; loading 8 spaces via `loadRecent()` results in the subsequent block of code being executed 8 times - including `saveRecent()`. As a result, the user can expect to make `${_recent.length}` PATCH calls every time the service is initialized. In this PR, the 'spaceChanged' logic is separated and handled independently from the initial loading of recent spaces from the profileService, this way `saveRecent()` should be called only if 'spaceChanged' or 'spaceDeleted' is broadcasted.

Additionally, there is no code for removing deleted spaces from the list of recent spaces; the ContextService has such functionality by subscribing to the events service, but SpacesService currently does not. This could be seen as problematic and has been reported as a bug a handful of times [3]. In this PR, a message ('spaceDeleted') is broadcasted when a space is deleted, and causes the SpacesService to determine if that space is in the list of recent spaces, and if so, remove it [4].

I've included tests that cover all lines/functions/statements to try and capture the behaviour of the service.

Let me know what you think about this implementation, and if there is anything I've missed or should be on the look out for. Thanks!

**Screenshots**

**Before:**
/api/users requests from the chrome inspect network tab:
old-home

console warning every time a save occurs:
old-space

**After:**
/api/users requests from the chrome inspect network tab:
after-home

console warning every time a save occurs:
after-space

[0] https://openshift.io/openshiftio/Openshift_io/plan/detail/157
[1] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/shared/spaces.service.ts#L35
[2] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/shared/spaces.service.ts#L40
[3] https://github.com/openshiftio/openshift.io/issues/2196
[4] https://github.com/aptmac/fabric8-ui/blob/93e89febcb97dcea667ec3a36400294b34dc502c/src/app/shared/spaces.service.ts#L63

### DIFF
--- a/src/app/shared/spaces.service.ts
+++ b/src/app/shared/spaces.service.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@angular/core';
 import { Broadcaster } from 'ngx-base';
 import { Contexts, Space, Spaces, SpaceService } from 'ngx-fabric8-wit';
-import { ConnectableObservable, Observable, ReplaySubject, Subscription } from 'rxjs';
+import { Observable, ReplaySubject, Subscription } from 'rxjs';
 import { ExtProfile, ProfileService } from '../profile/profile.service';
 
 @Injectable()
@@ -10,7 +10,8 @@ export class SpacesService implements Spaces {
   static readonly RECENT_SPACE_LENGTH = 8;
 
   private _current: Observable<Space>;
-  private _recent: ConnectableObservable<Space[]>;
+  private _recent: ReplaySubject<Space[]> = new ReplaySubject<Space[]>(1);
+  private recentChanged: boolean = false;
 
   constructor(
     private contexts: Contexts,
@@ -31,50 +32,63 @@ export class SpacesService implements Spaces {
 
   private initRecent(): void {
     // load existing recent spaces from profile.store.recentSpaces
-    this._recent = this.loadRecent().publishReplay(1);
-    this._recent.connect();
+    this.loadRecent().subscribe(spaces => {
+      this._recent.next(spaces);
+    });
 
     // update _recent with newly seen space when spaceChanged is emitted
     this.broadcaster.on<Space>('spaceChanged')
-      .combineLatest(this._recent)
-      .subscribe((combined: [Space, Space[]]) => {
-        let changedSpace: Space = combined[0];
-        let recentSpaces: Space[] = combined[1];
-        let index: number = this.findRecentIndexById(recentSpaces, changedSpace.id);
-        // if changedSpace exists in _recent
-        if (index !== -1) {
-          // move the space to the front of the list if it is already in _recent
-          if (recentSpaces[0].id !== changedSpace.id) {
-            recentSpaces.splice(index, 1);
+      .flatMap((changedSpace: Space) => {
+        return this._recent.first().map((recentSpaces: Space[]) => {
+          let index: number = this.findRecentIndexById(recentSpaces, changedSpace.id);
+          // if changedSpace exists in _recent
+          if (index !== -1) {
+            // move the space to the front of the list if it is already in _recent
+            if (recentSpaces[0].id !== changedSpace.id) {
+              recentSpaces.splice(index, 1);
+              recentSpaces.unshift(changedSpace);
+              this.recentChanged = true;
+            }
+            // otherwise, no operation required; changedSpace is already index 0
+          // if changedSpace is new to the recent spaces
+          } else {
             recentSpaces.unshift(changedSpace);
-            this.saveRecent(recentSpaces);
+            // trim _recent if required to ensure it is length =< 8
+            if (recentSpaces.length > SpacesService.RECENT_SPACE_LENGTH) {
+              recentSpaces.pop();
+            }
+            this.recentChanged = true;
           }
-          // otherwise, no operation required; changedSpace is already index 0
-        // if changedSpace is new to the recent spaces
-        } else {
-          recentSpaces.unshift(changedSpace);
-          // trim _recent if required to ensure it is length =< 8
-          if (recentSpaces.length > SpacesService.RECENT_SPACE_LENGTH) {
-            recentSpaces.pop();
-          }
+          return recentSpaces;
+        });
+      })
+      .subscribe((recentSpaces: Space[]) => {
+        if (this.recentChanged) {
           this.saveRecent(recentSpaces);
+          this.recentChanged = false;
+          this._recent.next(recentSpaces);
         }
-      }
-    );
+      });
 
     // if a space is deleted, check to see if it should be removed from _recent
     this.broadcaster.on<Space>('spaceDeleted')
-      .combineLatest(this._recent)
-      .subscribe((combined: [Space, Space[]]) => {
-        let deletedSpace: Space = combined[0];
-        let recentSpaces: Space[] = combined[1];
-        let index: number = this.findRecentIndexById(recentSpaces, deletedSpace.id);
-        if (index !== -1) {
-          recentSpaces.splice(index, 1);
+      .flatMap((deletedSpace: Space) => {
+        return this._recent.map((recentSpaces: Space[]) => {
+          let index: number = this.findRecentIndexById(recentSpaces, deletedSpace.id);
+          if (index !== -1) {
+            recentSpaces.splice(index, 1);
+            this.recentChanged = true;
+          }
+          return recentSpaces;
+        });
+      })
+      .subscribe((recentSpaces: Space[]) => {
+        if (this.recentChanged) {
           this.saveRecent(recentSpaces);
+          this.recentChanged = false;
+          this._recent.next(recentSpaces);
         }
-      }
-    );
+      });
   }
 
   get current(): Observable<Space> {
@@ -99,7 +113,6 @@ export class SpacesService implements Spaces {
   }
 
   private saveRecent(recent: Space[]): Subscription {
-    this._recent.publish();
     let patch = {
       store: {
         recentSpaces: recent.map(val => val.id)
--- a/src/app/shared/spaces.service.ts
+++ b/src/app/shared/spaces.service.ts
@@ -58,6 +58,15 @@ export class SpacesService implements Spaces {
       // update the profile's store.recentSpaces attribute
       this.saveRecent(this._recent);
     });
+
+    // if a space is deleted, check to see if it should be removed from _recent
+    this.broadcaster.on<Space>('spaceDeleted').subscribe(space => {
+      let index: number = this.findRecentIndexById(space.id);
+      if (index !== -1) {
+        this._recent.splice(index, 1);
+        this.saveRecent(this._recent);
+      }
+    });
   }
 
   get current(): Observable<Space> {
--- a/src/app/shared/spaces.service.ts
+++ b/src/app/shared/spaces.service.ts
@@ -32,7 +32,7 @@ export class SpacesService implements Spaces {
 
   private initRecent(): void {
     // load existing recent spaces from profile.store.recentSpaces
-    this.loadRecent().subscribe(spaces => {
+    this.loadRecent().subscribe((spaces: Space[]) => {
       this._recent.next(spaces);
     });
 
@@ -40,7 +40,7 @@ export class SpacesService implements Spaces {
     this.broadcaster.on<Space>('spaceChanged')
       .flatMap((changedSpace: Space) => {
         return this._recent.first().map((recentSpaces: Space[]) => {
-          let index: number = this.findRecentIndexById(recentSpaces, changedSpace.id);
+          let index: number = recentSpaces.findIndex((space: Space) => space.id === changedSpace.id);
           // if changedSpace exists in _recent
           if (index !== -1) {
             // move the space to the front of the list if it is already in _recent
@@ -72,7 +72,7 @@ export class SpacesService implements Spaces {
     this.broadcaster.on<Space>('spaceDeleted')
       .flatMap((deletedSpace: Space) => {
         return this._recent.first().map((recentSpaces: Space[]) => {
-          let index: number = this.findRecentIndexById(recentSpaces, deletedSpace.id);
+          let index: number = recentSpaces.findIndex((space: Space) => space.id === deletedSpace.id);
           if (index !== -1) {
             recentSpaces.splice(index, 1);
             this.recentChanged = true;
@@ -118,13 +118,6 @@ export class SpacesService implements Spaces {
     } as ExtProfile;
     return this.profileService.silentSave(patch)
       .subscribe(() => {}, err => console.log('Error saving recent spaces:', err));
-  }
-
-  /**
-   * Given the id of a space, returns the index of the space in _recent
-   */
-  findRecentIndexById(spaces: Space[], id: string): number {
-    return spaces.findIndex(space => space.id === id);
   }
 
 }
--- a/src/app/shared/spaces.service.ts
+++ b/src/app/shared/spaces.service.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@angular/core';
 import { Broadcaster } from 'ngx-base';
 import { Contexts, Space, Spaces, SpaceService } from 'ngx-fabric8-wit';
-import { Observable } from 'rxjs';
+import { Observable, Subscription } from 'rxjs';
 import { ExtProfile, ProfileService } from '../profile/profile.service';
 
 @Injectable()
@@ -10,7 +10,7 @@ export class SpacesService implements Spaces {
   static readonly RECENT_SPACE_LENGTH = 8;
 
   private _current: Observable<Space>;
-  private _recent: Space[] = [];
+  private _recent: Observable<Space[]>;
 
   constructor(
     private contexts: Contexts,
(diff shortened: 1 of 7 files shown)


jiekang:
It seems like there could be overlap between recent contexts and recent spaces where recent spaces is just a subset of recent contexts that are spaces. Thoughts?

aptmac:
@jiekang, right, there are situations where recent spaces could be a subset or even the same as recent spaces (if the user is browsing multiple spaces), but also situations where they could vary greatly (if the user is browsing other user's profiles). Ultimately they perform similarly, and they both keep up to the last 8 recently viewed items saved. Their biggest difference seems to be where they are saved on the profile: https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/shared/spaces.service.ts#L98, and https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/shared/context.service.ts#L399

jiekang:
Okay, thanks for the explanation. I'm just probing to see if there's any way to reuse code for the two is they are quite related.

jiekang:
Can you clarify the reduction in calls? Is it now N instead of 2*N or something along those lines?

jiekang:
Also this is probably another PR worth of work but I think the recent spaces service should provide an Observable of recent spaces similar to how it provides an Observable of the current space.

aptmac:
It'd be a reduction of N, where N is the number of recent spaces stored on the profile. I'll try to give an example to explain it better.

Example: I have 8 recent spaces on my profile, spaces = [1, 2, 3, 4, 5, 6, 7, 8], with 1 being the most recent and 8 being the last recent. I start OSIO by going to space 8. Currently, I would see a PATCH request for each of the 8 spaces initially, plus an additional one for moving 8 to the front of recent spaces and saving it, (8 + 1). With this PR we would only see the request to update 8 to the front of the recent spaces (0 + 1).

jiekang:
Okay nice, thanks. So it used to be N but now it's a constant?

aptmac:
More like it used to be N but now it just doesn't do it. It was because of the `scan()` after merging the two observables at [0], it would run the code inside the `scan()` for each space it encountered including the N from the initial load. The solution here is to treat the initially loaded information and 'spaceChanged' broadcasted spaces separately.

[0] https://github.com/fabric8-ui/fabric8-ui/pull/3211/files#diff-103682c5eace85417d403c7c7643dcb5L40

jiekang:
Er actually it looks like the change to the recent from Observable to array is from this PR. It should be modified back to being an observable. There should be a way to use combinations of observable constructs to get the right behaviour...

aptmac:
Yeah sure I can rework it with observable operations instead

joshuawilson:
please rebase

jiekang:
I think my request to maintain recent as an observable has led to a bit of a complication. There are some timing issues with how it's currently written, as users access the service through a get function that returns a variable that changes it's value.

I'll try to describe with an example scenario:

User calls _recent and acquires a reference to the value. The initial return value is possibly Observable.of([]), an observable that emits the empty array. The user subscribes to this, and receives the emission. As this is an Observable.of, the observable also completes.

Some time later, a space addition is made, and a broadcast of spaceCreated occurs. This triggers a change in the space services value to Observable.of(_recent), but the user above will not see this unless calling the _recent getter function again.

The expectation should be that the original Observable would emit the new recent spaces so that the user can handle it in his subscription. To do this, the _recent should be a Subject (possibly ReplaySubject), and when any events occur, it emits an update to it's subscribers.

Does that make sense?

src/app/shared/spaces.service.ts
```diff
@@ -1,72 +1,80 @@
...
-      // The scan operator allows us to access the list of recent contexts and add ours
-      .scan((recent, s) => {
-        // First, check if this context is already in the list
-        // If it is, remove it, so we don't get duplicates
-        for (let i = recent.length - 1; i >= 0; i--) {
-          if (recent[i].id === s.id) {
-            recent.splice(i, 1);
+    // update _recent with newly seen space when spaceChanged is emitted
+    this.broadcaster.on<Space>('spaceChanged')
+      .combineLatest(this._recent)
```
jiekang: This usage of combineLatest isn't necessary as the array is accessible already. The broadcast event will emit the new space, and can be subscribed directly to receive it, then manipulating the array with the new space.

aptmac:
@jiekang Ah got it, yeah I had interpreted it as keeping the manipulation within observable operations, but that's a good example as to what behaviour should be expected, thanks!

joshuawilson:
Is this still being worked on? Or is it ready?

jiekang:
@joshuawilson I think it's still being worked on. Alex is away this week.

aptmac:
@jiekang I changed the `_recent` to be a `ConnectableObservable`, as it is currently, and have it publish values when the recent spaces are updated. Let me know what you think about that approach.

src/app/shared/spaces.service.ts
```diff
@@ -92,14 +98,27 @@ export class SpacesService implements Spaces {
     });
   }
 
-  private saveRecent(recent: Space[]) {
+  private saveRecent(recent: Space[]): Subscription {
+    this._recent.publish();
```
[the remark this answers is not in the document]
jiekang: I think you want to be looking at using a ReplaySubject so you can push a value for it to emit to it's listeners when saving the recent spaces.

src/app/shared/spaces.service.ts
```diff
@@ -1,72 +1,80 @@
...
-          val.splice(
-            SpacesService.RECENT_SPACE_LENGTH,
-            val.length - SpacesService.RECENT_SPACE_LENGTH
-          );
+      }
+    );
+
+    // if a space is deleted, check to see if it should be removed from _recent
+    this.broadcaster.on<Space>('spaceDeleted')
+      .combineLatest(this._recent)
```
jiekang: `combineLatest` will emit an event when either of the observables emit, when `spaceDeleted` is broadcast or when the recent observable emits. In this situation we really only want to do something when there is a `spaceDeleted` broadcast so this can probably be refactored to not run anything when the recent observable emits.

src/app/shared/spaces.service.ts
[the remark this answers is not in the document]
andrewazores: Agreed, this `publish` looks wrong and `ReplaySubject` (or something similar like `shareReplay`) looks like the right tool for the desired behaviour here

src/app/shared/spaces.service.ts
andrewazores: I think the `combineLatest` should be replaced by a `broadcaster.on<Space>('spaceDeleted').map()` call, where the map function (flat)maps the single deleted space onto `this._recent.first()`, which in turn is mapped and transformed so as to remove the deleted space. Essentially the logic currently in the subscriber would mostly be moved into the mapping function(s), leaving the subscriber with the sole responsibility of saving the result.

src/app/shared/spaces.service.ts
andrewazores: Similar to the `spaceDeleted` broadcast comment below, this `combineLatest` looks like it should really be a mapping, with the subscriber logic moved up into the mapping and the subscriber responsible just for saving.

src/app/shared/spaces.service.ts
```diff
@@ -1,72 +1,80 @@
...
-    // Create the recent context list
+      this.contexts.current.map(val => val.space),
+      this.broadcaster.on<Space>('spaceUpdated')
+    );
+  }
+
+  private initRecent(): void {
+    // load existing recent spaces from profile.store.recentSpaces
+    this._recent = this.loadRecent().publishReplay(1);
+    this._recent.connect();
```
andrewazores: I'm not seeing the need for this to be a `ConnectableObservable`. It's connected immediately after instantiation anyway, during the service's constructor. Does this have any practical effect?

src/app/shared/spaces.service.ts
```diff
@@ -1,72 +1,92 @@
...
-          );
-        }
+          return recentSpaces;
+        });
       })
-      .do(val => {
-        this.saveRecent(val);
+      .subscribe((recentSpaces: Space[]) => {
+        if (this.recentChanged) {
+          this.saveRecent(recentSpaces);
```
jiekang: These should take the first of `_recent` and emit a new item to `_recent` via `_recent.next(item)`, no?

aptmac: Yeah that sounds right I think, so it takes the first when subscribing [0] and if need be it emits the new spaces [1]

[0] https://github.com/fabric8-ui/fabric8-ui/pull/3211/files#diff-103682c5eace85417d403c7c7643dcb5R42
[1] https://github.com/fabric8-ui/fabric8-ui/pull/3211/files#diff-103682c5eace85417d403c7c7643dcb5R69

jiekang: Hum, the branch I checked out this morning didn't have these... Anyways, yes it looks good.

What's the purpose of the `recentChanged` boolean over updating immediately?

aptmac: I had just recent rebased with that changed around the same time as your comment.

The boolean usage is to prevent an update if `changedSpace` is already at the front of `recentSpaces`, I think it's the only use case where we don't have to update.

aptmac: It was also to separate the action of manipulating the list and saving the list; the flatMap returns the formatted list of spaces that can be saved when subscribed to afterwards. I can remove the boolean though and move the saving.

jiekang: Oh, that's valid; I just wanted to know the reason. Sorry for making it sound like I wanted it to be changed. It's up to you, whichever you think is better here works.

src/app/shared/spaces.service.ts
```diff
@@ -1,72 +1,84 @@
...
-            SpacesService.RECENT_SPACE_LENGTH,
-            val.length - SpacesService.RECENT_SPACE_LENGTH
-          );
-        }
+          return recentSpaces;
+        });
       })
-      .do(val => {
-        this.saveRecent(val);
+      .subscribe(() => {});
```
andrewazores: This empty subscriber is pretty weird. Could you not move the `this._recent.next()` part of the `flatMap` into this subscriber? It is also weird that you have that side-effect within your mapping.

src/app/shared/spaces.service.ts
```diff
@@ -1,72 +1,84 @@
...
+            this.saveRecent(recentSpaces);
+            this._recent.next(recentSpaces);
+          }
+          return recentSpaces;
+        });
       })
-      .multicast(() => new ReplaySubject(1));
-    this._recent.connect();
-    this.loadRecent().subscribe(val => val.forEach(space => this.addRecent.next(space)));
+      .subscribe(() => {});
```
andrewazores: Same here re empty subscriber and map side effects

src/app/shared/spaces.service.ts
```diff
@@ -1,72 +1,84 @@
...
+            // otherwise, no operation required; changedSpace is already index 0
+          // if changedSpace is new to the recent spaces
+          } else {
+            recentSpaces.unshift(changedSpace);
+            // trim _recent if required to ensure it is length =< 8
+            if (recentSpaces.length > SpacesService.RECENT_SPACE_LENGTH) {
+              recentSpaces.pop();
+            }
+            this.saveRecent(recentSpaces);
+            this._recent.next(recentSpaces);
```
jiekang: It looks like this is called after every `saveRecent(...)`, maybe it can be combined as one function?

src/app/shared/spaces.service.ts
```diff
@@ -92,14 +102,26 @@ export class SpacesService implements Spaces {
...
+  /**
+   * Given the id of a space, returns the index of the space in _recent
+   */
+  findRecentIndexById(spaces: Space[], id: string): number {
+    for (let i: number = 0; i < spaces.length; i++) {
+      if (spaces[i].id === id) {
+        return i;
+      }
+    }
+    return -1;
```
andrewazores: This looks a lot like findIndex: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex

`return spaces.findIndex(space => space.id === id)`

src/app/shared/spaces.service.ts
aptmac: Sure, I could put it into `saveRecent`

src/app/shared/spaces.service.ts
aptmac: That's fair, my mistake I had interpreted Jie's review comment [0] as a request to change rather than a question when I made that change. I had the saving logic in the subscription and used a boolean flag to only save if the list was manipulated in the flatMap, I'll go back to that.

[0] https://github.com/fabric8-ui/fabric8-ui/pull/3211#discussion_r214964039

src/app/shared/spaces.service.ts
aptmac: Ah thanks for the heads up, that does look like what I want.

src/app/shared/spaces.service.ts
aptmac: It's not really the emission that I'm worried about, it's the patch request to update the profile from saveRecent that I want to eliminate. I haven't dug around enough to figure out exactly what happens, but it seems that if you send a certain number of requests in a limited amount of time, then the profile doesn't get updated correctly. i.e., the more spaces (or contexts) I visit in quick succession, the less likely the profile is able to properly be updated with the values to persist a refresh.

src/app/shared/spaces.service.ts
```diff
@@ -72,7 +72,7 @@ export class SpacesService implements Spaces {
     this.broadcaster.on<Space>('spaceDeleted')
       .flatMap((deletedSpace: Space) => {
         return this._recent.first().map((recentSpaces: Space[]) => {
-          let index: number = this.findRecentIndexById(recentSpaces, deletedSpace.id);
+          let index: number = recentSpaces.findIndex(space => space.id === deletedSpace.id);
```
andrewazores: Could you amend this commit and just add a type on the `space` param?

src/app/shared/spaces.service.ts
andrewazores: Ah right, that makes sense. I'll try to help brainstorm "better" ways to do this than `recentChanged`, but that might end up being the simplest way to do it. I would just really prefer to avoid having any kind of side effect in the mapping, even if it's as simple as setting essentially a dirty flag.

jiekang: Right now it's chaining off of the broadcast's separately and into the recent, I think if it goes from the recent and internalizes reading broadcasts inside into emitting recents, it could be cleaner. Sorry for the confusing description. I haven't thought about how to actualize it in code.

aptmac: @andrewazores with regards to removing recentChanged, what if the ReplaySubject was switched to be a BehaviorSubject instead? This way we could use a filter at the beginning of the observable chain and exit if the condition isn't satisfied.

e.g.,
```
this.broadcaster.on<Space>('spaceChanged')
  .filter((changedSpace: Space) => {
    return this._recent.getValue()[0].id !== changedSpace.id;
  })
  .flatMap((changedSpace: Space) => { ...
```
and
```
this.broadcaster.on<Space>('spaceDeleted')
  .filter((deletedSpace: Space) => {
    return this._recent.getValue().findIndex((space: Space) => space.id === deletedSpace.id) !== -1;
  }) ...
...
```

The `flatMap` wouldn't need to toggle the boolean and the subscription at the end could simply be the call to `saveRecent()`.

src/app/shared/spaces.service.ts
```diff
@@ -1,72 +1,90 @@
...
-      // The scan operator allows us to access the list of recent contexts and add ours
-      .scan((recent, s) => {
-        // First, check if this context is already in the list
-        // If it is, remove it, so we don't get duplicates
-        for (let i = recent.length - 1; i >= 0; i--) {
-          if (recent[i].id === s.id) {
-            recent.splice(i, 1);
+    // update _recent with newly seen space when spaceChanged is emitted
+    this.broadcaster.on<Space>('spaceChanged')
+      .flatMap((changedSpace: Space) => {
```
andrewazores: Please add a declaration for the return type of this lambda function

src/app/shared/spaces.service.ts
andrewazores: I had an idea thinking back on my `distinctUntilChanged` thought here again. Doesn't that still meet our needs here? ie remove `recentChanged` completely, and continue performing `saveRecent` within the subscriber, but simply insert a `distinctUntilChanged` to transform the `flatMap` result before the result is passed in to the subscriber. The way I had it written out (and thought about) before was assuming consumers might or might not care about duplicate emissions - so why don't we just treat our own subscriber as a consumer, and use provided operators to filter the duplicates?

src/app/shared/spaces.service.ts
```diff
@@ -1,72 +1,90 @@
...
-        this.saveRecent(val);
+      .subscribe((recentSpaces: Space[]) => {
+        if (this.recentChanged) {
+          this.saveRecent(recentSpaces);
+        }
+      });
+
+    // if a space is deleted, check to see if it should be removed from _recent
+    this.broadcaster.on<Space>('spaceDeleted')
+      .flatMap((deletedSpace: Space) => {
```
andrewazores: Lambda return type

src/app/shared/spaces.service.ts
```diff
@@ -1,72 +1,90 @@
...
+            recentSpaces.splice(index, 1);
+            this.recentChanged = true;
+          }
+          return recentSpaces;
+        });
       })
-      .multicast(() => new ReplaySubject(1));
-    this._recent.connect();
-    this.loadRecent().subscribe(val => val.forEach(space => this.addRecent.next(space)));
+      .subscribe((recentSpaces: Space[]) => {
```
andrewazores: Lambda return type (void)

src/app/shared/spaces.service.ts
andrewazores: Ha, we were writing our comments at the same time. Yea, using a `filter` instead of `distinctUntilChanged` could work too. Whichever comes out cleaner, really.

src/app/shared/spaces.service.ts
```diff
@@ -10,15 +10,15 @@ export class SpacesService implements Spaces {
...
-  private recentChanged: boolean = false;
+  private _recent: BehaviorSubject<Space[]>;
 
   constructor(
     private contexts: Contexts,
     private spaceService: SpaceService,
     private broadcaster: Broadcaster,
     private profileService: ProfileService
   ) {
+    this._recent = new BehaviorSubject<Space[]>([] as Space[]);
```
andrewazores: Is the `as Space[]` necessary? Seems like `[]` alone should be sufficient for the compiler to infer the type...

src/app/shared/spaces.service.ts
```diff
@@ -32,22 +32,24 @@ export class SpacesService implements Spaces {
...
       this._recent.next(spaces);
     });
 
     // update _recent with newly seen space when spaceChanged is emitted
     this.broadcaster.on<Space>('spaceChanged')
-      .flatMap((changedSpace: Space) => {
-        return this._recent.first().map((recentSpaces: Space[]) => {
-          let index: number = recentSpaces.findIndex((space: Space) => space.id === changedSpace.id);
+      .filter((changedSpace: Space): boolean => {
+        return this._recent.getValue()[0].id !== changedSpace.id;
```
andrewazores: Eek, using `getValue` isn't so nice. Is there some other way? Perhaps something like `on('spaceChanged').merge(this._recent.first())`?

src/app/shared/spaces.service.ts
aptmac: Yeah it should probably be enough, just figured I would throw them in wherever I could.

src/app/shared/spaces.service.ts
aptmac: Sure I can play around with that. Is there any reason in particular getting the value of a BehaviorSubject isn't safe? As far as I can tell BehaviorSubjects only store the last value emitted, so the return from `getValue` should be the current list of spaces.

andrewazores: It's just not very idiomatic and doesn't read like Rx. It also makes it significantly more involved to maintain if in the future somebody else comes along and has a reason to change the `BehaviorSubject` to some other kind of `Subject`. IMO declaring the type of `this._recent` as simply `Subject<Space[]>` rather than any *specific* kind of `Subject` would make sense too.

src/app/shared/spaces.service.ts
andrewazores: I don't think throwing in type casts wherever they fit is a good habit, frankly...

src/app/shared/spaces.service.ts
aptmac: That sounds fair to me. I played around with this and had difficulty using merge because the broadcaster emits a space and _recent is `Space[]`, so after merging the type was `Space | Space[]`. 

I did however stumble upon `withLatestFrom`, which looks pretty good for this use case I think. It's similar to `combineLatest`, but instead of emitting whenever either of the observables emits, it only does so for the primary observable.

After the `withLatestFrom`, we have a combination of the changed space and list of recent spaces, which makes it pretty easy for value comparison. I added a `.map()` call prior to the `.filter()` with the purpose of including the (potential) index of the space in recent, so that it could be passed around with the space information and not have to be re-calculated in subsequent functions.

andrewazores: Nice, I like that a lot. Looks good to me now.

PR #3221, opened 2018-08-02
feat(http-client): upgrade GettingStarted and FeatureAcknowledgement …

andrewazores:
…to HttpClient

Upgrade GettingStartedService and subclass FeatureAcknowledgementService to use newer HttpClient in
anticipation of Angular upgrade. Tests for components that consume these services while also
providing them at the component level are also updated, overriding the component-level provides for
these services, so that the indirect Http/HttpClient module dependencies can be dropped from those tests
altogether.

related to https://openshift.io/openshiftio/Openshift_io/plan/detail/348


src/app/profile/tenant/tenant.component.spec.ts
```diff
@@ -52,9 +54,16 @@ describe('TenantComponent', () => {
...
+        { provide: ProviderService, useValue: mockProviderService },
+        { provide: GitHubService, useValue: mockGitHubService },
         { provide: WIT_API_URL, useValue: 'http://example.com'}
       ]
     });
+    TestBed.overrideProvider(AuthenticationService, { useValue: mockAuthenticationService });
+    TestBed.overrideProvider(GettingStartedService, { useValue: mockGettingStartedService });
+    TestBed.overrideProvider(TenantService, { useValue: mockTenantService });
+    TestBed.overrideProvider(ProviderService, { useValue: mockProviderService });
+    TestBed.overrideProvider(GitHubService, { useValue: mockGitHubService });
```
christianvogt: Why is this override here and also configured in the above call to `configureTestingModule`?

src/app/profile/update/update.component.spec.ts
```diff
@@ -66,6 +65,12 @@ describe('UpdateComponent', () => {
       ],
       schemas: [NO_ERRORS_SCHEMA]
     });
+    TestBed.overrideProvider(AuthenticationService, { useValue: mockAuthenticationService });
+    TestBed.overrideProvider(GettingStartedService, { useValue: mockGettingStartedService });
+    TestBed.overrideProvider(GitHubService, { useValue: mockGitHubService });
+    TestBed.overrideProvider(ProviderService, { useValue: mockProviderService });
+    TestBed.overrideProvider(TenantService, { useValue: mockTenantService });
+    TestBed.overrideProvider(UserService, { useValue: mockUserService });
```
christianvogt: Again, why the duplicate providers?

src/app/getting-started/services/getting-started.service.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+          useFactory: (): jasmine.SpyObj<AuthenticationService> => {
+            const svc: jasmine.SpyObj<AuthenticationService> = createMock(AuthenticationService);
+            svc.getToken.and.returnValue('mock-auth-token');
+            return svc;
+          }
+        },
+        {
+          provide: UserService,
+          useFactory: (): jasmine.SpyObj<UserService> => {
+            const svc: jasmine.SpyObj<UserService> = createMock(UserService);
```
christianvogt: Simplify to a single return statement.

src/app/getting-started/services/getting-started.service.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+            return svc;
+          }
+        },
+        {
+          provide: Logger,
+          useFactory: (): jasmine.SpyObj<Logger> => {
+            const logger: jasmine.SpyObj<Logger> = createMock(Logger);
```
christianvogt: Simplify to a single return statement.

src/app/getting-started/services/getting-started.service.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+  });
+
+  describe('HTTP Tests', () => {
+
+    afterEach(function(this: TestContext): void {
+      this.controller.verify();
+    });
+
+    describe('#getExtProfile', () => {
+      it('should send a GET', function(this: TestContext, done: DoneFn): void {
```
christianvogt: Can this be simplified to just `function(done: DoneFn): void {` ?

andrewazores: If I extract the references contained in the `TestContext` out and just hold them in an enclosing scope or something, sure, but it's as it is intentionally.

christianvogt: Ok. I didn't look closely enough and with the param named `this` it was confusing.

src/app/feature-flag/service/feature-acknowledgement.service.spec.ts
```diff
@@ -1,5 +1,5 @@
+import { HttpClientTestingModule } from '@angular/common/http/testing';
```
joshuawilson: If this is not used, please remove it instead of upgrading it.

src/app/getting-started/services/getting-started.service.ts
```diff
@@ -67,13 +75,9 @@ export class GettingStartedService implements OnDestroy {
...
-      .get(url, { headers: this.headers })
-      .map(response => {
-        return response.json().data as ExtUser;
-      })
-      .catch((error) => {
-        return this.handleError(error);
-      });
+      .get<ExtUserResponse>(url, { headers: this.headers })
+      .map(response => response.data)
+      .catch((error) => this.handleError(error));
```
joshuawilson: please consider adding types for `response` and `error`

src/app/feature-flag/service/feature-acknowledgement.service.ts
```diff
@@ -1,5 +1,5 @@
+import { HttpClient } from '@angular/common/http';
```
joshuawilson: If this is not used, please remove it instead of upgrading it.

joshuawilson:
@andrewazores and @rohitkrai03 please deconflict your PRs. You are both updating the profile page.

rohitkrai03:
@joshuawilson Andrew had some dependency on profile section components and to fix tests for those he had to make changes to the same components as my PR. We can rebase once any one of our PRs get merged.

joshuawilson:
@andrewazores I'm going to merge this because it is needed for another of your PRs. Please take a look at the comments and update the code when you get a chance.

src/app/getting-started/services/getting-started.service.ts
andrewazores: PR opened here: https://github.com/fabric8-ui/fabric8-ui/pull/3263

src/app/feature-flag/service/feature-acknowledgement.service.ts
andrewazores: This and the equivalent in the associated spec file are actually used and required to upgrade, although indirectly - the HttpClient instance is passed up through `super()` in the service constructor, and then directly used by the superclass.

andrewazores: Although, perhaps it makes sense to refactor this for injected composition rather than inheritance. It'll loosen the coupling. Would you like me to take a look at that?

PR #3234, opened 2018-08-03
Convert landing page's use of http

chrislessard:
Hi, 

This patch removes the use of Http (which needed to be converted [1]) from the landing page, as it was dead code. It also adds a basic set of unit tests to the landing page component.

[1] https://openshift.io/openshiftio/Openshift_io/plan/detail/348


src/app/landing-page/landing-page.component.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+
+@Component({
+  template: '<alm-landing-page></alm-landing-page>'
+})
+class HostComponent { }
+
+describe('LandingPageComponent', () => {
+  type Context = TestContext<LandingPageComponent, HostComponent>;
+
+  let mockAuthenticationService: AuthenticationService;
```
andrewazores: This looks unused, is it?

chrislessard: I used it when I declare the AuthenticationService provider a bit later on, I believe this is necessary ? Line 40.

andrewazores: You redeclare a variable by the same name on line 41, shadowing this first declaration, and this new variable is returned at the end of its scope (the factory anonymous function). Later when you access this instance on line 51 you do so via `TestBed.get`. So there are no apparent accessors of this declaration on line 26.

PR #3246, opened 2018-08-10
fix(launcher): update ngx-launcher to 1.0.16

vikram-raj:
This version contains these fixes.
1. openshiftio/openshift.io#3443
2. Create app flow of this issue https://github.com/openshiftio/openshift.io/issues/4140
3. openshiftio/openshift.io#3927
4. https://github.com/fabric8-launcher/ngx-launcher/issues/349 and https://github.com/openshiftio/openshift.io/issues/3997

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -714,6 +714,16 @@
         "@types/uglify-js": "3.0.3"
       }
     },
+    "JSONStream": {
+      "version": "1.3.4",
+      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.4.tgz",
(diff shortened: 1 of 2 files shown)


bartoszmajsak:
@joshuawilson how do we test such updates?

bartoszmajsak:
@vikram-raj as there is a newer version out can we update this PR to have that shipped instead.

@joshuawilson how we can make merging PRs smoother?

joshuawilson:
In what way @bartoszmajsak?

@vikram-raj please rebase

bartoszmajsak:
@joshuawilson In a way that library update does not hang 5d as a PR :) But it is general observation of PR-to-master time. We should think about how to make that easier (boils down to ensuring we don't break things... but anyway).

vikram-raj:
@bartoszmajsak @joshuawilson  I updated this PR with the latest version of ```ngx-launcher```.

joshuawilson:
The `continuous-integration/jenkins/pr-merge` build step is not required. I've just disabled that webhook, so it should be gone in future PRs.

PR #3259, opened 2018-08-17
feat(redirect-status): generic landing page for various redirect status

rohitkrai03:
Fixes https://github.com/openshiftio/openshift.io/issues/4102
Fixes https://github.com/fabric8-services/fabric8-auth/issues/566

Related UX issue - https://github.com/openshiftio/openshift.io/issues/3929

Currently we have status pages for redirects coming from verification of emails.
- Success - https://prod-preview.openshift.io/_verifyEmail?verified=true
- Error - https://prod-preview.openshift.io/_verifyEmail?verified=false&error=some_error

#### Now we needed a similar landing page for invitations API. Based on the comments by @catrobson  https://github.com/openshiftio/openshift.io/issues/3929#issuecomment-405682187 this PR introduces a new generic component for all redirects which can be configured for specific redirect scenarios with custom messages or CTA links. URL pattern for redirects would be like - 
- `https://openshift.io/_redirects/_verifyEmail?status=success`
- `https://openshift.io/_redirects/_verifyEmail?status=fail&error=some_error_msg`
- `https://openshift.io/_redirects/_acceptInvitation?status=fail&error=some_error_msg`

#### Screenshots

screen shot 2018-08-20 at 6 41 15 pm

screen shot 2018-08-20 at 6 41 55 pm


jiekang:
Could you recommend a good way for me to test the features being added? Also it looks like `src/app/layout/redirect-status/redirect-status.component.less` is empty.

rohitkrai03:
@jiekang You can test it locally by just going to the below URLs -

- `https://openshift.io/_redirects/_verifyEmail?status=success`
- `https://openshift.io/_redirects/_verifyEmail?status=fail&error=some_error_msg`

is that what you meant by good way of testing the feature?

Also, deleted the unused less file.

jiekang:
Okay testing via visiting that url is okay. I was expecting some flow around it like I need to have an account with an unverified e-mail and then getting some e-mail verification and whatnot. But that's hopefully tested via E2E tests or something?

rohitkrai03:
@jiekang It cannot be tested via the end to end email verification workflow since the backend still redirects to the old verification page. Once this gets merged, we'll update the backend with new URL and remove the old code from frontend.

src/app/layout/redirect-status/redirect-status.component.ts
```diff
@@ -0,0 +1,44 @@
+import { Component, OnInit } from '@angular/core';
+import { ActivatedRoute } from '@angular/router';
+
+@Component({
+  selector: 'f8-redirect-status',
+  templateUrl: 'redirect-status.component.html',
+  styleUrls: ['./redirect-status.component.less']
```
jiekang: This reference should also be removed.

jiekang:
Okay, makes sense. Thanks!

jiekang:
@rohitkrai03 Can you provide screenshots of the success and error and ask UX for review? Also in the error state, an error notification pop-up with the message is also shown. I think it's redundant, but I'm not sure where it's coming from.

src/app/layout/redirect-status/redirect-status.component.ts
rohitkrai03: Done

rohitkrai03:
@jiekang That error notification comes from route guard which basically sends a notification anytime a URL param with name `error` comes in.

jiekang:
Hmm okay; I guess that's a separate issue. Do you happen to know what other situations a URL with param `error` occurs? I'd be inclined to remove that functionality.

src/app/layout/redirect-status/status/status.component.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+      expect(failImgEl.nativeElement).toBeFalsy();
+      expect(successImgEl.nativeElement.getAttribute('src'))
+        .toEqual('../../../../assets/images/trophy.png');
+
+      expect(ctaEl.nativeElement.getAttribute('href'))
+        .toEqual('/' + successData.ctaLink);
+    });
+  });
+
+  it('Should have fail image, msg and cta_link set', () => {
```
jiekang: Prefer lower case 'should' to match other specs. Also occurs in the other `it` in this file.

src/app/layout/redirect-status/status/status.component.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    ctaEl = fixture.debugElement.query(By.css('.cta-button'));
+    msgEl = fixture.debugElement.query(By.css('.primary-msg'));
+    secMsgEl = fixture.debugElement.query(By.css('.secondary-msg'));
+  });
+
+  it('Should have success image, msg and cta_link set', () => {
+    component.status = 'success';
+    component.data = successData;
+    fixture.detectChanges();
+    fixture.whenStable().then(() => {
```
jiekang: I think this expects to be run in an async block, similar to the test in `src/app/layout/redirect-status/redirect-status.spec.ts`. This applies to the other test in this spec.

src/app/layout/redirect-status/status/status.component.spec.ts
```diff
@@ -0,0 +1,86 @@
+import { DebugElement, NO_ERRORS_SCHEMA } from '@angular/core';
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { RouterTestingModule } from '@angular/router/testing';
+import { StatusComponent } from './status.component';
+
+describe('Component: Status', () => {
```
jiekang: Prefer `StatusComponent` to follow convention in other tests.

src/app/layout/redirect-status/redirect-status.component.ts
```diff
@@ -0,0 +1,43 @@
...
+        message: '',
+        secMessage: 'It appears there is a problem with validating your e-mail. You can reset your e-mail on your Profile Page',
+        ctaLink: '_profile',
+        ctaLabel: 'profile'
+      }
+    }
+  };
+
+  private redirectStatus: string;
+  private redirectData: object;
```
jiekang: Please create a type for this instead of using 'object'. This will also help when it's moved to ngx-widgets as users will know what fields to expect.

jiekang: Er more specifically, when the `status.component`, which expects data of some format, is moved.

src/app/layout/redirect-status/redirect-status.component.html
```diff
@@ -0,0 +1,5 @@
+<div  class="container">
```
andrewazores: There are two spaces between `div` and `class` here

src/app/layout/redirect-status/redirect-status.component.ts
```diff
@@ -0,0 +1,45 @@
...
+  private redirectStatus: string;
+  private redirectData: RedirectData;
+
+  constructor(private route: ActivatedRoute) {}
+
+  ngOnInit(): void {
+    const redirectType = this.route.snapshot.params['redirectType'];
+    const queryParams = this.route.snapshot.queryParams;
+    this.redirectStatus = queryParams['status'];
+    if (this.redirects.hasOwnProperty(redirectType) && this.redirectStatus) {
```
andrewazores: Could `redirects` be a better type than an `object` for this purpose? `Map` maybe?

src/app/layout/redirect-status/redirect-status.spec.ts
```diff
@@ -0,0 +1,48 @@
+import { NO_ERRORS_SCHEMA } from '@angular/core';
+import { async, TestBed } from '@angular/core/testing';
+import { ActivatedRoute, Router } from '@angular/router';
+import { RouterTestingModule } from '@angular/router/testing';
+
+import { RedirectStatusComponent } from './redirect-status.component';
+
+describe('Redirect Status Component', () => {
```
andrewazores: This file should be named `redirect-status.component.spec.ts` - it's missing the `component` part

src/app/layout/redirect-status/redirect-status.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+      schemas: [NO_ERRORS_SCHEMA]
+    });
+
+    fixture = TestBed.createComponent(RedirectStatusComponent);
+  });
+
+  it('should verify redirect status and error message is set through activated route', async(() => {
+    let comp = fixture.componentInstance;
+    fixture.detectChanges();
+    fixture.whenStable().then(() => {
```
andrewazores: Is the `async` and `whenStable` really necessary here? Also, this spec could be cleaned up to use the `TestContext` and `initContext` testing utils we have in the project, which might reduce a bit of boilerplate but more importantly make it the same consistent boilerplate as we see in many other places. IMO it's more readable and maintainable that way. I think it'll also help in removing the async/whenStable stuff since it has a little bit of included tooling around the component lifecycle.

src/app/layout/redirect-status/status/status.component.html
```diff
@@ -0,0 +1,29 @@
+<div class="blank-slate-pf">
+  <div class="rh-branding-logo">
+    <img class="width-100"
+      src="../../../../assets/images/Logotype_RH_OpenShift-io_RGB_RedGray.png">
```
andrewazores: I seem to recall that there are aliased paths for things like this, but I don't know if Webpack is configured so that those paths are usable from within HTML. Could you double check that, so these paths can look like `~assets/images/foo` and avoid all the relative pathing `../../` parts?

src/app/layout/redirect-status/status/status.component.html
```diff
@@ -0,0 +1,29 @@
...
+  </div>
+  <div>
+    <img
+      *ngIf="status === 'success'"
```
andrewazores: Maybe `ngSwitch` here instead of `ngIf`?

src/app/layout/redirect-status/redirect-status.spec.ts
```diff
@@ -0,0 +1,48 @@
+import { NO_ERRORS_SCHEMA } from '@angular/core';
+import { async, TestBed } from '@angular/core/testing';
+import { ActivatedRoute, Router } from '@angular/router';
+import { RouterTestingModule } from '@angular/router/testing';
+
+import { RedirectStatusComponent } from './redirect-status.component';
+
+describe('Redirect Status Component', () => {
+  let fixture;
+  let routeMock: any;
```
andrewazores: Is there a type that can be specified here rather than `any`?

src/app/layout/redirect-status/redirect-status.spec.ts
```diff
@@ -0,0 +1,48 @@
+import { NO_ERRORS_SCHEMA } from '@angular/core';
+import { async, TestBed } from '@angular/core/testing';
+import { ActivatedRoute, Router } from '@angular/router';
+import { RouterTestingModule } from '@angular/router/testing';
+
+import { RedirectStatusComponent } from './redirect-status.component';
+
+describe('Redirect Status Component', () => {
+  let fixture;
```
andrewazores: Needs type annotation

src/app/layout/redirect-status/status/status.component.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  let component: StatusComponent;
+  let fixture: ComponentFixture<StatusComponent>;
+  let successImgEl: DebugElement;
+  let failImgEl: DebugElement;
+  let ctaEl: DebugElement;
+  let msgEl: DebugElement;
+  let secMsgEl: DebugElement;
+
+  const successData = {
```
andrewazores: Is there a type that can be added here?

src/app/layout/redirect-status/status/status.component.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    ctaLink: '_profile',
+    ctaLabel: 'profile'
+  };
+
+  beforeEach(() => {
+
+    TestBed.configureTestingModule({
+      imports: [RouterTestingModule.withRoutes([])],
+      declarations: [StatusComponent],
+      schemas: [ NO_ERRORS_SCHEMA ]
```
andrewazores: Inconsistent use of `[ FOO ]` vs `[FOO]`

src/app/layout/redirect-status/status/status.component.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    ctaEl = fixture.debugElement.query(By.css('.cta-button'));
+    msgEl = fixture.debugElement.query(By.css('.primary-msg'));
+    secMsgEl = fixture.debugElement.query(By.css('.secondary-msg'));
+  });
+
+  it('should have success image, msg and cta_link set', () => {
+    component.status = 'success';
+    component.data = successData;
+    fixture.detectChanges();
+    fixture.whenStable().then(async(() => {
```
andrewazores: Same as in the other spec - is the `whenStable/async` really needed here? If so, could you explain why? It looks like all dumb, presentational, synchronous stuff in here. Does it have to do with the `ngIf`s or image loading or something?

src/app/layout/redirect-status/status/status.component.ts
```diff
@@ -0,0 +1,15 @@
...
+  selector: 'f8-status',
+  templateUrl: 'status.component.html',
+  styleUrls: ['./status.component.less']
+})
+export class StatusComponent {
+
+  @Input() status: 'success' | 'fail';
+  @Input() data: RedirectData;
+
+  constructor() {}
```
andrewazores: Empty constructor can be omitted

src/app/models/redirect-data.ts
```diff
@@ -0,0 +1,6 @@
+export class RedirectData {
+  message: string;
+  secMessage: string;
+  ctaLink: string;
+  ctaLabel: string;
```
andrewazores: Could these properties be renamed so that they're a little more self-explanatory?

src/app/layout/redirect-status/redirect-status.component.ts
rohitkrai03: We could do something like 

```js
private redirects: Map<string, Map<string, RedirectData>> = new Map([
    [
      '_verifyEmail', new Map([
        ['success', {
          message: 'Your e-mail has been confirmed.',
          secMessage: 'Thank you for validating your e-mail address. You can now continue to use Openshift.io',
          ctaLink: '_home',
          ctaLabel: 'home dashboard'
        }],
        ['fail', {
          message: '',
          secMessage: 'It appears there is a problem with validating your e-mail. You can reset your e-mail on your Profile Page',
          ctaLink: '_profile',
          ctaLabel: 'profile'
        }]
      ])
    ]
  ]);

```

But it increases the complexity a lot. 

A simpler version maybe with just one `Map` - 

```js
  private redirects: Map<string, any> = new Map([
    [
      '_verifyEmail', {
        success: {
          message: 'Your e-mail has been confirmed.',
          secMessage: 'Thank you for validating your e-mail address. You can now continue to use Openshift.io',
          ctaLink: '_home',
          ctaLabel: 'home dashboard'
        },
        fail: {
          message: '',
          secMessage: 'It appears there is a problem with validating your e-mail. You can reset your e-mail on your Profile Page',
          ctaLink: '_profile',
          ctaLabel: 'profile'
        }
      }
    ]
  ]);

```

andrewazores: The `Map<string, any>` looks better to me, although I'd like to see an actual type used rather than `any`.

aptmac:
Small nit but the commit messages should be written in the present tense, also your first commit message has typos: "genric" and "varioud"

src/app/layout/redirect-status/status/status.component.spec.ts
```diff
@@ -4,8 +4,10 @@ import { By } from '@angular/platform-browser';
 import { RouterTestingModule } from '@angular/router/testing';
 import { StatusComponent } from './status.component';
 
+import { RedirectData } from '../../../models/redirect-data';
 
-describe('StatusComponent', () => {
+
+fdescribe('StatusComponent', () => {
```
andrewazores: `fdescribe`

src/app/layout/redirect-status/status/status.component.spec.ts
rohitkrai03: Probably because of image loading and ngSwitch tests fail without using `whenStable/async`.

src/app/models/redirect-data.ts
rohitkrai03: Do you have any suggestion?

src/app/layout/redirect-status/status/status.component.html
rohitkrai03: I can't seem to find anywhere in `fabric8-ui` that uses links like that. Everywhere it seems to be relative path so i don't think we have webpack setup for that. Might be worth looking into it though.

rohitkrai03:
@aptmac Updated commit messages.

src/app/layout/redirect-status/status/status.component.html
andrewazores: There's some configuration for testing utils at least. We definitely have specs that do `import { createMock } from 'testing/mock'` for example. If there's nothing similar for assets then I'd like to see it added, but that's a separate concern from this PR.

src/app/layout/redirect-status/redirect-status.component.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { RedirectData } from '../../models/redirect-data';
+
+@Component({
+  selector: 'f8-redirect-status',
+  templateUrl: 'redirect-status.component.html'
+})
+export class RedirectStatusComponent implements OnInit {
+
+  private redirects: Map<string, any> = new Map([
```
andrewazores: I'd still like to see the `any` here replaced with a type definition describing the shape of the success/fail RedirectData structure within.

rohitkrai03: Updated type. Is it better now?

jiekang:
@rohitkrai03 let me know when conflicts are resolved and I'll put this on queue for merging!

src/app/app-routing.module.ts
```diff
@@ -44,6 +44,15 @@ export const routes: Routes = [
       title: 'Verify'
     }
   },
+
```
jiekang: tslint is failing here due to trailing whitespace

rohitkrai03: @jiekang Fixed. Used Github UI for resolving conflicts which is why this typo happened.

jiekang:
@catrobson Could the UX take a look at the designs posted at the head of this PR? Are they okay?

rohitkrai03:
@jiekang From design side, nothing changed here.

jiekang:
@rohitkrai03 I meant to ask here actually: What's the expected result if a user visits this when not logged in? How does this work? I.e. for e-mail verification: User receives e-mail with link. User clicks link. If user is logged in: ? If user is not logged in: ?

rohitkrai03:
@jiekang The page will work as expected irrespective of user being logged in or not since the actual link will be to the backend API which will then redirect to this page with status of the action.

jiekang:
I think the page has a link to 'return to home'. Would that still make sense when not logged in? Maybe it should be 'you can now log-in' and direct to the login flow if the user is not logged in.

rohitkrai03:
@jiekang I think going to home will automatically trigger the login process.

jiekang:
Okay, sounds good, thanks!

PR #3262, opened 2018-08-20
refactor(shared-services): upgrade to HttpClient from http

rohitkrai03:
- Upgraded to `HttpClient` in shared services.


src/app/shared/account/provider.service.spec.ts
```diff
@@ -1,71 +1,46 @@
...
-            },
-            {
-              provide: Broadcaster,
-              useValue: BroadcasterTestProvider.broadcaster
-            },
-            Logger
-          ]
-        });
+      const mockAuthenticationService: jasmine.SpyObj<AuthenticationService> = createMock(AuthenticationService);
+      mockAuthenticationService.getToken.and.returnValue('eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiY2xpZW50X3Nlc3Npb24iOiJURVNUU0VTU0lPTiIsInNlc3Npb25fc3RhdGUiOiJURVNUU0VTU0lPTlNUQVRFIiwiYWRtaW4iOnRydWUsImp0aSI6ImY5NWQyNmZlLWFkYzgtNDc0YS05MTk0LWRjM2E0YWFiYzUwMiIsImlhdCI6MTUxMDU3MTMxOSwiZXhwIjoxNTEwNTgwODI3fQ.l0m6EFvk5jbND3VOXL3gTkzTz0lYQtPtXS_6C24kPQk');
```
andrewazores: Can we replace this token string with something a little more obviously mocked, like `mock-auth-token`, while we're here?

src/app/shared/account/provider.service.spec.ts
```diff
@@ -1,71 +1,46 @@
...
+        providers: [
+          { provide: Logger, useValue: mockLogger },
+          { provide: AuthenticationService, useValue: mockAuthenticationService },
+          { provide: AUTH_API_URL, useValue: 'https://auth.fabric8.io/api/' },
+          { provide: Broadcaster, useValue: BroadcasterTestProvider.broadcaster },
+          ProviderService
+        ]
       });
+      service = TestBed.get(ProviderService);
+      controller = TestBed.get(HttpTestingController);
```
andrewazores: The `controller` doesn't look used anywhere. Can a `controller.expectOne('')` and `controller.verify()` be added in the tests here? If there's nowhere where this makes sense to do, then this would seem to imply that the spec doesn't need to use the HttpClientTestingModule.

src/app/shared/account/provider.service.spec.ts
rohitkrai03: This token actually contains information which is decoded using `jwt-decode` to form the URL.

rohitkrai03: Done.

src/app/shared/account/provider.service.spec.ts
rohitkrai03: Added a few more tests for http requests.

PR #3276, opened 2018-08-22
fix(spaces): add loading widget to profile spaces component

aptmac:
This PR addresses the loading state of the profile spaces component, and it's current lack of the loading widget while retrieving spaces. Currently, an empty state screen appears initially, and then gets updated once the list of spaces returns. This PR uses the loading widget similar to other components in OSIO to display a spinner while the list of spaces is being retrieved.

Before:
before

After:
after

### DIFF
--- a/src/app/profile/overview/spaces/spaces.component.html
+++ b/src/app/profile/overview/spaces/spaces.component.html
@@ -8,7 +8,10 @@
     </div>
   </div>
   <div class="card-pf-body f8-card-body">
-    <div *ngIf="spaces.length === 0; then showBlankState else showSpaces"></div>
+    <fabric8-loading-widget message="Please wait while we load the spaces"
+                            *ngIf="loading"></fabric8-loading-widget>
+    <div *ngIf="spaces.length > 0 && !loading then showSpaces"></div>
+    <div *ngIf="spaces.length === 0 && !loading then showBlankState"></div>
     <ng-template #showBlankState>
       <div class="f8-blank-slate-card">
         <div class="f8-blank-slate-icon">
--- a/src/app/profile/overview/spaces/spaces.component.spec.ts
+++ b/src/app/profile/overview/spaces/spaces.component.spec.ts
@@ -1,13 +1,12 @@
-import { DebugNode, NO_ERRORS_SCHEMA } from '@angular/core';
+import { DebugNode, ErrorHandler, NO_ERRORS_SCHEMA } from '@angular/core';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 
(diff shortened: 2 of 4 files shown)


src/app/profile/overview/spaces/spaces.component.html
```diff
@@ -8,7 +8,10 @@ <h2 class="card-pf-title">
     </div>
   </div>
   <div class="card-pf-body f8-card-body">
-    <div *ngIf="spaces.length === 0; then showBlankState else showSpaces"></div>
+    <fabric8-loading-widget message="Please wait while we load the spaces"
+                            *ngIf="loading === true"></fabric8-loading-widget>
```
andrewazores: `loading === true` and `loading === false` can be simplified to `loading` and `!loading` respectively. Or, `*ngIf="loading; else #loaded"` or something similar.

src/app/profile/overview/spaces/spaces.component.spec.ts
```diff
@@ -65,6 +60,12 @@ describe('SpacesComponent', () => {
   });
 
   describe('#initSpaces', () => {
+
+    // ensure the component is not left in a loading state
+    afterEach(() => {
+      expect(component.loading).toEqual(false);
+    });
```
andrewazores: Feels weird to put a test assertion in an afterEach like that, but I guess the intent makes sense. What happens if this fails though, since the `expect` isn't within an `it`? Does it accurately report which `it` block caused the failure?

aptmac: Yeah, it displays the test that it failed on, and what value it was expecting. This afterEach acts within the `describe('#initSpaces', ... )` block it is in, so it won't apply to every test in the spec file. For example, I slipped a `component.loading = true` into the last test of that block to force a failure, and the output was:

```
FAILED TESTS:
  SpacesComponent
    #initSpaces
      ✖ should log an error if the context or context.user is empty
        PhantomJS 2.1.1 (Linux 0.0.0)
      Expected true to equal false.
```

andrewazores: Cool, good to know. If you add an `it` inside the `afterEach`/around the `expect`, maybe the error reporting would be even clearer?

aptmac: Neat, if you put code inside of an `it` inside of an `afterEach` it doesn't seem to run. The `expect` wasn't hit, along with a bunch of logs that I threw in there.

andrewazores: Interesting. It's fine as-is anyway, I was just curious as I don't tend to use `afterEach`.

PR #3277, opened 2018-08-23
feat(notifications-settings): add internal feature flag for notifications tab

rohitkrai03:
Fixes https://github.com/openshiftio/openshift.io/issues/4235
Fixes https://github.com/fabric8-services/fabric8-auth/issues/619

- Adds feature flag with `internal` level to notifications tab under settings section.

### DIFF
--- a/src/app/profile/settings/settings-routing.module.ts
+++ b/src/app/profile/settings/settings-routing.module.ts
@@ -3,6 +3,8 @@ import { RouterModule, Routes } from '@angular/router';
 
 import { SettingsComponent } from './settings.component';
 
+import { FeatureFlagResolver } from 'ngx-feature-flag';
+
 const routes: Routes = [
   {
     path: '',
@@ -18,7 +20,14 @@ const routes: Routes = [
       },
       {
         path: 'notifications',
-        loadChildren: './notifications/notifications.module#NotificationsModule'
+        loadChildren: './notifications/notifications.module#NotificationsModule',
+        resolve: {
+          featureFlagConfig: FeatureFlagResolver
+        },
+        data: {
+          title: 'Notifications',
+          featureName: 'ProfileSettingsNotifications'
+        }
       },
       {
         path: 'resources',
--- a/src/app/profile/settings/settings-routing.module.ts
+++ b/src/app/profile/settings/settings-routing.module.ts
@@ -3,8 +3,6 @@ import { RouterModule, Routes } from '@angular/router';
 
 import { SettingsComponent } from './settings.component';
 
-import { FeatureFlagResolver } from 'ngx-feature-flag';
-
 const routes: Routes = [
   {
     path: '',
@@ -17,17 +15,6 @@ const routes: Routes = [
       {
         path: 'feature-opt-in',
         loadChildren: './feature-opt-in/feature-opt-in.module#FeatureOptInModule'
-      },
-      {
-        path: 'notifications',
-        loadChildren: './notifications/notifications.module#NotificationsModule',
-        resolve: {
-          featureFlagConfig: FeatureFlagResolver
-        },
-        data: {
-          title: 'Notifications',
-          featureName: 'ProfileSettingsNotifications'
-        }
       },
       {
         path: 'resources',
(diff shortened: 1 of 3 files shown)


alien-ike:
### Ike Plugins (test-keeper)

jiekang:
Changes look okay to me.

@corinnekrych @rohitkrai03 
What is the user experience if they are not an internal Red Hat user? The notifications tab will be visible, they click on it. It shows the internal feature flag restriction banner with link to Features Opt-In tab and they won't see the option to enable internal features. Is that right? I think having it as Experimental may be more appropriate.

andrewazores:
Good point, Jie. There was a discussion a while back about what the various levels should be used for, and `internal` was IIRC just for things that should be hidden from external users due to things like embargo - otherwise, `experimental` would be the default choice for new and, well, experimental, features.

alexeykazakov:
I agree with @jiekang and @andrewazores.
Since the tab does nothing beyond showing not functional UI I would hide the tab completely for external users. But if it's technically tricky to hide then let's use the experimental flag for the content of the tab.

rohitkrai03:
@alexeykazakov @jiekang @andrewazores  I tried adding feature flag to the tab as well. But since its a custom sub-menu of settings page there were some CSS issues due to use of `<f8-feature-toggle` element. I've fixed that using some CSS changes but it still seem like a hack to me. 
I've updated the PR with new changes. If it seems okay to you guys then we can go ahead with this or we can move the feature level to `experimental`.

andrewazores:
The CSS can be done in a less hacky way by refactoring it a bit first, say by extracting a class to apply to the `<li>` elements. There's likely an even better way than that but I'm far from a CSS expert.

rohitkrai03:
I've not added any extra CSS specifically for this. I've just used already present pf classes.

alexeykazakov:
I wonder if we should just remove this tab at all. It does nothing. And we don't have any plans to bring any functionality to this notification tab in the near future. Not sure when we will have time to work on it..
If we can't hide this tab quickly without risking to break something I would remove it.

@catrobson what do you think? We do want to do some UI clean up anyway, right? :)

jiekang:
That's a fair point actually. I'd be in favour of removing for the time being.

rohitkrai03:
@alexeykazakov @jiekang What are we deciding here then? Should i just remove the tab from UI altogether?

rohitkrai03:
@alexeykazakov @jiekang FYI, the current changes that is there in PR doesn't break anything on the UI. It works as intended.

Screenshots - 

Default - 
screen shot 2018-08-27 at 8 03 25 pm

With internal feature flag - 

screen shot 2018-08-27 at 8 03 54 pm

alexeykazakov:
@rohitkrai03 wait.. I thought you said that we will see the notification tab anyway? But when clicking on the tab it will ask to enable "experimental features" to see the content of the tab. It's not the case?
You actually can hide the tab completely?

rohitkrai03:
@alexeykazakov In the comment https://github.com/fabric8-ui/fabric8-ui/pull/3277#issuecomment-415853676 I mentioned I was able to hide the tab completely using feature flag but it had some CSS issue when shown which I handled using CSS fix.

alexeykazakov:
@rohitkrai03 right... so, this is with the “fixed” CSS.

Let’s just remove this tab. We will bring it back when we have something on the backend side.

rohitkrai03:
@alexeykazakov Done. 
- Removed the tab from submenu.
- Code for the notification component is still there.

cc: @jiekang

jiekang:
@rohitkrai03 What's the expected path if a user visits this page while not logged in?

rohitkrai03:
@jiekang It shows 404 error when visiting the URL without logging in.

rohitkrai03:
@jiekang I've actually removed notifications from routing. So, now the URL won't be accessible as well.

jiekang:
@rohitkrai03 My bad... I actually asked that question on the wrong PR. But I guess it caused you to change something, so maybe it applied here too? Sorry about that though!!

jiekang:
Is it worth it to make an issue to track the re-enablement of this notifications page feature in the future?

PR #3282, opened 2018-08-24
Upgrade use of new work item service from planner

sanbornsen:
The work item service in planner has changed. Now it takes a formatted filter expression which can be constructed by the filter service in planner.

In this PR fabric8-ui is updated to use the new work item service. The old get work item end point is deprecated and will be discontinued from backend services.

### DIFF
--- a/src/app/profile/overview/work-items/work-items.component.ts
+++ b/src/app/profile/overview/work-items/work-items.component.ts
@@ -105,20 +105,6 @@ export class WorkItemsComponent implements OnDestroy, OnInit {
         .map(val => val.workItems)
         .map(workItems => filterOutClosedItems(workItems))
         // Resolve the work item type, creator and area
-        .do(workItems => workItems.forEach(workItem => this.workItemService.resolveType(workItem)))
-        .do(workItems => workItems.forEach(workItem => {
-          try {
-            this.workItemService.resolveAreaForWorkItem(workItem);
-          } catch (error) { /* No space */ }
-        }))
-        .do(workItems => {
-          workItems.forEach(workItem => {
-            if (workItem.relationalData === undefined) {
-              workItem.relationalData = {};
-            }
-          });
-        })
-        .do(workItems => workItems.forEach(workItem => this.workItemService.resolveCreator(workItem)))
         .subscribe(workItems => {
           this.workItems = workItems;
         })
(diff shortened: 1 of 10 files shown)


corinnekrych:
LGTM

src/app/profile/overview/work-items/work-items.component.ts
```diff
@@ -84,28 +85,43 @@ export class WorkItemsComponent implements OnDestroy, OnInit {
...
-            included?: WorkItem[];
-            ancestorIDs?: string[];
-          }): void => {
-            this.workItems = filterOutClosedItems(result.workItems);
-          }
-        )
+        .getWorkItems(100000, {expression: filters})
+        .map(val => val.workItems)
+        .map(workItems => filterOutClosedItems(workItems))
+        // Resolve the work item type, creator and area
```
christianvogt: This seems like a potential performance issue wherein for each work item you're now fetching from the server all this extra information.

Why is this info now necessary when previously it wasn't?

sudsen: Good catch. We don't use resolvers in planner anymore. Thus I moved all those operations for resolving in here. But true, that we are resolving more than we need. I'll fix that.

jiekang:
@sanbornsen Please address the unit test failure introduced in this PR:

```
SUMMARY:
✔ 799 tests completed
ℹ 7 tests skipped
✖ 1 test failed

FAILED TESTS:
  WorkItemWidgetComponent
    ✖ should set relational data to an empty obj if it does not exist prior
```

jiekang:
The work items widget on the space overview page also now shows work items across spaces, rather than only those in the specific space. The links to the work items assume they come from the space being viewed and are broken for the work items that aren't in said space. This is another regression that needs to be addressed.

aptmac:
The unit test is failing because the line of code that creates the empty `relationalData` attribute if it doesn't exist was removed in commit [[0]](https://github.com/fabric8-ui/fabric8-ui/pull/3282/commits/2fa1bed6345fb4fd51ed0a6ed2f2deed5b6a20d4) at line [[1]](https://github.com/fabric8-ui/fabric8-ui/commit/2d974cce9f2dc480458ec53be6c6b6c40d323bfa#diff-4fe297ab8149ae0e1897fc8a9767ef3cL100)

[0] https://github.com/fabric8-ui/fabric8-ui/pull/3282/commits/2fa1bed6345fb4fd51ed0a6ed2f2deed5b6a20d4
[1] https://github.com/fabric8-ui/fabric8-ui/commit/2d974cce9f2dc480458ec53be6c6b6c40d323bfa#diff-4fe297ab8149ae0e1897fc8a9767ef3cL100

PR #3283, opened 2018-08-24
fix(css): use google cdn for opensans fonts

jiekang:
This addresses https://openshift.io/openshiftio/Openshift_io/plan/detail/156

The OpenSans fonts are downloaded from CDN instead of being packaged in our distribution. A local prod build via `npm run build:prod` has a reduction of ~13 MB, from 36 MB to 23 MB on my machine.


alien-ike:
### Ike Plugins (test-keeper)

jiekang:
At this time `pre-build.sh` exists to create the fontless css/less Patternfly files used in the project. An issue has been opened for Pattiernfly asking if they can package these in their distribution. Otherwise I'll be trying to run these 'pre-build' in the npm scripts.

post-install.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# Removes first line of patternfly.min.css : This assumes this is the line containing the Open Sans font definitions
+tail -n +2 "node_modules/patternfly/dist/css/patternfly.min.css" > "node_modules/patternfly/dist/css/patternfly.fontless.min.css"
```
andrewazores: I don't really like this "remove the first line" approach... couldn't a similar `sed` trick as below be done for this step, too?

jiekang: Yeah good point. I've addressed it now

alien-ike:
Hey @andrewazores! It seems you tried to trigger `/ok-without-tests` command, but this will not have any effect due to insufficient permission. You have to be admin or requested reviewer or pull request approver, but not pull request creator for this command to take an effect.

andrewazores:
I'll test this locally but I think it seems reasonable from a reading.

jiekang:
Interesting failure from CICO:
```
++ cico node get -f value -c ip_address -c comment
The requested operation failed as no inventory is available.
+ cico_output=
+ '[' 1 -eq 0 ']'
+ '[' 6 -gt 5 ']'
+ echo 'giving up on '\''cico node get'\'''
giving up on 'cico node get'
+ exit 1
Build step 'Execute shell' marked build as failure
```

PR #3285, opened 2018-08-27
feat(interceptors): add interceptors to replicate Fabric8UIHttpService

rohitkrai03:
- Adds an interceptor to cache GET requests.
- Adds an interceptor to add `X-Request-Id` header for `WIT` and `AUTH` URLs.

### DIFF
--- a/src/app/shared/interceptors/cache.interceptor.ts
+++ b/src/app/shared/interceptors/cache.interceptor.ts
@@ -59,10 +59,7 @@ function sendRequest(
   next: HttpHandler,
   cache: RequestCache): Observable<HttpEvent<any>> {
 
-  // No headers allowed in npm search request
-  const noHeaderReq = req.clone({ headers: new HttpHeaders() });
-
-  return next.handle(noHeaderReq).pipe(
+  return next.handle(req).pipe(
     tap(event => {
       // There may be other events besides the response.
       if (event instanceof HttpResponse) {
--- a/src/app/shared/interceptors/cache.interceptor.ts
+++ b/src/app/shared/interceptors/cache.interceptor.ts
@@ -21,18 +21,20 @@ export class CacheInterceptor implements HttpInterceptor {
 
   intercept(req: HttpRequest<any>, next: HttpHandler) {
     // continue if not cachable.
-    if (!isCachable(req)) { return next.handle(req); }
+    if (!isCachable(req)) {
+      return next.handle(req);
+    }
 
     const cachedResponse = this.cache.get(req);
 
     return Observable.create(observer => {
       if (cachedResponse) {
         return cachedResponse.subscribeOn(Scheduler.async).subscribe(observer);
-      } else {
-        const asyncResponse = new AsyncSubject<HttpResponse<any>>();
-        this.cache.set(req, asyncResponse);
-        return next.handle(req).subscribe(asyncResponse);
       }
+
+      const asyncResponse = new AsyncSubject<HttpResponse<any>>();
+      this.cache.set(req, asyncResponse);
+      return next.handle(req).subscribe(asyncResponse);
     });
   }
 }
--- a/src/app/shared/interceptors/cache.interceptor.ts
+++ b/src/app/shared/interceptors/cache.interceptor.ts
@@ -2,7 +2,7 @@
 import { Injectable } from '@angular/core';
 
 import {
-  HttpEvent, HttpHandler, HttpHeaders,
+  HttpEvent, HttpHandler,
   HttpInterceptor, HttpRequest, HttpResponse
 } from '@angular/common/http';
 
--- a/src/app/shared/interceptors/cache.interceptor.ts
+++ b/src/app/shared/interceptors/cache.interceptor.ts
@@ -1,0 +1,74 @@
+// #docplaster
+import { Injectable } from '@angular/core';
+
+import {
+  HttpEvent, HttpHandler, HttpHeaders,
+  HttpInterceptor, HttpRequest, HttpResponse
+} from '@angular/common/http';
+
+import { Observable } from 'rxjs';
+import { startWith, tap } from 'rxjs/operators';
+
+import { RequestCache } from '../request-cache.service';
+
+/**
+ * If request is cachable (e.g., package search) and
+ * response is in cache return the cached response as observable.
+ * If has 'x-refresh' header that is true,
+ * then also re-run the package search, using response from next(),
+ * returning an observable that emits the cached response first.
+ *
+ * If not in cache or not cachable,
+ * pass request through to next()
+ */
+@Injectable()
+export class CacheInterceptor implements HttpInterceptor {
+  constructor(private cache: RequestCache) {}
+
+  intercept(req: HttpRequest<any>, next: HttpHandler) {
+    // continue if not cachable.
+    if (!isCachable(req)) { return next.handle(req); }
+
+    const cachedResponse = this.cache.get(req);
+    // cache-then-refresh
+    if (req.headers.get('x-refresh')) {
+      const results$ = sendRequest(req, next, this.cache);
+      return cachedResponse ?
+        results$.pipe(startWith(cachedResponse)) :
+        results$;
+    }
+    // cache-or-fetch
+    return cachedResponse ?
+      Observable.of(cachedResponse) : sendRequest(req, next, this.cache);
+  }
+}
+
+
+/** Is this request cachable? */
+function isCachable(req: HttpRequest<any>) {
+  // Only GET requests are cachable
+  return req.method === 'GET';
+}
+
+/**
+ * Get server response observable by sending request to `next()`.
+ * Will add the response to the cache on the way out.
+ */
+function sendRequest(
+  req: HttpRequest<any>,
+  next: HttpHandler,
+  cache: RequestCache): Observable<HttpEvent<any>> {
+
+  // No headers allowed in npm search request
+  const noHeaderReq = req.clone({ headers: new HttpHeaders() });
+
+  return next.handle(noHeaderReq).pipe(
+    tap(event => {
+      // There may be other events besides the response.
+      if (event instanceof HttpResponse) {
+        cache.set(req, event); // Update the cache.
+      }
+    })
+  );
+}
+
--- a/src/app/shared/interceptors/cache.interceptor.ts
+++ b/src/app/shared/interceptors/cache.interceptor.ts
@@ -1,22 +1,16 @@
-// #docplaster
 import { Injectable } from '@angular/core';
 
 import {
-  HttpEvent, HttpHandler,
-  HttpInterceptor, HttpRequest, HttpResponse
+  HttpHandler, HttpInterceptor,
+  HttpRequest, HttpResponse
 } from '@angular/common/http';
 
-import { Observable } from 'rxjs';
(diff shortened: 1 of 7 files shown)


alien-ike:
### Ike Plugins (test-keeper)

src/app/shared/request-cache.service.ts
```diff
@@ -0,0 +1,56 @@
...
+  response: HttpResponse<any>;
+  lastRead: number;
+}
+
+export abstract class RequestCache {
+  abstract get(req: HttpRequest<any>): HttpResponse<any> | undefined;
+  abstract set(req: HttpRequest<any>, response: HttpResponse<any>): void;
+}
+
+const maxAge = 300000; // maximum cache age (ms)
```
christianvogt: Why such a large number? This has potential to continue to show stale data. The original cache was put in place only to prevent many duplicate requests going out at the same time.

src/app/shared/request-cache.service.ts
```diff
@@ -0,0 +1,56 @@
...
+import { HttpRequest, HttpResponse } from '@angular/common/http';
+
+
+export interface RequestCacheEntry {
+  url: string;
+  response: HttpResponse<any>;
+  lastRead: number;
+}
+
+export abstract class RequestCache {
```
christianvogt: This can be made an interface.

src/app/shared/interceptors/cache.interceptor.ts
```diff
@@ -0,0 +1,74 @@
...
+  cache: RequestCache): Observable<HttpEvent<any>> {
+
+  // No headers allowed in npm search request
+  const noHeaderReq = req.clone({ headers: new HttpHeaders() });
+
+  return next.handle(noHeaderReq).pipe(
+    tap(event => {
+      // There may be other events besides the response.
+      if (event instanceof HttpResponse) {
+        cache.set(req, event); // Update the cache.
```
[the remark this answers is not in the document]
rohitkrai03: @christianvogt I've updated the implementation to use `AsyncSubject`. Please check.

christianvogt:
Why are you moving the auth interceptor?

rohitkrai03:
@christianvogt There are a few reasons why i thought it makes sense - 
- Currently `ngx-login-client` is not updated in fabric8-ui so auth interceptor could not be added now. Since we moved to `HttpClient`, a lot of requests now are not going through the error catching for 401 request that old http.service from `ngx-login-client` used to handle. This fixes that bug and handles error.
- The extra logic `X-Request-Id` currently present in `Fabric8UIHttpService` has some dependencies that are not there in `ngx-login-client`.
- Also, i need to work on RPT token refresh workflow for which i am thinking of adding the logic in auth interceptor only. This way i can start working on it and test it better.
- I also think its better to have all the interceptors in one place and not scattered through various libs so that its easy to make changes or update them later.

Let me know if it doesn't seem like a good idea. I can revert back the last commit.

corinnekrych:
@rohitkrai03 I think there should be only one Auth interceptor. I think it makes sense to have the auth interceptor packaged in `ngx-login-client`. inteceptors will not be scattered but nicely fit in the right box. Auth interceptor => auth lib.
Agreed with you it seems https://github.com/fabric8-ui/ngx-login-client/blob/master/src/app/shared/auth.interceptor.ts does not handle request-id like you did here: https://github.com/fabric8-ui/fabric8-ui/blob/f637afb70f94032a464b799ca2f7bbcf071e054b/src/app/shared/interceptors/auth.interceptor.ts#L43-L52 let's do a PR on `ngx-login-client`.
Let's discuss on MM channel on how to release and 2e2 test. cc @christianvogt @joshuawilson

rohitkrai03:
@corinnekrych Makes sense. Will remove auth interceptor from here and raise a PR on `ngx-login-client`. The question is can we do a 1.x release on `ngx-login-client` so that it can be used in fabric8-ui before angular upgrade? Or we should wait for fabric8-ui to start using 2.x version?

christianvogt:
Why is `x-request-id` part of of the auth interceptor? Seems to me like a separate interceptor altogether.

rohitkrai03:
@christianvogt Since auth interceptor is already adding a header I thought it could also add `X-Request-Id` header. But now rethinking about it seems that auth interceptor is adding auth header so it would be better to separate adding of other headers in separate interceptor.

joshuawilson:
You can make a v1 branch that I can manually release. Base it off of https://github.com/fabric8-ui/ngx-login-client/commit/9804f23

I would also like this change in master too. Please submit 2 PRs.

joshuawilson:
Is this ready? It doesn't update the login-client version.

rohitkrai03:
@joshuawilson 1.4.6 of ngx-login-client has some problem with directory structure. I've left a comment on the PR there as well. Can you take a look?

joshuawilson:
can we merge this without that?

rohitkrai03:
Yes, we can merge it without that.

joshuawilson:
ok, I'll review it. thanks

src/app/shared/interceptors/request-id.interceptor.ts
```diff
@@ -0,0 +1,41 @@
...
+    @Inject(AUTH_API_URL) private authApiUrl: string
+  ) {}
+
+  intercept(request: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
+    const url = request.url;
+
+    // Attach a X-Request-Id to all requests that don't have one
+    if (url.startsWith(this.witApiUrl) || url.startsWith(this.authApiUrl)) {
+      if (!request.headers.has('X-Request-Id')) {
+        request = request.clone({
```
joshuawilson: I'm not sure you can do this. I think the `request` is immutable and you are writing back to it. 

When I run the application it has some 401 errors with malformed headers. I'm not sure if this is caused by this change or if it is old.

src/app/shared/request-cache.service.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    // remove expired cache entries
+    const expired = Date.now() - CACHE_TTL;
+    this.cache.forEach(cacheItem => {
+      if (cacheItem.lastRead < expired) {
+        this.cache.delete(cacheKey);
+      }
+    });
+
+    console.log(`Request cache size: ${this.cache.size}.`);
```
bartoszmajsak: That should not be part of "production" code.

src/app/shared/interceptors/request-id.interceptor.ts
rohitkrai03: @joshuawilson Yes, `request` is immutable and that is wy we clone it and mutate the cloned object rather than the original. This is how the example on angular.io adds header in interceptor.

rohitkrai03: @joshuawilson The 401s might not be related to this.

src/app/shared/request-cache.service.ts
rohitkrai03: I'll remove all the console.log references. Maybe we can add `babel-plugin-transform-remove-console` in webpack config and have all the console.log lines removed at build time if its build for prod?

christianvogt: IIRC uglify can remove console log too with a single config option.

src/app/shared/interceptors/request-id.interceptor.ts
joshuawilson: You close the `request` and then write it back to the same `request`. The examples create a new `const`, they don't use the `request`.

rohitkrai03: Updated

joshuawilson:
Is this ready? Please resolve issues and rebase.

rohitkrai03:
@christianvogt Can you review this again?

src/app/shared/request-cache.service.ts
```diff
@@ -0,0 +1,56 @@
...
+  get(req: HttpRequest<any>): AsyncSubject<HttpResponse<any>> | undefined {
+    const cacheKey = createCacheKey(req);
+    const cached = this.cache.get(cacheKey);
+
+    if (!cached) {
+      return undefined;
+    }
+
+    const isExpired = cached.lastRead < (Date.now() - CACHE_TTL);
+    const expired = isExpired ? 'expired ' : '';
```
christianvogt: This variable is not used.

src/app/shared/request-cache.service.ts
```diff
@@ -0,0 +1,56 @@
...
+  asyncResponse: AsyncSubject<HttpResponse<any>>;
+  lastRead: number;
+}
+
+const CACHE_TTL = 300; // maximum cache age (ms)
+
+@Injectable()
+export class RequestCache {
+
+  cache = new Map<string, RequestCacheItem>();
```
christianvogt: Make `cache` private.

src/app/shared/request-cache.service.ts
```diff
@@ -0,0 +1,56 @@
...
+        this.cache.delete(cacheKey);
+      }
+    });
+  }
+}
+
+// Creates a cache key from headers, query params, and URL
+function createCacheKey(req: HttpRequest<any>): string {
+  let key;
+  key = req.urlWithParams;
```
christianvogt: Combine this line with the declaration of `key` above.

src/app/shared/interceptors/cache.interceptor.ts
```diff
@@ -0,0 +1,45 @@
...
+ * If not in cache or not cachable,
+ * pass request through to next()
+ */
+@Injectable()
+export class CacheInterceptor implements HttpInterceptor {
+  constructor(private cache: RequestCache) {}
+
+  intercept(req: HttpRequest<any>, next: HttpHandler) {
+    // continue if not cachable.
+    if (!isCachable(req)) { return next.handle(req); }
```
christianvogt: Format this line so that the block is on separate lines.

src/app/shared/interceptors/cache.interceptor.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    const cachedResponse = this.cache.get(req);
+
+    return Observable.create(observer => {
+      if (cachedResponse) {
+        return cachedResponse.subscribeOn(Scheduler.async).subscribe(observer);
+      } else {
```
christianvogt: This else block is unnecessary due to the if block having a `return`. Remove the else, but move the code outside of the block.

rohitkrai03:
@christianvogt Updated PR based on review comments.

aptmac:
There seems to be something odd happening with the after this PR (the cache interceptor in particular), for an example check out the recent workspaces widget on the home page (under experimental >=). 

The widget is stuck in a loading state because the get request from the codebases service `getCodebases` doesn't return. The get request is made in my network tab, and I can see the response, but the chain of events doesn't continue so the component never receives the data. This is not present on the codebase widgets on the dashboard, because despite the network tab saying that component is using a get request, the cache interceptor is telling me it's a patch, so it's not getting caught at [[0]](https://github.com/fabric8-ui/fabric8-ui/pull/3285/files#diff-ac190afe59688366d751b04f253f3119R24).

[0] https://github.com/fabric8-ui/fabric8-ui/pull/3285/files#diff-ac190afe59688366d751b04f253f3119R24

joshuawilson:
@christianvogt are you submitting a fix for it?

PR #3294, opened 2018-08-29
feat(my-spaces): add feature toggle for search feature

andrewazores:
related to github.com/openshiftio/openshift.io/issues/3655


andrewazores:
The MySpaces.SearchSpaces feature has been defined for both prod and preview as an "experimental" level feature.

PR #3296, opened 2018-08-29
fix(auth): add auth header in provider service

rohitkrai03:
Fixes https://github.com/openshiftio/openshift.io/issues/4268
Fixes https://github.com/fabric8-services/fabric8-auth/issues/633

### DIFF
--- a/src/app/shared/account/provider.service.spec.ts
+++ b/src/app/shared/account/provider.service.spec.ts
@@ -17,6 +17,7 @@ describe('Service: Provider Service', () => {
     let service: ProviderService;
     let controller: HttpTestingController;
     let mockLogger: jasmine.SpyObj<Logger>;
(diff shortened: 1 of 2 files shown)


alien-ike:
### Ike Plugins (test-keeper)

rohitkrai03:
@jiekang Can you review this? Its a SEV-1 issue.

sbose78:
Hi @rohitkrai03 , Would it be reasonable to update the test to try validating the presence of the header to https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/shared/account/provider.service.spec.ts ?

rohitkrai03:
@sbose78 The tests actually validate presence of auth header by mocking authService https://github.com/fabric8-ui/fabric8-ui/blob/5ea0a594b37587905c96c184e647ec3945c4b1ab/src/app/shared/account/provider.service.spec.ts#L28. that is the reason this was not caught since all tests were passing.

rohitkrai03:
@sbose78 Actually no, the token was for validating something else. I am adding tests to verify auth header.

rohitkrai03:
@jiekang I've added tests for header. Also, https://github.com/fabric8-ui/fabric8-ui/pull/3285 is for replicating `Fabric8UIHttpService` by use of interceptors. I am working on adding auth interceptor from `ngx-login-client` which adds header for `HttpClient` request.

src/app/shared/account/provider.service.spec.ts
```diff
@@ -12,11 +12,12 @@ import { ProviderService } from './provider.service';
 
 import { createMock } from 'testing/mock';
 
-describe('Service: Provider Service', () => {
+fdescribe('Service: Provider Service', () => {
```
jiekang: I think this is for local testing?

rohitkrai03: Sorry, removed.

src/app/shared/account/provider.service.spec.ts
```diff
@@ -60,6 +60,7 @@ describe('Service: Provider Service', () => {
 
       const req: TestRequest = controller.expectOne('http://example.com/api/token?force_pull=true&for=https://github.com');
       expect(req.request.method).toEqual('GET');
+      expect(req.request.headers.get('Authorization')).toEqual(`Bearer ${mockToken}`);
```
sbose78: Thanks!

joshuawilson:
We should consider removing the customer http service with the addition of the new httpclient.

PR #3298, opened 2018-08-31
fix(profile): clean up overview spaces component code

jiekang:
This cleans up code in the overview spaces component and renames the file and it's selector to not conflict with another existing `spaces.component.ts` with selector `alm-spaces`. The conflict doesn't actually affect anything but the distinction makes it less confusing for developers.


src/app/profile/overview/spaces/overview-spaces.component.ts
```diff
@@ -48,14 +42,11 @@ export class SpacesComponent implements OnDestroy, OnInit  {
     });
   }
 
-  // Actions
-
-  initSpaces(event: any): void {
+  initSpaces(pageSize: number): void {
```
andrewazores: This function is only called once via `ngOnInit`, so it'll only ever have `pageSize` of `this.pageSize` passed to it, right? `this.pageSize` looks to be effectively a constant. Is there any reason for it to be a mutable property, or could it be extracted out to a proper constant and referenced directly by `initSpaces` rather than passed through as a param?

PR #3299, opened 2018-08-31
feat(add-codebase-widget): add Che workspace widget to codebase items

andrewazores:
Add Che workspace widget to codebase items, allowing users to quickly jump into a Che workspace for a given codebase - or create a workspace if none exist - without first visiting the Create > Codebases page
 
This is the same workspace widget (reused component) from the Codebases page.

A "Codebases.QuickChe" feature toggle has been added to hide this added component from the dashboard Codebases widget. It can also be used for related sibling workitems such as adding deep links from the Pipelines view back to Codebases, or possibly for enabling deep linking functionality on the Codebases page itself.

Before:

image

After:

image

fixes https://openshift.io/openshiftio/Openshift_io/plan/detail/504


src/app/dashboard-widgets/add-codebase-widget/codebase-item/codebase-item.component.html
```diff
@@ -0,0 +1,24 @@
...
+  </div>
+  <div class="list-view-pf-main-info f8-card-list-view-pf-main-info">
+    <div class="list-view-pf-body">
+      <div class="list-view-pf-description">
+        <div class="list-group-item-heading f8-card-list-group-item-heading">
+          <a href="{{codebase.attributes.url}}" target="_blank" class="f8-card-codebase-url truncate">{{codebase.name}}</a>
+        </div>
+        <div class="list-group-item-text f8-card-list-group-item-text">
+          codebase last updated 03/12/2018
+          <!-- replace with real code -->
```
jiekang: Hmm.. do you think this could be replaced now? Otherwise I'd prefer it hidden rather than using a placeholder.

andrewazores: Not sure, I intentionally just copied the component verbatim since that part is out of scope of this feature addition. FWIW looking at the `Codebase` type I don't see a property for "last updated". Perhaps it's meant to be the `codebase.gitHubRepo.pushedAt`? I'm not sure what the expectation for the "last updated" item is supposed to be and if a push event is the only thing that would be included.

jiekang: Hmm maybe it's `pushedAt`? No idea; could you comment the whole div?

src/app/dashboard-widgets/add-codebase-widget/codebase-item/codebase-item.component.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  type TestingContext = TestContext<CodebaseItemComponent, HostComponent>;
+  initContext(CodebaseItemComponent, HostComponent, {
+    declarations: [
+      MockCodebasesItemWorkspacesComponent,
+      MockFeatureToggleComponent
+    ]
+  });
+
+  it('should be instantiable', function(this: TestingContext): void {
+    expect(this.testedDirective).toBeDefined();
```
jiekang: I'd be inclined not to include this.  Thoughts?

andrewazores: Sure, I can take it out. I usually write those sorts of tests first just as a sanity check for the spec suite itself before continuing on to other more meaningful tests.

src/app/dashboard-widgets/add-codebase-widget/codebase-item/codebase-item.component.html
andrewazores: At least for my codebases in my test-space, the `gitHubRepo` property is undefined and so `pushedAt` is not there either. I'll file a separate issue to fix that placeholder and address it separately.

andrewazores: https://openshift.io/openshiftio/Openshift_io/plan/detail/542

PR #3304, opened 2018-09-05
fix(profile): clean update component code and fix tests

jiekang:
This cleans up the code in update component, removing unused code and adding type definitions where possible. The irrelevant tests are removed and the spec is fixed to properly init and destroy, so the following no longer occurs.

```
ERROR: 'Error during cleanup of component', UpdateComponent [...]
```


src/app/profile/update/update.component.spec.ts
```diff
@@ -11,25 +11,23 @@ import { AuthenticationService, User, UserService } from 'ngx-login-client';
...
 import { GettingStartedService } from '../../getting-started/services/getting-started.service';
-import { ProviderService } from '../../shared/account/provider.service';
 import { GitHubService } from '../../space/create/codebases/services/github.service';
 import { CopyService } from '../services/copy.service';
 import { TenantService } from '../services/tenant.service';
 import { TenantUpdateStatus, UpdateComponent } from './update.component';
 
 
-describe('UpdateComponent', () => {
+fdescribe('UpdateComponent', () => {
```
rohitkrai03: I think this was for local testing.

jiekang: Right, removed. Thanks!

andrewazores:
In manually testing this branch I have some strange behaviour when attempting to update my name. I edited my name from "Andrew Azores" to "Andrew J Azores" and hit "Save", which caused a "Failed to update profile" notification to appear and the view to not reset to read-only state. However when I refresh the view, my name is correctly updated. If I test the same name editing workflow on production, hitting "Save" is followed by a "Profile updated" success notification, with the view going into read-only state and reflecting the changed name. I haven't manually tested any other functionality yet.

jiekang:
Could you check the error in more detail? I've been updating multiple fields in my profile locally without issue multiple times. I had the issue you described when trying on production, but it cleared up after a hard refresh and I can't get it to happen again.

andrewazores:
```
{…}​error: Object { errors: (1) […] }​headers: Object { normalizedNames: Map(0), lazyUpdate: null, lazyInit: lazyInit() }​message: "Http failure response for https://openshift.io/api/users: 500 Internal Server Error"​name: "HttpErrorResponse"​ok: false​status: 500​statusText: "Internal Server Error"​url: "https://openshift.io/api/users"​<prototype>: Object { constructor: HttpErrorResponse() } console.js:32
```

```
{"errors":[{"code":"internal_error","detail":"received a non-200 response 403 Forbidden while fetching keycloak user profile https://sso.openshift.io/auth/realms/fabric8/account","status":"500","title":"Internal error"}]}
```

Huh - that does look like it's caused by something other than this patch. I'll dig in.

PR #3310, opened 2018-09-07
fix(work-item-widget): fix incorrect total of work items

aptmac:
This PR addresses https://github.com/openshiftio/openshift.io/issues/4303, where the dashboard work item widget can display an incorrect number of total work items. This occurs because it seems the workitemservice will return at most 500 work items, and the total count is obtained by getting the length of the array of work items. However, the workitemservice also returns an actual number with the total number of work items that we can use here.

### DIFF
--- a/src/app/dashboard-widgets/work-item-widget/work-item-widget.component.ts
+++ b/src/app/dashboard-widgets/work-item-widget/work-item-widget.component.ts
@@ -83,9 +83,15 @@ export class WorkItemWidgetComponent implements OnInit {
           100000, {expression: {'space': `${space.id}`}}
         );
       })
+      .do((val: { workItems: WorkItem[], nextLink: string, totalCount?: number, included?: WorkItem[], ancestorIDs?: string[]}): void => {
+        if (val.totalCount) {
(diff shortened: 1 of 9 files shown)


src/app/dashboard-widgets/work-item-widget/work-item-widget.component.ts
```diff
@@ -83,9 +83,15 @@ export class WorkItemWidgetComponent implements OnInit {
           100000, {expression: {'space': `${space.id}`}}
         );
       })
+      .do((val: { workItems: WorkItem[], nextLink: string, totalCount?: number, included?: WorkItem[], ancestorIDs?: string[]}): void => {
```
andrewazores: Does `WorkItemService` export a definition for this type? If not, I'd like to see this lifted out from this inline definition to the file top-level, or maybe at the class level. I would also probably open an issue against Planner to export a definition for consumers to use.

aptmac:
I added a type definition for this under the `workitem-utils` in the `shared` folder, so it's more intuitive when used throughout the codebase. I also added this type to all instances where `getWorkItems` is used.

PR #3311, opened 2018-09-11
fix(cache): fix cache interceptor to return the proper observable

christianvogt:
Related to: https://github.com/fabric8-ui/fabric8-ui/pull/3285#issuecomment-420281716

The cache interceptor wasn't returning the correct observable.
Observed the Recent Workspaces widget was stuck on loading even though the network request went through.

I was able to reproduce this in dev environment and with the fix, the widget loaded correctly.

image

### DIFF
--- a/src/app/shared/interceptors/cache.interceptor.spec.ts
+++ b/src/app/shared/interceptors/cache.interceptor.spec.ts
@@ -1,19 +1,30 @@
-import { HTTP_INTERCEPTORS, HttpClient, HttpHeaders, HttpParams } from '@angular/common/http';
+import {
+  HTTP_INTERCEPTORS,
+  HttpClient,
+  HttpEvent,
+  HttpHandler,
+  HttpHeaders,
+  HttpParams,
+  HttpRequest,
+  HttpResponse
+} from '@angular/common/http';
 import {
   HttpClientTestingModule,
   HttpTestingController
 } from '@angular/common/http/testing';
-import { TestBed } from '@angular/core/testing';
+import { fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { Observable } from 'rxjs';
 import { CacheInterceptor } from './cache.interceptor';
 
 import { RequestCache } from '../request-cache.service';
 
-describe(`AuthHttpInterceptor`, () => {
+describe(`CacheHttpInterceptor`, () => {
   const testUrl: string = 'http://localhost/test';
   const oriDateNow: (() => number) = Date.now;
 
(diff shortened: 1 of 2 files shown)


src/app/shared/interceptors/cache.interceptor.ts
```diff
@@ -25,16 +25,16 @@ export class CacheInterceptor implements HttpInterceptor {
       return next.handle(req);
     }
 
-    const cachedResponse = this.cache.get(req);
-
     return Observable.create(observer => {
-      if (cachedResponse) {
-        return cachedResponse.subscribeOn(Scheduler.async).subscribe(observer);
+      let asyncResponse = this.cache.get(req);
```
rohitkrai03: @christianvogt The only major change i can see here is the ` let asyncResponse = this.cache.get(req);` is inside `Observable.create`. Apart from that only the ordering of condition check is different. But i am not sure how this affects returning proper response?

rohitkrai03: Aah, got it. Never mind. `Observable.create` used to close over the value of `cachedResponse` variable which is why it was not returning proper response.

src/app/shared/interceptors/cache.interceptor.spec.ts
```diff
@@ -91,4 +103,35 @@ describe(`AuthHttpInterceptor`, () => {
     httpClient.get('test2').subscribe();
     httpMock.expectOne('test2');
   });
+
+  fit('should return cached observable and not make several handler requests', fakeAsync(() => {
```
rohitkrai03: I guess this is for local testing purpose.

aptmac: `fit` -> `it`

christianvogt:
Added a unit test.

You guys are quick :) I was in the process of changing `fit` -> `it`.

bartoszmajsak:
@christianvogt it's interesting to see though that this PR was approved by the bot. Trying to figure out what could be the cause of that.... 

> alien-ike/test-keeper Seems that this PR doesn't need to have tests

rohitkrai03:
@bartoszmajsak Even my last related PR - https://github.com/fabric8-ui/fabric8-ui/pull/3285 had tests in it but the bot was saying it didn't have any tests. We had to add `/ok-without-tests` before merging the PR.

bartoszmajsak:
@rohitkrai03 interesting... so it somewhat reversed the check value... I will have to investigate

PR #3312, opened 2018-09-12
refactor(profile): use rxjs pipe instead of operator chaining

rohitkrai03:
- Use rxjs `pipe()` for chained operators.
- `Observable.of()` -> `of()`

### DIFF
--- a/src/app/profile/cleanup/cleanup.component.ts
+++ b/src/app/profile/cleanup/cleanup.component.ts
@@ -9,6 +9,10 @@ import { EventService } from '../../shared/event.service';
 import { IModalHost } from '../../space/wizard/models/modal-host';
 import { TenantService } from '../services/tenant.service';
 
+import { forkJoin } from 'rxjs/observable/forkJoin';
+import { of } from 'rxjs/observable/of';
+import { catchError, map } from 'rxjs/operators';
+
 @Component({
   encapsulation: ViewEncapsulation.None,
   selector: 'fabric8-cleanup',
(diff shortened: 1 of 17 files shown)


andrewazores:
Please amend the commit message to point to this issue: https://openshift.io/openshiftio/Openshift_io/plan/detail/579

rohitkrai03:
@andrewazores Done.

christianvogt:
Changes look ok but please confirm: shouldn't all the imports be like this?
`import { of } from 'rxjs';`
instead of 
`import { of } from 'rxjs/observable/of';`

src/app/profile/cleanup/cleanup.component.ts
```diff
@@ -9,6 +9,10 @@ import { EventService } from '../../shared/event.service';
 import { IModalHost } from '../../space/wizard/models/modal-host';
 import { TenantService } from '../services/tenant.service';
 
+import { forkJoin } from 'rxjs/observable/forkJoin';
+import { of } from 'rxjs/observable/of';
```
joshuawilson: This should just be 
```
import { forkJoin, of } from 'rxjs';
```

andrewazores:
I recall something similar with imports being done in the past (I want to say by Dana?) since it would help with tree shaking, since Webpack would only be importing the parts of the library that are specifically used. Is that something that the newer version of Webpack is able to handle more intelligently now? Or is it just not a valid concern?

PR #3314, opened 2018-09-12
fix(deployments): fix filtering of deployment cards

jiekang:
This addresses https://github.com/openshiftio/openshift.io/issues/4317

A test is added to verify the changes.

### DIFF
--- a/src/app/space/create/deployments/apps/deployments-apps.component.spec.ts
+++ b/src/app/space/create/deployments/apps/deployments-apps.component.spec.ts
@@ -14,7 +14,7 @@ import {
   TestContext
 } from 'testing/test-context';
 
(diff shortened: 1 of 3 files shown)


src/app/space/create/deployments/apps/deployments-apps.component.spec.ts
```diff
@@ -14,6 +14,8 @@ import {
   TestContext
 } from 'testing/test-context';
 
+import { Filter, FilterEvent } from 'patternfly-ng/filter';
```
aptmac: Filter is unused and can be removed

jiekang: Nice find. I've removed it as well as another unused field in the component.

PR #3318, opened 2018-09-13
fix(pipelines): initialize sort with default value

jiekang:
This fixes the pipelines page to have a correct initial sort. The sorting code called at the start uses this immediately, but the initial state of the toolbar is ascending, not descending, so it would be in the wrong order.


alien-ike:
### Ike Plugins (test-keeper)

PR #3320, opened 2018-09-13
fix(context-service): reduce number of calls to /api/users regarding recent contexts

aptmac:
This PR is a follow up to #3211[[0]](https://github.com/fabric8-ui/fabric8-ui/pull/3211), and resolves [OSIO defect #157](https://openshift.io/openshiftio/Openshift_io/plan/detail/157).

The main mechanism for populating `_recent` in the Context Service is similar to how it is done in the Spaces Service, so much of the core of this PR mirrors the changes from #3211[0]. While making these changes, I went and cleaned up some of the code (unused functions, missing types, etc.), and removed the no longer necessary event service that was previously used to handle deleted spaces on the context service side (now replaced with a broadcast). 

When visiting the home page (with fully loaded recent spaces & recent contexts), here is what the network tab looks like before #3211 [0]
before

and after this PR.
after

[0] https://github.com/fabric8-ui/fabric8-ui/pull/3211
[1] https://openshift.io/openshiftio/Openshift_io/plan/detail/157


andrewazores:
Since the ContextService and SpaceService both behave so similarly and are implemented similarly, would it be possible to refactor them to both extend a common generic base class to provide the "recents" functionality?

andrewazores:
Sure, doing it as a separate enhancement PR after the fact is fine too.

FWIW I don't really agree with this `RecentDataService` approach, and having the `SpacesService` and `ContextService` inject a shared instance which backs and maintains both of their (now shared) states. I was thinking of an abstract utility class that both `SpacesService` and `ContextService` would subclass, so that they implement the same behaviour the same way, and can simply take advantage of generic data types to differentiate between recent Spaces vs recent Contexts (ie `Subject<T>` in the abstract superclass could be `Subject<Context>` or `Subject<Space>`).

aptmac:
> FWIW I don't really agree with this `RecentDataService` approach, and having the `SpacesService` and `ContextService` inject a shared instance which backs and maintains both of their (now shared) states. I was thinking of an abstract utility class that both `SpacesService` and `ContextService` would subclass, so that they implement the same behaviour the same way, and can simply take advantage of generic data types to differentiate between recent Spaces vs recent Contexts (ie `Subject<T>` in the abstract superclass could be `Subject<Context>` or `Subject<Space>`).

Ah I see, I can take a look into that approach.

> Is this PR ready for review now since the WIP label has been removed?

Yes it should be okay for review now.

src/app/shared/context.service.ts
```diff
@@ -51,107 +46,123 @@ export class ContextService implements Contexts {
...
-            val.length - this.RECENT_CONTEXT_LENGTH
-          );
+  initDefault(): void {
+   // Initialize the default context when the logged in user changes
+   this._default = this.userService.loggedInUser
+    .pipe(
+    // First use map to convert the broadcast event to just a username
+      map((val: User): string => {
+        if (!(val && val.id)) {
+          // this is a logout event
```
andrewazores: Should there be an explicit return here?

aptmac: Yeah it'd probably be best to return something here. Since there is no valid username at this point, I've made it return a null that gets filtered out before continuing with the function.

aptmac:
> As a general comment, I would prefer to see Subscriptions properly managed in the ContextService. Right now there are dangling Subscribers that have no path to being unsubscribed and cleaned up. This is okay by coincidence of the fact that the ContextService is global and has the same lifecycle as the entire application - but it would lead to a resource leak in the future if the ContextService ever had its lifecycle narrowed. For that reason I'd like to see the Subscriptions tracked and freed in an ngOnDestroy, even if at the moment that hook will never be called in practice.

Yes that sounds fair, I've added a list of subscriptions similar to how subscriptions are managed elsewhere in the codebase, and unsubscribe from them in an `ngOnDestroy`.

PR #3322, opened 2018-09-14
fix(SpacesService): add check for length of recentSpaces in loadRecent()

aptmac:
This PR addresses an issue that is causing E2E tests to fail where the user has no empty spaces.

The problem was the check of the profile's recent spaces within `loadRecent()`. If the contextInformation on the profile has an empty list for the recentSpaces, then the if-statement takes the branch that will try to forkJoin and will never return. Trying to create spaces or view other spaces will result in an empty list, because the initial list was never populated. The fix here is to continue with fetching spaces by id only if recentSpaces exists and is non-empty.


alien-ike:
### Ike Plugins (test-keeper)

aptmac:
It looks like forkJoin requires a value to be emitted in order for it to complete. In this case, even if on an empty list is seems that forkJoin short-circuits because nothing gets returned.

Related discussion: https://github.com/ReactiveX/rxjs/issues/2816

ppitonak:
When do you plan to merge this PR? Almost all Jenkins jobs are failing because of this bug.

ibuziuk:
@joshuawilson would it be possible to merge this PR since currently che related tests before promotion to prod are also broken  - https://github.com/openshiftio/saas-openshiftio/pull/1063

PR #3324, opened 2018-09-18
fix(app-launcher): fix github linking due to extra headers

rohitkrai03:
Fixes https://github.com/openshiftio/openshift.io/issues/4333
Fixes https://github.com/fabric8-services/fabric8-auth/issues/653

- Due to extra headers being added in github linking request, the request was failing in CORS preflight request.
- Uses `ProviderService` for linking github account instead of custom implementation.


corinnekrych:
@rohitkrai03 could that be a potential issue: https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/space/app-launcher/services/app-launcher-dependency-editor.service.ts#L46 too? or it's only for external git provider?

rohitkrai03:
@corinnekrych This was just an issue with git provider. `getCoreDependencies` method calls the recommender api url which i think needs that header.

PR #3326, opened 2018-09-18
fix(webpack): fix errors with loading css due to extract text plugin

christianvogt:
After fixing the space-name directive errors, runtime errors related to css were showing up.
Tracked down an issue with extract text plugin and webpack 4. Implemented their fix.

We should probably revisit the extract text plugin but for now this gets us one step closer.


alien-ike:
### Ike Plugins (test-keeper)

PR #3329, opened 2018-09-19
style(deployments): code cleanup

andrewazores:
Add type annotations, use const where appropriate, do not leave
subscriptions dangling in tests, and other code quality improvements.

### DIFF
--- a/src/app/space/create/deployments/apps/delete-deployment-modal.component.spec.ts
+++ b/src/app/space/create/deployments/apps/delete-deployment-modal.component.spec.ts
@@ -1,14 +1,8 @@
 import { Component } from '@angular/core';
-import {
-  ModalDirective,
-  ModalModule
-} from 'ngx-bootstrap/modal';
+import { ModalModule } from 'ngx-bootstrap/modal';
 import { first } from 'rxjs/operators';
-import { createMock } from 'testing/mock';
-import {
-  initContext,
-  TestContext
-} from 'testing/test-context';
+
+import { initContext } from 'testing/test-context';
 import { DeleteDeploymentModal } from './delete-deployment-modal.component';
 
 @Component({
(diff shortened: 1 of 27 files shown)


andrewazores:
Huh. The CI bot has a test failure like this:

```
Summary of all failing tests
FAIL src/app/space/create/deployments/deployments-donut/deployments-donut-chart/deployments-donut-chart.component.spec.ts (11.486s)
  ● DeploymentsDonutChartComponent › Mini chart › should show mini text

    TypeError: n.querySelector is not a function

      155 |     }
      156 | 
    > 157 |     donutChartTitle = d3.select(this.chart.element).select('text.c3-chart-arcs-title');
          |                                                     ^
      158 |     if (!donutChartTitle) {
      159 |       return;
      160 |     }

```

but I don't see this locally.

aptmac:
> Huh. The CI bot has a test failure [...] but I don't see this locally.

Neat! I see the same thing; local tests complete and pass as expected.

andrewazores:
Oh right - I had those import changes at the back of my mind while working through the rebase, and then forgot about them at the end before pushing. Thanks, I'll go back for them now. Good catch.

aptmac:
The test failure may be a result of an async issue somewhere, there's something similar reported at https://github.com/wallabyjs/public/issues/808

christianvogt:
There are some intermittent test failures due to async. We can skip them for now and address them later. We skipped two yesterday and this one has the same error as one I've seen before.

andrewazores:
@christianvogt is it just that Deployment Donut component test you're talking about? I can `x` that one out as part of this PR if you're like.

christianvogt:
@andrewazores Yes I was talking about that one. Go ahead and `x` it out. We already `x`'d out one other test in that file yesterday.

PR #3330, opened 2018-09-19
style(settings): code cleanup

andrewazores:
General code style cleanup and improvements. Mostly cosmetic, with some improvements to Subscription handling

### DIFF
--- a/src/app/profile/settings/connected-accounts/connected-accounts.component.ts
+++ b/src/app/profile/settings/connected-accounts/connected-accounts.component.ts
@@ -1,7 +1,7 @@
 import { Component, OnDestroy, OnInit, ViewEncapsulation } from '@angular/core';
 import { Context, Contexts } from 'ngx-fabric8-wit';
 import { AuthenticationService } from 'ngx-login-client';
-import { UserService } from 'ngx-login-client';
+import { User, UserService } from 'ngx-login-client';
 import { Subscription } from 'rxjs';
 import { map } from 'rxjs/operators';
 import { ProviderService } from '../../../shared/account/provider.service';
@@ -15,7 +15,6 @@ import { TenantService } from '../../services/tenant.service';
 })
 export class ConnectedAccountsComponent implements OnDestroy, OnInit {
   context: Context;
-  subscriptions: Subscription[] = [];
 
   gitHubLinked: boolean = false;
   gitHubUserName: string;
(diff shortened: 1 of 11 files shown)


alien-ike:
### Ike Plugins (test-keeper)

joshuawilson:
Would you be willing to apply these changes to the `upgrade` branch. I think some of them would be lost in a merge or it will just be a hard merge as the imports have changes.

src/app/profile/settings/connected-accounts/connected-accounts.component.ts
```diff
@@ -26,30 +25,32 @@ export class ConnectedAccountsComponent implements OnDestroy, OnInit {
...
 
-  constructor(private contexts: Contexts,
-    private auth: AuthenticationService,
-    private userService: UserService,
-    private providerService: ProviderService) {
-    this.subscriptions.push(auth.gitHubToken.subscribe(token => {
+  private readonly subscriptions: Subscription[] = [];
+
+  constructor(
+    contexts: Contexts,
```
aptmac: out of curiosity, why the change away from private?

src/app/profile/settings/connected-accounts/connected-accounts.spec.ts
```diff
@@ -2,8 +2,8 @@ import { Component } from '@angular/core';
 import { Contexts } from 'ngx-fabric8-wit';
 import { AuthenticationService } from 'ngx-login-client';
 import { UserService } from 'ngx-login-client';
-import { empty as observableEmpty, Observable,  of ,  throwError as observableThrowError } from 'rxjs';
-import { initContext, TestContext } from 'testing/test-context';
+import { empty as observableEmpty, of , throwError } from 'rxjs';
```
aptmac: similar to the changes in the deployments PR, should `empty as observableEmpty` simply become `empty`?

joshuawilson: If you do make that switch use the upper case EMPTY.

src/app/profile/settings/connected-accounts/connected-accounts.component.ts
andrewazores: The `private` keyword there makes the parameters become private members of the class, but these parameters were only referenced directly within the constructor body and nowehere else in the class, so there is no need to make them class members.

PR #3336, opened 2018-09-24
feat(deployments): allow scaling only if remaining quota is sufficient

andrewazores:
~~WIP: DeploymentApiService currently returns a mocked data structure. This needs to be replaced with an actual HTTP request once the backend support (https://github.com/fabric8-services/fabric8-wit/pull/2286) is merged and deployed. Additionally, there should be another piece of logic added on top to only allow a scale-up attempt for the maximum number of pods that can be supported in the remaining quota, so the user cannot over-request (ie attempt to scale 0->3 when there is only quota for 2).~~

Allow users to scale up deployments only if the remaining quota is sufficient to support an
additional pod. Previously, if any free quota remained a scale operation would always be permitted,
even if the amount of remaining quota was too small to support an additional pod.

related to https://github.com/openshiftio/openshift.io/issues/3389

### DIFF
--- a/src/app/space/create/deployments/deployments-donut/deployments-donut.component.html
+++ b/src/app/space/create/deployments/deployments-donut/deployments-donut.component.html
@@ -2,7 +2,7 @@
 </deployments-donut-chart>
 <div column class="deployments-donut-scale-controls fade-inline" *ngIf="!isIdled && !mini">
   <div>
-    <a id="scaleUp" title="Scale up" role="button" (click)="$event.preventDefault(); scaleUp()" *ngIf="!atQuota; else scaleUpDisabled">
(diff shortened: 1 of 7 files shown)


andrewazores:
This PR is essentially ready, but should not be merged until after https://github.com/fabric8-services/fabric8-wit/pull/2286 is merged and ready in production. The final commit in this series contains commented code and disabled tests to switch on backend querying for this feature.

andrewazores:
@aptmac I added a commit which enables the actual HTTP query, but uses the previous "mock" (actually matches the real backend default allocation) as a fallback value in case of HTTP errors. The backend is still not ready, even in prod-preview, so this hardening also makes this commit safe to merge in advance of the backend becoming available. Please re-review the last commit(s).

aptmac:
Great, I'll take a look at it now.

PR #3338, opened 2018-09-25
fix(work-item-widget): fix logout bug

rohitkrai03:
Fixes https://github.com/openshiftio/openshift.io/issues/4348

- This comment explains the issue - https://github.com/openshiftio/openshift.io/issues/4348#issuecomment-424018724

- Use already present `loggedInUser` to get `user.id` so no need to subscribe to `userService.getUser()`

### DIFF
--- a/src/app/home/work-item-widget/work-item-widget.component.spec.ts
+++ b/src/app/home/work-item-widget/work-item-widget.component.spec.ts
@@ -93,7 +93,7 @@ describe('Home: WorkItemWidgetComponent', () => {
       { provide: UserService, useFactory: () => {
           let userService = createMock(UserService);
           userService.getUser.and.returnValue(fakeUser);
-          userService.loggedInUser = fakeUser.publish() as ConnectableObservable<User> & jasmine.Spy;
+          userService.loggedInUser = fakeUser as ConnectableObservable<User> & jasmine.Spy;
           return userService;
         }
       }, {
--- a/src/app/home/work-item-widget/work-item-widget.component.ts
+++ b/src/app/home/work-item-widget/work-item-widget.component.ts
@@ -18,7 +18,6 @@ export class WorkItemWidgetComponent implements OnDestroy, OnInit  {
   currentSpaceId: string = 'default';
   loading: boolean = false;
   loggedInUser: User;
-  loggedInUserId: string;
   recentSpaces: Space[] = [];
   recentSpaceIndex: number = 0;
   subscriptions: Subscription[] = [];
@@ -31,7 +30,6 @@ export class WorkItemWidgetComponent implements OnDestroy, OnInit  {
       private userService: UserService) {
     this.subscriptions.push(userService.loggedInUser.subscribe(user => {
       this.loggedInUser = user;
-      this.loggedInUserId = user.id;
     }));
     this.subscriptions.push(spacesService.recent.subscribe(spaces => {
       this.recentSpaces = spaces;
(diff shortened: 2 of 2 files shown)


alien-ike:
### Ike Plugins (test-keeper)

rohitkrai03:
@corinnekrych Can you review this one? Its a sev-1

aptmac:
Aside from the tests, this looks good to go.

rohitkrai03:
@aptmac @corinnekrych I've pushed changes which fixes all the unit tests for the component.

src/app/home/work-item-widget/work-item-widget.component.ts
```diff
@@ -18,6 +18,7 @@ export class WorkItemWidgetComponent implements OnDestroy, OnInit  {
   currentSpaceId: string = 'default';
   loading: boolean = false;
   loggedInUser: User;
+  loggedInUserId: string;
```
aptmac: This looks redundant to me because `loggedInUser` is now no longer used for anything. It would make more sense to me if we used the `loggedInUser.id` value elsewhere in the component as opposed to subscribing to the `UserService` to retrieve only an `id` value.

There seems to be a problem in the `.spec` file hooking up the `fakeUser` to the mocked user service .. removing the `publish()` here [[0]](https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/home/work-item-widget/work-item-widget.component.spec.ts#L96) seems to rectify the situation, and we can use the `fakeUser` observable as a connectable observable.

[0] https://github.com/fabric8-ui/fabric8-ui/blob/master/src/app/home/work-item-widget/work-item-widget.component.spec.ts#L96

src/app/home/work-item-widget/work-item-widget.component.ts
```diff
@@ -62,45 +64,44 @@ export class WorkItemWidgetComponent implements OnDestroy, OnInit  {
    */
   private fetchWorkItemsBySpace(space: Space): void {
     this.currentSpace = space;
-    this.subscriptions.push(this.userService
-      .getUser()
-      .do(() => this.loading = true)
-      .do(() => this.workItemService._currentSpace = space)
-      .do(() => this.workItemService.buildUserIdMap())
-      .switchMap(() => this.userService.loggedInUser)
```
aptmac: Actually the more I think about it, would it be possible to bring back the `userService.loggedInUser` into the observable chain, similar to how it is here? There is no guarantee that the constructors subscription to retrieve the `loggedInUser` has completed by the time the queries would be created (and thus `loggedInUser` may not contain a value), and it may be safer to handle it in sequence along with fetching the work items.

rohitkrai03: @aptmac But that was the issue here. After the logout `loggedInUser` was undefined since it was an observable which caused the problem.

aptmac: Hm right, ignore that then haha.

rohitkrai03:
@aptmac @corinnekrych Merging it so that it can be moved to prod soon.

src/app/home/work-item-widget/work-item-widget.component.spec.ts
```diff
@@ -93,7 +93,7 @@ describe('Home: WorkItemWidgetComponent', () => {
       { provide: UserService, useFactory: () => {
           let userService = createMock(UserService);
           userService.getUser.and.returnValue(fakeUser);
-          userService.loggedInUser = fakeUser.publish() as ConnectableObservable<User> & jasmine.Spy;
+          userService.loggedInUser = fakeUser as ConnectableObservable<User> & jasmine.Spy;
```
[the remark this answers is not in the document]
joshuawilson: ok

PR #3341, opened 2018-09-25
feat(recent): introduce recent-utils to be extended by context and spaces service for "recent" logic

aptmac:
~~Note: this PR is meant to be on top of https://github.com/fabric8-ui/fabric8-ui/pull/3320 [[0]](https://github.com/fabric8-ui/fabric8-ui/pull/3320), so I'll keep it a WIP for the time being. Only the last commit in this PR is relevant here.~~

This PR addresses a comment [[1]](https://github.com/fabric8-ui/fabric8-ui/pull/3320#issuecomment-421436050) on [[0]](https://github.com/fabric8-ui/fabric8-ui/pull/3320) regarding extracting common "recent" logic to an extendable common class. Here, `recent-utils.ts` is introduced and takes care of the:
1. declaration and return of `_recent`
2. logic as a result of broadcaster events `'spaceChanged', 'contextChanged', 'spaceDeleted'`
3. saving the patch to the profile

The loading of recent contexts and spaces happen in similar functionality, but not similar enough that it makes sense to combine the two IMO. The recent spaces are just loaded by simply iterating through the `profile.store.recentSpaces` and getting the space by id from the `SpaceService` on each of them, where the context service needs to distinguish between user contexts and space contexts while loading, potentially building a context to contain spaces with after retreiving the space by id from `SpaceService`.

Additionally I made use of the error handler to remove a few instances of `console.log()`, and added a couple of functions that help out the `recent-utils` by determining if the object in question is a space or context. Also, adding a list of subscriptions in the `SpacesService` so they can be unsubscribed from when `ngOnDestroy` gets called, similar to the changes in the `ContextService`.

[0] https://github.com/fabric8-ui/fabric8-ui/pull/3320
[1] https://github.com/fabric8-ui/fabric8-ui/pull/3320#issuecomment-421436050

### DIFF
--- a/src/app/shared/context.service.ts
+++ b/src/app/shared/context.service.ts
@@ -140,7 +140,7 @@ export class ContextService extends RecentUtils<Context> implements Contexts {
         .pipe(
           withLatestFrom(this._recent),
           map(([deletedSpace, recentContexts]: [Space, Context[]]): RecentData<Context> =>
-            this.onBroadcastDeleted(deletedSpace, recentContexts)
+            this.onBroadcastSpaceDeleted(deletedSpace, recentContexts, this.compareContextToSpace)
           ),
           filter((recentData: RecentData<Context>): boolean => recentData.isSaveRequired)
         )
@@ -148,7 +148,20 @@ export class ContextService extends RecentUtils<Context> implements Contexts {
           this.saveRecentContexts(recentData.data)
         )
     );
+  }
 
+  // implements recent-utils compareElements()
+  compareElements(c1: Context, c2: Context): boolean {
+    return c1.name === c2.name;
+  }
+
+  // comparator function for space deletion events
+  compareContextToSpace(c: Context, s: Space): boolean {
+    if (c.space) {
+      return c.space.id === s.id;
+    } else {
+      return false;
+    }
   }
 
   get current(): Observable<Context> {
--- a/src/app/shared/context.service.ts
+++ b/src/app/shared/context.service.ts
@@ -131,7 +131,7 @@ export class ContextService extends RecentUtils<Context> implements Contexts {
           filter((recentData: RecentData<Context>): boolean => recentData.isSaveRequired)
         )
         .subscribe((recentData: RecentData<Context>): void =>
-          this.saveRecentContexts(recentData.data)
+          this.saveRecentContexts(recentData.recent)
         )
     );
 
@@ -145,7 +145,7 @@ export class ContextService extends RecentUtils<Context> implements Contexts {
           filter((recentData: RecentData<Context>): boolean => recentData.isSaveRequired)
         )
         .subscribe((recentData: RecentData<Context>): void =>
-          this.saveRecentContexts(recentData.data)
+          this.saveRecentContexts(recentData.recent)
         )
     );
   }
--- a/src/app/shared/context.service.ts
+++ b/src/app/shared/context.service.ts
@@ -116,37 +116,37 @@ export class ContextService extends RecentUtils<Context> implements Contexts {
 
   private initRecent(): void {
     this.subscriptions.push(
-      this.loadRecentContexts().subscribe((contexts: Context[]): void => {
-        this._recent.next(contexts);
-      })
+      this.loadRecentContexts().subscribe((contexts: Context[]): void =>
+        this._recent.next(contexts)
+      )
     );
 
     this.subscriptions.push(
       this.broadcaster.on<Context>('contextChanged')
         .pipe(
           withLatestFrom(this.recent),
-          map(([changedContext, recentContexts]: [Context, Context[]]): RecentData<Context> => {
-            return this.onBroadcastChanged(changedContext, recentContexts);
-          }),
-          filter((recentData: RecentData<Context>): boolean => recentData.isSaveRequired === true)
-          )
-        .subscribe((recentData: RecentData<Context>): void => {
-          this.saveRecentContexts(recentData.data);
-        })
+          map(([changedContext, recentContexts]: [Context, Context[]]): RecentData<Context> =>
+            this.onBroadcastChanged(changedContext, recentContexts)
+          ),
+          filter((recentData: RecentData<Context>): boolean => recentData.isSaveRequired)
+        )
+        .subscribe((recentData: RecentData<Context>): void =>
+          this.saveRecentContexts(recentData.data)
+        )
     );
 
     this.subscriptions.push(
       this.broadcaster.on<Space>('spaceDeleted')
         .pipe(
           withLatestFrom(this._recent),
-          map(([deletedSpace, recentContexts]: [Space, Context[]]): RecentData<Context> => {
-            return this.onBroadcastDeleted(deletedSpace, recentContexts);
-          }),
-          filter((recentData: RecentData<Context>): boolean => recentData.isSaveRequired === true)
+          map(([deletedSpace, recentContexts]: [Space, Context[]]): RecentData<Context> =>
+            this.onBroadcastDeleted(deletedSpace, recentContexts)
+          ),
+          filter((recentData: RecentData<Context>): boolean => recentData.isSaveRequired)
         )
-        .subscribe((recentData: RecentData<Context>): void => {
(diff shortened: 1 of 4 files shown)


src/app/shared/recent-utils.ts
```diff
@@ -0,0 +1,88 @@
...
+import { Context, Space } from 'ngx-fabric8-wit';
+import { Observable, ReplaySubject, Subject } from 'rxjs';
+import { ExtProfile, ProfileService } from '../profile/profile.service';
+
+export const RECENT_LENGTH: number = 8;
+
+export interface RecentData <T extends Context | Space> {
+  data: T[];
+  isSaveRequired: boolean;
+}
```
andrewazores: Looks good overall, but I'd like to find a way around this `T extends Context | Space` thing - that's not a nice type constraint, because the end result is a very neutered form of "genericism". Ideally we want to make it so that this abstract class never needs to be touched or updated in any way simply to allow for a "Recents Store" of another data type than Spaces or Contexts. This type assertion technique also gets hairy if different datatypes end up with a lot of properties with the same name. Specific thoughts on potential solutions below on related bits of code.

src/app/shared/recent-utils.ts
```diff
@@ -0,0 +1,88 @@
...
+    return this._recent;
+  }
+
+  onBroadcastChanged(changed: T, recent: T[]): RecentData<T> {
+    let index: number = 0;
+    if (isContext(changed)) {
+      index = (recent as Context[]).findIndex((context: Context) => context.name === (changed as Context).name);
+    } else if (isSpace(changed)) {
+      index = (recent as Space[]).findIndex((space: Space): boolean => space.id === (changed as Space).id);
+    }
```
andrewazores: Rather than having the abstract class handle this logic by doing a type assertion, could we have the subclasses provide a conversion function to the superclass (via constructor) to map the concrete type to an agnostic identifier/index? ie the subclasses would look like:

```
constructor(
  errorHandler: ErrorHandler,
  profileService: ProfileService
  ... etc
) {
  super(errorHandler, profileService, (ctx: Context): string => ctx.name));
}
```

and


```
constructor(
  errorHandler: ErrorHandler,
  profileService: ProfileService
  ... etc
) {
  super(errorHandler, profileService, (space: Space): string => space.id));
}
```

so that the type signature of this adapter/index function is simply `T => string`.

`onBroadcastChanged` would then start off like:

```
onBroadcastChanged(changed: T, recent: T[]): RecentData<T> {
  let index: number = this.indexFn(changed);
```

src/app/shared/recent-utils.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (recent.length !== 0 && isContext(recent[0])) {
+      index = (recent as Context[]).findIndex((context: Context): boolean => {
+        if (context.space) {
+          return context.space.id === deletedSpace.id;
+        } else {
+          return false;
+        }
+      });
+    } else if (recent.length !== 0 && isSpace(recent[0])) {
+      index = (recent as Space[]).findIndex((space: Space): boolean => space.id === deletedSpace.id);
```
andrewazores: Same as above, this type check could be replaced with an index function invocation.

src/app/shared/recent-utils.ts
```diff
@@ -0,0 +1,88 @@
...
+  onBroadcastChanged(changed: T, recent: T[]): RecentData<T> {
+    let index: number = 0;
+    if (isContext(changed)) {
+      index = (recent as Context[]).findIndex((context: Context) => context.name === (changed as Context).name);
+    } else if (isSpace(changed)) {
+      index = (recent as Space[]).findIndex((space: Space): boolean => space.id === (changed as Space).id);
+    }
+
+    if (index === 0) { // continue only if changed is new, or requires a move within recent
+      return { data: recent, isSaveRequired: false } as RecentData<T>;
```
andrewazores: Is the `as` needed?

src/app/shared/recent-utils.ts
```diff
@@ -0,0 +1,88 @@
+import { ErrorHandler } from '@angular/core';
+import { Context, Space } from 'ngx-fabric8-wit';
+import { Observable, ReplaySubject, Subject } from 'rxjs';
+import { ExtProfile, ProfileService } from '../profile/profile.service';
+
+export const RECENT_LENGTH: number = 8;
+
+export interface RecentData <T extends Context | Space> {
+  data: T[];
```
andrewazores: I wonder if maybe `recents` or something is a better name than `data`.

src/app/shared/context.service.ts
```diff
@@ -113,75 +114,41 @@ export class ContextService implements Contexts {
...
-      })
-      .subscribe((recentContexts: Context[]): void => {
-        this.saveRecent(recentContexts);
-      })
+        .pipe(
+          withLatestFrom(this._recent),
+          map(([deletedSpace, recentContexts]: [Space, Context[]]): RecentData<Context> => {
+            return this.onBroadcastDeleted(deletedSpace, recentContexts);
+          }),
+          filter((recentData: RecentData<Context>): boolean => recentData.isSaveRequired === true)
```
andrewazores: `=== true` is redundant

src/app/shared/context.service.ts
```diff
@@ -113,75 +114,41 @@ export class ContextService implements Contexts {
...
-        recentContexts.splice(index, 1);
-        return recentContexts;
-      })
-      .subscribe((recentContexts: Context[]): void => {
-        this.saveRecent(recentContexts);
-      })
+        .pipe(
+          withLatestFrom(this._recent),
+          map(([deletedSpace, recentContexts]: [Space, Context[]]): RecentData<Context> => {
+            return this.onBroadcastDeleted(deletedSpace, recentContexts);
```
andrewazores: Just a style thing but I would drop the curly braces and `return` here, personally.

src/app/shared/context.service.ts
```diff
@@ -392,7 +359,7 @@ export class ContextService implements Contexts {
     });
   }
 
-  private saveRecent(recent: Context[]): void {
+  private saveRecentContexts(recent: Context[]) {
```
andrewazores: What happened to `: void`?

src/app/shared/context.service.ts
```diff
@@ -113,75 +114,41 @@ export class ContextService implements Contexts {
...
-      })
+        .pipe(
+          withLatestFrom(this._recent),
+          map(([deletedSpace, recentContexts]: [Space, Context[]]): RecentData<Context> => {
+            return this.onBroadcastDeleted(deletedSpace, recentContexts);
+          }),
+          filter((recentData: RecentData<Context>): boolean => recentData.isSaveRequired === true)
+        )
+        .subscribe((recentData: RecentData<Context>): void => {
+          this.saveRecentContexts(recentData.data);
```
andrewazores: Same re. dropping curly braces

andrewazores:
#3320 is merged, please rebase and remove WIP

src/app/shared/spaces.service.ts
```diff
@@ -92,7 +92,7 @@ export class SpacesService extends RecentUtils<Space> implements Spaces {
           return this.spaceService.getSpaceById(id).catch((): Observable<Space> => Observable.of(null));
         }))
         .map((spaces: Space[]): Space[] => {
-          return spaces.filter(((space: Space): boolean => { return space !== null; }));
+          return spaces.filter(((space: Space): boolean => space !== null));
```
andrewazores: There's still an outer level of braces that can be removed here :)

src/app/shared/recent-utils.ts
aptmac: That sounds fair, I've taken your suggested route with adding a function `compareElements` to be implemented by the subclass for use in `findIndex`. The only problem I ran into is that the Context Service needs to respond to space deletion events, so in this case it isn't enough to have a comparator for type `T`, because the first element would be a space and the second would be a context. What I've done here is added an additional comparator function for the context service that will be passed into `onBroadcastSpaceDeleted`, which defaults to the implemented `compareElements` if a function is not passed to it.

PR #3345, opened 2018-09-27
cleanup(imports): remove unused imports after upgrade

rohitkrai03:
- Just a few cleanup of imports after upgrade rebase


src/app/layout/header/header.component.spec.ts
```diff
@@ -55,7 +55,7 @@ describe('HeaderComponent', () => {
   let fixture: ComponentFixture<HeaderComponent>;
   let component: DebugNode['componentInstance'];
 
-  const testContext = initContext(HeaderComponent, HostComponent, {
+  const testContext: TestingContext = initContext(HeaderComponent, HostComponent, {
```
joshuawilson: Does this need to be something like `TestContext<LandingPageComponent, HostComponent>`

rohitkrai03: It is equivalent to that only. There is a type definition in this file above like this `type TestingContext = TestContext<HeaderComponent, HostComponent>;`

joshuawilson: I was not trying to say exactly what to add, just referencing the other one. Doesn't it need it here too?

rohitkrai03: Maybe i was not clear earlier, we do need type for `const testContext` and `TestingContext` is a type defined earlier in the test file at https://github.com/fabric8-ui/fabric8-ui/blob/05179d1eb83baf6b6bce502282854b53999d25a1/src/app/layout/header/header.component.spec.ts#L49. 

So,
```
type TestingContext = TestContext<HeaderComponent, HostComponent>;
const testContext: TestingContext = initContext(HeaderComponent, HostComponent, {
```

is equivalent to 
```
const testContext: TestContext<HeaderComponent, HostComponent> = initContext(HeaderComponent, HostComponent, {
```

PR #3348, opened 2018-09-28
fix(css): omit css sourceMap in prod and extract css to separate file

christianvogt:
fixes https://github.com/openshiftio/openshift.io/issues/3826

I could not find a solution to extracting the source maps for css to a separate file. Therefore for now we will disable css source maps in production altogether.
Reduced total JS / CSS asset size from 12.08MB to 6.44MB.

Also re-instated the extraction of css to external file which was disabled during the upgrade due to config errors. This moves a good chunk of css (but not all) to a separate file (~440kb). This fixes the loading screen issue.


alien-ike:
### Ike Plugins (test-keeper)

config/webpack.common.js
```diff
@@ -48,7 +48,7 @@ const METADATA = {
  * See: http://webpack.github.io/docs/configuration.html#cli
  */
 module.exports = function (options) {
-  const isProd = options.env === 'production';
+  const isProd = !!options.isProd;
```
joshuawilson: Why did you change this and how does this work?

joshuawilson: oh, I see you added it at the bottom. Still not sure why you needed this.

config/webpack.common.js
```diff
@@ -254,12 +254,11 @@ module.exports = function (options) {
         {
           test: /^(?!.*component).*\.css$/,
           use: [
-            'style-loader',
-            // isProd ? MiniCssExtractPlugin.loader : 'style-loader',
+            isProd ? MiniCssExtractPlugin.loader : 'style-loader',
```
joshuawilson: Did you confirm this works? I left it commented out as I had issues with it.

config/webpack.common.js
christianvogt: I had to change it to run perf. Since per sets the environment to performance.

Maybe perf should also be production?

config/webpack.common.js
christianvogt: Yes I tested running build:prod and then running a local server.

config/webpack.common.js
joshuawilson: yes, we should set perf to prod since it is a reflection of it

christianvogt: ok let me change the environment instead of adding this option

PR #3357, opened 2018-10-04
feat(monorepo): initial monorepo setup

christianvogt:
Here's the lerna monorepo from my POC with linting and scripts.

@corinnekrych @rohitkrai03 @joshuawilson


joshuawilson:
@christianvogt the Jest tests are failing to run.

PR #2984, opened 2018-06-04
fix(header): Change stylings for subheaders

chrislessard:
This patch changes the styling for the sub navigation bars to align with the new user settings page.

Screenshots : 
2018-06-04-162517_1919x148_scrot

2018-06-04-162502_1919x136_scrot

solves : https://github.com/openshiftio/openshift.io/issues/3163

PR #3364, opened 2018-10-10
fix(header): fix header navbar colouring to be dark

aptmac:
This PR addresses https://github.com/openshiftio/openshift.io/issues/4375 and https://openshift.io/openshiftio/Openshift_io/plan/detail/783, where the header navbar should be made to be a darker colour.

It essentially just reverts an older commit [[0]](https://github.com/fabric8-ui/fabric8-ui/pull/2984) to make the navbar lighter, and here are some screenshots I took:

Base navbar:
2018-10-10-131442_263x77_scrot

Base navbar with hover:
2018-10-10-131458_1916x1033_scrot 1

Planner navbar:
2018-10-10-131626_319x516_scrot

[0] https://github.com/fabric8-ui/fabric8-ui/pull/2984

### DIFF
--- a/src/app/layout/header/header.component.less
+++ b/src/app/layout/header/header.component.less
@@ -72,22 +72,21 @@
     > .active > a {
       background-color: @color-pf-black-900;
       background-image: linear-gradient(to top, @color-pf-black-700 0, @color-pf-black-900 100%);
+      border-bottom-color: @color-pf-black-900;
     }
     &.persistent-secondary {
-      //background-color: @color-pf-black-150;
+      background: none;
       > li {
         &.active {
           &.with-no-children { margin-bottom: 0; }
         }
         .navbar-persistent {
-          background-color: @color-pf-black-150;
-          > li:hover:before {
-            background: none;
-          }
+          border-bottom: 1px solid;
+          border-bottom-color: @color-pf-black;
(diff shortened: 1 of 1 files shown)


catrobson:
@aptmac can you take a screenshot of what this will look like now on the Space Settings or User Profile page?  I think the same secondary navigation component was used there but we didn't want that to turn dark for those situations.

joshuawilson:
collaborators openshift io
profile settings - openshift io

aptmac:
> @aptmac can you take a screenshot of what this will look like now on the Space Settings or User Profile page? I think the same secondary navigation component was used there but we didn't want that to turn dark for those situations.

Ah yeah I had overlooked that bit, I'll fix that up.

joshuawilson:
@catrobson the changes that @aptmac did match the invision design. https://redhat.invisionapp.com/share/MBOEF1JX4VD#/screens/323634780

Or am I missing something?

aptmac:
The dark colour has been applied to header component navbar, so it will leave the settings navbar untouched (it will remain light).

catrobson:
Your screenshots @aptmac follow what I would expect - the navigation is light ONLY if it is below a dark header area like shown on the User Settings page.  Otherwise, if tight against primary navigation, it is dark.  Thanks!

PR #3365, opened 2018-10-10
fix(space-dashboard): fix work-item filter query in create work-item widget

divyanshiGupta:
fixes https://github.com/openshiftio/openshift.io/issues/4381

In create work-item widget work-item filter query was not using space id to get the work-items for that particular space only, so in this PR I have-

- created filter query using space id to solve the above issue
- removed deprecated functions form home work-item-widget

### DIFF
--- a/src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.spec.ts
+++ b/src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.spec.ts
@@ -60,7 +60,7 @@ describe('CreateWorkItemWidgetComponent', () => {
         { provide: Broadcaster, useValue: createMock(Broadcaster) },
         { provide: Contexts, useValue: ({ current: new Subject<Context>() }) },
         {
-          provide: WorkItemService, useFactory: () => {
+          provide: WorkItemService, useFactory: (): WorkItemService => {
             let workItemServiceMock = jasmine.createSpyObj('WorkItemService', ['resolveType', 'resolveAreaForWorkItem', 'getWorkItems']);
             workItemServiceMock.resolveType.and.stub();
             workItemServiceMock.resolveAreaForWorkItem.and.stub();
@@ -69,7 +69,7 @@ describe('CreateWorkItemWidgetComponent', () => {
           }
         },
         {
-          provide: FilterService, useFactory: () => {
+          provide: FilterService, useFactory: (): FilterService => {
             let filterServiceMock: jasmine.SpyObj<FilterService> = jasmine.createSpyObj('FilterService', ['queryBuilder', 'queryJoiner']);
             return filterServiceMock;
           }
@@ -164,7 +164,7 @@ describe('CreateWorkItemWidgetComponent', () => {
         { provide: Broadcaster, useValue: createMock(Broadcaster) },
         { provide: Contexts, useValue: ({ current: new Subject<Context>() }) },
         {
-          provide: WorkItemService, useFactory: () => {
+          provide: WorkItemService, useFactory: (): WorkItemService => {
             let workItemServiceMock = jasmine.createSpyObj('WorkItemService', ['resolveType', 'resolveAreaForWorkItem', 'getWorkItems']);
             workItemServiceMock.resolveType.and.stub();
             workItemServiceMock.resolveAreaForWorkItem.and.stub();
@@ -173,7 +173,7 @@ describe('CreateWorkItemWidgetComponent', () => {
           }
         },
         {
-          provide: FilterService, useFactory: () => {
+          provide: FilterService, useFactory: (): FilterService => {
             let filterServiceMock: jasmine.SpyObj<FilterService> = jasmine.createSpyObj('FilterService', ['queryBuilder', 'queryJoiner']);
             return filterServiceMock;
           }
@@ -197,7 +197,7 @@ describe('CreateWorkItemWidgetComponent', () => {
       ]
     });
 
-    it('should fetch the correct work items', async(function() {
+    it('should fetch the correct work items', async(() => {
       testContext.testedDirective.myWorkItems = observableOf([]);
       testContext.testedDirective.fetchWorkItems();
       testContext.testedDirective.myWorkItems.subscribe((workItems: WorkItem[]) => {
--- a/src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.spec.ts
+++ b/src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.spec.ts
@@ -5,7 +5,7 @@ import { ActivatedRoute, Router, RouterModule } from '@angular/router';
 import { NgArrayPipesModule } from 'angular-pipes';
 import { FilterService, WorkItem, WorkItemService } from 'fabric8-planner';
 import { Broadcaster } from 'ngx-base';
-import { Context, Contexts } from 'ngx-fabric8-wit';
+import { Context, Contexts, Space } from 'ngx-fabric8-wit';
 import { User, UserService } from 'ngx-login-client';
 import { ConnectableObservable,  Observable ,  of as observableOf, Subject } from 'rxjs';
 import { publish } from 'rxjs/operators';
@@ -19,16 +19,21 @@ import { WorkItemsData } from '../../shared/workitem-utils';
 import { CreateWorkItemWidgetComponent } from './create-work-item-widget.component';
 
 @Component({
-  template: '<fabric8-create-work-item-widget [userOwnsSpace]="userOwnsSpace"></fabric8-create-work-item-widget>'
+  template: `
+    <fabric8-create-work-item-widget
+      [userOwnsSpace]="userOwnsSpace"
+      [currentSpace]="space"
+      [loggedInUser]="loggedInUser">
+    </fabric8-create-work-item-widget>
+  `
 })
 class HostComponent {
   userOwnsSpace: boolean;
-}
-
-describe('CreateWorkItemWidgetComponent', () => {
-  type TestingContext = TestContext<CreateWorkItemWidgetComponent, HostComponent>;
-
-  let fakeUser: Observable<User> = observableOf({
+  space: Space = {
+    attributes: {},
+    id: 'some-space-id'
+  } as Space;
+  loggedInUser: User = {
     id: 'fakeId',
     type: 'fakeType',
     attributes: {
@@ -36,7 +41,11 @@ describe('CreateWorkItemWidgetComponent', () => {
       imageURL: 'null',
       username: 'fakeUserName'
     }
-  } as User);
+  };
+}
+
+describe('CreateWorkItemWidgetComponent', () => {
+  type TestingContext = TestContext<CreateWorkItemWidgetComponent, HostComponent>;
 
   const testContext = initContext(CreateWorkItemWidgetComponent, HostComponent, {
     declarations: [ MockFeatureToggleComponent ],
@@ -47,24 +56,15 @@ describe('CreateWorkItemWidgetComponent', () => {
       { provide: Broadcaster, useValue: createMock(Broadcaster) },
       { provide: Contexts, useValue: ({ current: new Subject<Context>() }) },
       {
-        provide: UserService, useFactory: () => {
-          let userService = createMock(UserService);
-          userService.getUser.and.returnValue(fakeUser);
-          userService.loggedInUser = fakeUser.pipe(publish()) as ConnectableObservable<User> & jasmine.Spy;
-          return userService;
-        }
-      },
-      {
         provide: WorkItemService, useFactory: () => {
           let workItemServiceMock = jasmine.createSpyObj('WorkItemService', ['buildUserIdMap', 'getWorkItems']);
-          workItemServiceMock.buildUserIdMap.and.returnValue(fakeUser);
           workItemServiceMock.getWorkItems.and.returnValue(observableOf({ workItems: [] }) as Observable<WorkItemsData>);
           return workItemServiceMock;
         }
       },
       {
         provide: FilterService, useFactory: () => {
-          let filterServiceMock = jasmine.createSpyObj('FilterService', ['queryBuilder']);
+          let filterServiceMock = jasmine.createSpyObj('FilterService', ['queryBuilder', 'queryJoiner']);
           return filterServiceMock;
         }
       },
@@ -89,7 +89,7 @@ describe('CreateWorkItemWidgetComponent', () => {
 
   it('should enable buttons if the user owns the space', function() {
     testContext.hostComponent.userOwnsSpace = true;
-    testContext.testedDirective.myWorkItemsCount = observableOf(0);
+    testContext.testedDirective.myWorkItemsCount = 0;
     testContext.detectChanges();
 
     expect(testContext.fixture.debugElement.query(By.css('#spacehome-my-workitems-add-button'))).not.toBeNull();
@@ -99,7 +99,7 @@ describe('CreateWorkItemWidgetComponent', () => {
 
   it('should disable buttons if the user does not own the space', function() {
     testContext.hostComponent.userOwnsSpace = false;
-    testContext.testedDirective.myWorkItemsCount = observableOf(0);
+    testContext.testedDirective.myWorkItemsCount = 0;
     testContext.detectChanges();
 
     expect(testContext.fixture.debugElement.query(By.css('#spacehome-my-workitems-add-button'))).toBeNull();
--- a/src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.spec.ts
+++ b/src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.spec.ts
@@ -3,12 +3,12 @@ import { Component, NO_ERRORS_SCHEMA } from '@angular/core';
 import { By } from '@angular/platform-browser';
 import { ActivatedRoute, Router, RouterModule } from '@angular/router';
 import { NgArrayPipesModule } from 'angular-pipes';
-import { FilterService, WorkItem, WorkItemService } from 'fabric8-planner';
+import { FilterService, WorkItemService } from 'fabric8-planner';
+import { NgLetModule } from 'fabric8-planner/app/shared/ng-let';
 import { Broadcaster } from 'ngx-base';
 import { Context, Contexts, Space } from 'ngx-fabric8-wit';
-import { User, UserService } from 'ngx-login-client';
-import { ConnectableObservable,  Observable ,  of as observableOf, Subject } from 'rxjs';
-import { publish } from 'rxjs/operators';
+import { User } from 'ngx-login-client';
+import { Observable, of as observableOf, Subject } from 'rxjs';
 import { createMock } from 'testing/mock';
 import { MockFeatureToggleComponent } from 'testing/mock-feature-toggle.component';
 import {
@@ -47,9 +47,9 @@ class HostComponent {
 describe('CreateWorkItemWidgetComponent', () => {
   type TestingContext = TestContext<CreateWorkItemWidgetComponent, HostComponent>;
 
-  const testContext = initContext(CreateWorkItemWidgetComponent, HostComponent, {
+  const testContext: TestingContext = initContext(CreateWorkItemWidgetComponent, HostComponent, {
     declarations: [ MockFeatureToggleComponent ],
-    imports: [NgArrayPipesModule, RouterModule],
+    imports: [NgArrayPipesModule, RouterModule, NgLetModule],
     providers: [
       { provide: ActivatedRoute, useValue: jasmine.createSpy('ActivatedRoute') },
       { provide: LocationStrategy, useValue: jasmine.createSpyObj('LocationStrategy', ['prepareExternalUrl']) },
@@ -64,7 +64,7 @@ describe('CreateWorkItemWidgetComponent', () => {
       },
       {
         provide: FilterService, useFactory: () => {
-          let filterServiceMock = jasmine.createSpyObj('FilterService', ['queryBuilder', 'queryJoiner']);
+          let filterServiceMock: jasmine.SpyObj<FilterService> = jasmine.createSpyObj('FilterService', ['queryBuilder', 'queryJoiner']);
           return filterServiceMock;
         }
       },
@@ -89,7 +89,6 @@ describe('CreateWorkItemWidgetComponent', () => {
 
   it('should enable buttons if the user owns the space', function() {
     testContext.hostComponent.userOwnsSpace = true;
-    testContext.testedDirective.myWorkItemsCount = 0;
     testContext.detectChanges();
 
     expect(testContext.fixture.debugElement.query(By.css('#spacehome-my-workitems-add-button'))).not.toBeNull();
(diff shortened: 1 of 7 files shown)


src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
```diff
@@ -14,41 +14,73 @@ import { filterOutClosedItems, WorkItemsData } from '../../shared/workitem-utils
 export class CreateWorkItemWidgetComponent implements OnInit {
 
   @Input() userOwnsSpace: boolean;
+  @Input() loggedInUser: User;
+  @Input() currentSpace: Space;
 
-  private _myWorkItems: ConnectableObservable<WorkItem[]>;
-  myWorkItemsCount: Observable<number>;
+  private _myWorkItems: WorkItem[];
```
aptmac: I would rather these remain Observables so that the component reacts to the values coming in.

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
```diff
@@ -14,41 +14,73 @@ import { filterOutClosedItems, WorkItemsData } from '../../shared/workitem-utils
...
+          tap(workItems => workItems.forEach(workItem => this.workItemService.resolveAreaForWorkItem(workItem))),
+          tap(workItems => {
+            workItems.forEach(workItem => {
+              if (workItem.relationalData === undefined) {
+                workItem.relationalData = {};
+              }
+            });
+          })
+        ).subscribe((workitems: WorkItem[]) =>  {
+         this._myWorkItems = workitems;
```
aptmac: There is only 1 space indent here.

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
```diff
@@ -1,8 +1,8 @@
 import { Component, Input, OnInit, ViewEncapsulation } from '@angular/core';
 import { FilterService, WorkItem, WorkItemService } from 'fabric8-planner';
-import { Contexts } from 'ngx-fabric8-wit';
-import { UserService } from 'ngx-login-client';
-import { ConnectableObservable, Observable } from 'rxjs';
+import { Contexts, Space } from 'ngx-fabric8-wit';
+import { User, UserService } from 'ngx-login-client';
```
aptmac: UserService isn't used anymore, and there are currently a few other unused imports that can be cleaned up (e.g., ConnectableObservable, publishReplay, switchMap)

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
```diff
@@ -14,41 +14,73 @@ import { filterOutClosedItems, WorkItemsData } from '../../shared/workitem-utils
...
+      'space', this.filterService.equal_notation, this.currentSpace.id
+    );
+    const filters = this.filterService.queryJoiner(
+      assigneeQuery, this.filterService.and_notation, spaceQuery
+    );
+    this.subscriptions.push(
+      this.workItemService
+       .getWorkItems(100000, {expression: filters}).pipe(
+          map((val: WorkItemsData) => val.workItems),
+          map(workItems => filterOutClosedItems(workItems)),
```
aptmac: It'd be appreciated it the type definitions could be added here (and elsewhere in this function chain too), e.g.,
`map(workItems: WorkItem[]) => [...]`

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
divyanshiGupta: @aptmac we do not need observables as this has been taken care of in `ngOnChanges()` that will call `getCurrentSpaceWorkItems()` if any of the input properties changes.

divyanshiGupta:
@aptmac I have two different spaces which i tested this with. One is Agile and one is SDD. It seems to show work items from both of them to me as intended. Can you explain the issue you are facing a bit more?

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
sahil143: @divyanshiGupta I agree with @aptmac.

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
```diff
@@ -14,41 +14,73 @@ import { filterOutClosedItems, WorkItemsData } from '../../shared/workitem-utils
...
+      'space', this.filterService.equal_notation, this.currentSpace.id
+    );
+    const filters = this.filterService.queryJoiner(
+      assigneeQuery, this.filterService.and_notation, spaceQuery
+    );
+    this.subscriptions.push(
+      this.workItemService
+       .getWorkItems(100000, {expression: filters}).pipe(
+          map((val: WorkItemsData) => val.workItems),
+          map((workItems: WorkItem[]) => filterOutClosedItems(workItems)),
```
sahil143: Why are we not adding the filters while making the request for workitems?

divyanshiGupta: What do you mean by adding filters while making request? We are adding filters here. Which filters are you talking about?

sahil143: To filter out the closed work items. We can do that while making the request to the backend.
In this case, we are first fetching the workitems and then filter closed workitem in `filterOutClosedItems` method.

aptmac: That'd be great if this could be done in the request, in follow up PRs this could be ported to the other work-item widgets.

aptmac:
> @aptmac I have two different spaces which i tested this with. One is Agile and one is SDD. It seems to show work items from both of them to me as intended. Can you explain the issue you are facing a bit more?

Whoops! Disregard that I must have unassigned myself from my test work items when testing a PR, it works as you describe, sorry about that.

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
divyanshiGupta: @sahil143 Is there a documentation for query language where i can get info of how to build this query join it with present one?

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
aptmac: > @sahil143 I don't think so using observables for everything is a good solution. There are scenarios where observables might create unnecessary changes in values. If you look at this issue [openshiftio/openshift.io#4348](https://github.com/openshiftio/openshift.io/issues/4348), it was introduced because of using observables when not needed.

I don't recall it being "because" of observables, but rather the log out event changed the loggedInUser value to be `{}` because there was no logged in user. As a result, calling `.trim()` on it somewhere caused an error.

> Regarding _myWorkItems being an Observable, i think it is not necessary here since its value will only be changing when the input properties change which will be handled by change detection and managed by the lifecycle hook.
In my opinion the code earlier was unnecessarily complicated by using ConnectableObservable, publishReplay(1) and _myWorkItems.connect().

I agree the older implementation might be a bit complicated; I recently refactored some of our frequently used services to move away from that type of implementation. Having said that, it needs to be replaced with something similar, and the work items come in as an observable it should be easy enough to keep it that way and let the front end update once the request finishes.

My main concern here is the assumption that subscribing to these observables completes before we need the data. We currently have (& had) some racy issues with services that provide absolute values of observables because of assumptions that it would be ready in time to serve the information. (IIRC, it was the user service providing both a `User` and `Observable<User>` value, and using the former could cause complications if it hadn't returned from the subscription fast enough .. e.g., in the case of slow internet connections).

divyanshiGupta: @aptmac I understand the concern for racy conditions. But, converting `_myWorkItems` as Observable will also need `myWorkItemCount` to be converted to an Observable since `myWorkItemCount` is being used in the template before `_myWorkItems` subscription using `async`. 
Something like this `this.myWorkItemsCount = this._myWorkItems.pipe(map(workItems => workItems.length));`.

Because of this there will be 4 subscriptions and API calls via `async` pipes instead of 1. That is my concern.

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
divyanshiGupta: @sahil143 Adding a method in `workitems-util` won't work since it will need to use `FilterService` which is an `Injectable`. That is why it can only be used by injecting it in an Angular Component or Service. Will either need to create another service in `fabric8-ui` which will build and return filter queries or duplicate the same code everywhere the filter is needed.

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
aptmac: What we can do here is add something like `*ngIf="(myWorkItems | async)?.length as myWorkItemsCount"` to one of the higher level divs in the html, and then reference it the count elsewhere in the html so the subscription only occurs once.

divyanshiGupta: @aptmac There are three places where we need to use `myWorkItemsCount`, two of them are with `ngIf` and first one is just `interpolation`. And they are distributed across multiple divs. Only higher level div left would be the top level div. If we add `*ngIf` to it like you suggested then until the workitem data returns from API call, the `div` will not be rendered at all which is not desired UX behaviour. 
the other way we can do this is to do a subscription for `myWorkItemsCount` in `component.ts` which will reduce the subscription to only 2. But then again it will defeat the whole thing of making it an Observable and using with `async` pipe in the first place.

divyanshiGupta: Also, about the racy conditions i don't think here it will be a problem since we are only rendering data in template and not trying to manipulate it anywhere.

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
sahil143: What we can do here is pass injected `FilterService` as a parameter to the method.

divyanshiGupta: That will solve the problem. But that doesn't seem like a good solution. Passing injectable service as a method parameter seems wrong to me because then the method will have to rely on the calling component to pass correct service to it which means the code can break if correct service is not passed to it.

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
sahil143: > There are three places where we need to use myWorkItemsCount, two of them are with ngIf and first one is just interpolation. And they are distributed across multiple divs. Only higher level div left would be the top level div. If we add *ngIf to it like you suggested then until the workitem data returns from API call, the div will not be rendered at all which is not desired UX behaviour.

We dont have to use `*ngIf`. We can use NgLetDirective(`*ngLet = "(myWorkItems | async) as workitems"`) from fabric8-planner. It can be used at top level div, it just store the results in a variable and then do `workitems.length` to get the workitem count

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
sahil143: > But that doesn't seem like a good solution

Why?

> Passing injectable service as a method parameter seems wrong to me because then the method will have to rely on the calling component to pass correct service to it which means the code can break if correct service is not passed to it

You can do type checking. If the passed instance is not of the correct type. It will fail during compile time.
This will look something like this
```ts
closedWorkitemStateQuery(filterService: FilterService) {
....
}
```

divyanshiGupta:
@aptmac @sahil143 I have made the necessary changes. Please have a look.

src/app/shared/workitem-utils.ts
```diff
@@ -11,6 +11,36 @@ export function filterOutClosedItems(items) {
...
+
+export function buildSpaceQuery(filterService: FilterService, spaceId) {
+  return filterService.queryBuilder(
+    'space', filterService.equal_notation, spaceId
+  );
+}
+
+export function buildClosedWorkItemQuery(filterService: FilterService) {
+  let stateQuery = {};
+  ['closed', 'Done', 'Removed', 'Closed'].forEach(state => {
```
aptmac: missing type definition on `state`

src/app/shared/workitem-utils.ts
```diff
@@ -11,6 +11,36 @@ export function filterOutClosedItems(items) {
...
+  );
+}
+
+export function buildSpaceQuery(filterService: FilterService, spaceId) {
+  return filterService.queryBuilder(
+    'space', filterService.equal_notation, spaceId
+  );
+}
+
+export function buildClosedWorkItemQuery(filterService: FilterService) {
```
aptmac: missing type definition for the function, whatever type `stateQuery` returns

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.spec.ts
```diff
@@ -4,10 +4,11 @@ import { By } from '@angular/platform-browser';
 import { ActivatedRoute, Router, RouterModule } from '@angular/router';
 import { NgArrayPipesModule } from 'angular-pipes';
 import { FilterService, WorkItem, WorkItemService } from 'fabric8-planner';
+import { NgLetModule } from 'fabric8-planner/app/shared/ng-let';
 import { Broadcaster } from 'ngx-base';
-import { Context, Contexts } from 'ngx-fabric8-wit';
+import { Context, Contexts, Space } from 'ngx-fabric8-wit';
```
aptmac: There are unused imports to clean up: `WorkItem`, `UserService`, `ConnectableObservable`, `publish`

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
```diff
@@ -1,54 +1,77 @@
...
+    ].forEach(query => {
+      filters = this.filterService.queryJoiner(
+        filters, this.filterService.and_notation, query
+      );
+    });
+
+    this._myWorkItems = this.workItemService
+      .getWorkItems(100000, {expression: filters}).pipe(
+        map((val: WorkItemsData) => val.workItems),
+        tap(workitems => console.log(workitems)),
```
aptmac: there's a `console.log` here that should be removed

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
```diff
@@ -1,54 +1,77 @@
...
     return this._myWorkItems;
   }
 
+  getCurrentSpaceWorkItems() {
+    let filters = {};
+    [
+      buildAssigneeQuery(this.filterService, this.loggedInUser.id),
+      buildSpaceQuery(this.filterService, this.currentSpace.id),
+      buildClosedWorkItemQuery(this.filterService)
+    ].forEach(query => {
```
aptmac: missing type definition on `query`

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
```diff
@@ -1,54 +1,77 @@
...
+      filters = this.filterService.queryJoiner(
+        filters, this.filterService.and_notation, query
+      );
+    });
+
+    this._myWorkItems = this.workItemService
+      .getWorkItems(100000, {expression: filters}).pipe(
+        map((val: WorkItemsData) => val.workItems),
```
aptmac: missing type definition e.g., `map((val: WorkItemsData): WorkItem[] => [..]`

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
```diff
@@ -1,54 +1,77 @@
...
+        tap(workitems => console.log(workitems)),
+        // Resolve the work item type
+        tap((workItems: WorkItem[]) => workItems.forEach((workItem: WorkItem) => this.workItemService.resolveType(workItem))),
```
aptmac: should have a `: void` for the inner fn of these `tap`s similar to the one above for `map`

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.spec.ts
```diff
@@ -19,52 +20,52 @@ import { WorkItemsData } from '../../shared/workitem-utils';
...
           let workItemServiceMock = jasmine.createSpyObj('WorkItemService', ['buildUserIdMap', 'getWorkItems']);
-          workItemServiceMock.buildUserIdMap.and.returnValue(fakeUser);
           workItemServiceMock.getWorkItems.and.returnValue(observableOf({ workItems: [] }) as Observable<WorkItemsData>);
           return workItemServiceMock;
         }
       },
       {
         provide: FilterService, useFactory: () => {
-          let filterServiceMock = jasmine.createSpyObj('FilterService', ['queryBuilder']);
+          let filterServiceMock = jasmine.createSpyObj('FilterService', ['queryBuilder', 'queryJoiner']);
```
aptmac: missing type on the `filterServiceMock`, e.g., `jasmine.SpyObj<FilterService>`

src/app/shared/workitem-utils.ts
divyanshiGupta: @aptmac `queryJoiner` and `queryBuilder` doesn't return any specific type of object

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.spec.ts
divyanshiGupta: Done

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
```diff
@@ -1,54 +1,77 @@
...
+  ngOnChanges() {
+    this.getCurrentSpaceWorkItems();
   }
 
   get myWorkItems(): Observable<WorkItem[]> {
     return this._myWorkItems;
   }
 
+  getCurrentSpaceWorkItems() {
+    let filters = {};
```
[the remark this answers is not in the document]
divyanshiGupta: As mentioned above there is no type definition for filters or queries returned from `filterService`.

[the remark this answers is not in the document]
aptmac: Ah, fair enough

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.ts
divyanshiGupta: same reason as mentioned above

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.spec.ts
```diff
@@ -19,52 +20,52 @@ import { WorkItemsData } from '../../shared/workitem-utils';
...
       fullName: 'fakeName',
       imageURL: 'null',
       username: 'fakeUserName'
     }
-  } as User);
+  };
+}
+
+describe('CreateWorkItemWidgetComponent', () => {
+  type TestingContext = TestContext<CreateWorkItemWidgetComponent, HostComponent>;
```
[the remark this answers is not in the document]
divyanshiGupta: Now it is being used as a type for `testContext`

nimishamukherjee:
@aptmac if all your concerns have been addressed, please approve this PR.

joshuawilson:
[test]

divyanshiGupta:
> The last part would be a unit test of some sort, perhaps make a test (or few) that are given 2+ fake work items with one assigned to a user.id, and verify that only the work item assigned to the user.id shows up.

@aptmac The issue with this kind of test is that the logic of filtering out `workItems` based on any filter criteria like `assignee` is not in this component. It is in `WorkItemService`, so if we mock the `getWorkItems` and return a few fake workitems all of them will be shown by the component since there is no filtering in component. So, this behavior cannot be tested in this component. It should be tested in the service or rather in the backend since that's where the actual filtering of workitems happen.

rohitkrai03:
@divyanshiGupta That does make sense. Perhaps we could test some other scenarios here.

divyanshiGupta:
@rohitkrai03 I have added some other test cases for rendering of workitems.

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.spec.ts
```diff
@@ -19,90 +21,189 @@ import { WorkItemsData } from '../../shared/workitem-utils';
...
+            return mockRouter;
+          }
+        }
+      ],
+      schemas: [
+        NO_ERRORS_SCHEMA
+      ]
+    });
+
+    it('should fetch the correct work items', async(function() {
```
aptmac: I'd rather this be `async(() => { ..` to match other similar functions throughout the codebase

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.spec.ts
```diff
@@ -19,90 +21,189 @@ import { WorkItemsData } from '../../shared/workitem-utils';
...
+          provide: WorkItemService, useFactory: () => {
+            let workItemServiceMock = jasmine.createSpyObj('WorkItemService', ['resolveType', 'resolveAreaForWorkItem', 'getWorkItems']);
+            workItemServiceMock.resolveType.and.stub();
+            workItemServiceMock.resolveAreaForWorkItem.and.stub();
+            workItemServiceMock.getWorkItems.and.returnValue(fakeWorkItemsObs as Observable<WorkItemsData>);
+            return workItemServiceMock;
+          }
+        },
+        {
+          provide: FilterService, useFactory: () => {
```
aptmac: missing type definition on `()`

src/app/dashboard-widgets/create-work-item-widget/create-work-item-widget.component.spec.ts
```diff
@@ -19,90 +21,189 @@ import { WorkItemsData } from '../../shared/workitem-utils';
...
+    const testContext: TestingContext = initContext(CreateWorkItemWidgetComponent, HostComponent, {
+      declarations: [ MockFeatureToggleComponent ],
+      imports: [NgArrayPipesModule, RouterModule, NgLetModule],
+      providers: [
+        { provide: ActivatedRoute, useValue: jasmine.createSpy('ActivatedRoute') },
+        { provide: LocationStrategy, useValue: jasmine.createSpyObj('LocationStrategy', ['prepareExternalUrl']) },
+        { provide: Broadcaster, useValue: createMock(Broadcaster) },
+        { provide: Contexts, useValue: ({ current: new Subject<Context>() }) },
+        {
+          provide: WorkItemService, useFactory: () => {
```
aptmac: missing type definition on ()

divyanshiGupta:
@aptmac Updated the PR with latest changes.

PR #3366, opened 2018-10-10
fix(stack-report-btn): fixes stack report button not appearing in pip…

invincibleJai:
- fixes stack report button not appearing in pipelines view wrt recent change in stage name

Tracks: https://github.com/openshiftio/openshift.io/issues/4405

### DIFF
--- a/src/a-runtime-console/kubernetes/ui/pipeline/build-stage-view/build-stage-view.component.html
+++ b/src/a-runtime-console/kubernetes/ui/pipeline/build-stage-view/build-stage-view.component.html
@@ -73,7 +73,7 @@
               </div>
             </div>
               <!-- Fabric-stack-analysis-ui -->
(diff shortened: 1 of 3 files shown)


joshuawilson:
What configuration gets a build with a report? I just tried Spring and REST and got no report.

invincibleJai:
@joshuawilson let me try with spring and rest. Ideally if we try this PR and build goes through successful, Stack report should show up in dashboard as well as stack report link should appear in pipelines.

Stack report should work fine if annotations from build are proper as in https://github.com/openshiftio/openshift.io/issues/3744#issuecomment-395299143

invincibleJai:
I tried spring and rest, it works for me.

screen shot 2018-10-11 at 8 36 31 am

joshuawilson:
I ran your branch locally and I don't see the report link.

invincibleJai:
@joshuawilson oh, can you pls share jenkins log for that build.

joshuawilson:
I can see it now.

PR #3368, opened 2018-10-11
fix(codebases): fix switch component change state trigger after upgrade

rohitkrai03:
Fixes https://github.com/openshiftio/openshift.io/issues/4407

- After the upgrade of `jw-bootstrap-switch-ng2` the change trigger event changed to `changeState` from `onChangeState`.
- This PR fixes that.


aptmac:
The changes to `changeState` make sense [[0]](https://github.com/JulioWar/jw-bootstrap-switch-ng2/blob/edd4c59134ccc8708ae1ae2d145e49f4ccf7c3bb/README.md#version-200). Having said that, both locally with this PR applied and on production I'm still seeing the issue where the notification pops up if you turn it on (but after a refresh the toggle is successfully turned on, and can be turned off without a problem). 

Here's what I'm seeing with this PR applied:
ezgif-1-e1dfd54cc5d5

[0] https://github.com/JulioWar/jw-bootstrap-switch-ng2/blob/edd4c59134ccc8708ae1ae2d145e49f4ccf7c3bb/README.md#version-200

rohitkrai03:
@aptmac I am not getting any such errors. The issue i was facing was related to backend. Which we discussed on MM platform channel. See https://github.com/openshiftio/openshift.io/issues/4407#issuecomment-428948357

PR #3377, opened 2018-10-15
fix(header): fix capitalization of 'Create Space' in the user dropdown

aptmac:
This one-liner fixes https://github.com/openshiftio/openshift.io/issues/4420, in which the capitalization of `Create space` to `Create Space` in the user menu is fixed.

Before:
before

After:
after


joshuawilson:
@serenamarie125 does this work for you?

PR #3378, opened 2018-10-15
Home page empty state

andrewazores:
This PR is a WIP for https://openshift.io/openshiftio/Openshift_io/plan/detail/828 / https://github.com/openshiftio/openshift.io/issues/4194 . Currently implemented is a home page empty state that directs the user toward creating their first Space, which can then be populated with a Codebase (Application), etc. PR updates are subject to design clarification of a few elements in the linked design issue. Primarily, the change to the Launcher UI is not implemented.

GIF showing current implementation of the empty state:

home-page-empty-state

Once a space is created the home page looks otherwise unchanged, with the small exception of a loading widget briefly appearing while the list of spaces is initially loaded (to check if the empty state should be displayed or the normal dashboard).


alien-ike:
### Ike Plugins (test-keeper)

andrewazores:
Here's a quick demo video showing the latest status of the working prototype:

https://youtu.be/CKWMpYbdvDU

This mostly meets the revised design objectives for the parent issue. The home page empty state is created, but it currently appears when the current logged in user owns 0 spaces, even if they are a collaborator on others. Clicking on the "Create New Codebase" or "Import Existing Codebase" Call-To-Actions summons the Add Space Overlay with the Space Name field focused. Continuing from there summons the Add App Overlay with the Application Name field focused and the Create/Import radio buttons preselected to match the Call-To-Action that was initially selected. Continuing then proceeds into an unmodified Launcher flow. The Add Space and Add App Overlays will also focus their input fields when summoned from other locations than the home page empty state widget, but the codebase create/import will not be prefilled.

@serenamarie125 please review the above video and let me know if anything presented does not match the design as we have agreed, other than the issue with owning vs collaborating on spaces as noted above.

andrewazores:
The empty state now appears only when the logged in user owns and collaborates on a total of 0 spaces. The commits have been cleaned up and tests added. Removing WIP label.

aptmac:
This looks good so far, I've found a couple things so far but will add to my comments as I go through this.

For testing purposes I'm just going into the `home.component` and forcing `spacesCount` to be 0. I've noticed that the empty-state applies to feature levels "experimental" and above, but not to "production" or "beta", was this intended?

There seems to be something a bit off with the stylesheets - when my page is full-screened the empty state doesn't sit vertically centred in my screen.
full-screen

Additionally, causing the page to use a half-screen layout causes some overlap.
half-screen

andrewazores:
@aptmac I've cleaned up the CSS some in the latest commit. Elements should no longer overlap or get pushed out of boundaries for reasonable screen sizes (it can still happen if you go very very small, I think). The card isn't perfectly centered on screen still, it sits a bit high, but it's consistent across screen sizes I believe and is pretty close to centered.

aptmac:
Just for reference, here's what it looks like now.

screenshot

joshuawilson:
@serenamarie125 @fabric8-ui/uxd please review

joshuawilson:
@fabric8-ui/test @fabric8-ui/docs this change will likely have impacts in your area.

andrewazores:
If the double "Loading ..." label you mean is the one shown immediately after login in the GIF, then yes, that's really there and is not a recording artefact, but it's unrelated to this change. That's been there since the Angular upgrade and IIRC has something to do with a change in the way that CSS is bundled with the application.

Which other "odd animation issues" do you mean?

christianvogt:
@serenamarie125 
@andrewazores is correct that the `Loading ...` label looks odd in the gif since the angular upgrade. This only shows up in the development environment and not on prod preview or production. This is something we will look to fix in the configuration of the monorepo setup.

joshuawilson:
`Loading...` is the default screen. It is replaced with the way we load from the server but when running locally it is the only thing we have. We can change the text.

Preeticp:
Raised doc PR for updating docs wrt these changes here: https://github.com/fabric8io/fabric8-online-docs/pull/448

PR #3380, opened 2018-10-18
fix(pipeline-call): adds OSIO X-App header

bartoszmajsak:
Fixes https://github.com/openshiftio/openshift.io/issues/4453

### DIFF
--- a/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
+++ b/src/a-runtime-console/kubernetes/ui/pipeline/input-action-dialog/input-action-dialog.component.ts
@@ -72,7 +72,9 @@ export class InputActionDialog implements OnDestroy {
         console.log('Warning no jenkinsNamespace on the Build!');
       } else {
         url = pathJoin(forgeUrl, '/api/openshift/services/jenkins/', jenkinsNamespace, url);
-        const headers = new HttpHeaders({ 'Authorization': 'Bearer ' + this.authService.getToken() });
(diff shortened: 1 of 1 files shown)


alien-ike:
### Ike Plugins (test-keeper)

Thank you @bartoszmajsak for this contribution!



bartoszmajsak:
@joshuawilson @christianvogt we need this kind of header in a lot of places and I see right now it's just copied and pasted in every Launcher service. Wouldn't it make sense to have an interceptor taking care of that and presumably token propagation?

rohitkrai03:
@bartoszmajsak Yes, this can be converted into an interceptor which will add `X-App` header to all request going to launcher endpoint.

bartoszmajsak:
It might need a test or two. Can't really do it this week due to travels so if anyone is willing do it I offer good :beer:

bartoszmajsak:
@rohitkrai03 could we plan this work?

rohitkrai03:
@bartoszmajsak Yes, we can add a planner story for this. I'll be willing to work on this. Also, you were talking about adding tests to this PR or the intercepter work?

gastaldi:
>(Do you need this merged while the interceptor is written?)

We need this fix to update the launcher-backend production to the latest version, so yes :)

joshuawilson:
@rohitkrai03 please create a task to deal with this in an interceptor.

rohitkrai03:
@joshuawilson Sure, will do that.

ppitonak:
@rohitkrai03 can you provide link to that issue here?

PR #3381, opened 2018-10-18
fix(spaces): improve the spaces page to display owned and shared spaces

aptmac:
This PR addresses OSIO task [Add filter button/ability to Spaces page header](https://openshift.io/openshiftio/Openshift_io/plan/detail/844) [0], which is a sub-task of [Improve the My Spaces page](https://openshift.io/openshiftio/Openshift_io/plan/detail/786) on OSIO [1] and [Improve the My Spaces page](https://github.com/openshiftio/openshift.io/issues/4326) on GitHub [2].

~~I'll be waiting to rebase this on top of https://github.com/fabric8-ui/fabric8-ui/pull/3378, so the function that requests information from `/api/user/spaces` can be included in the new `user-spaces.service.ts`.~~

Additionally, there were a couple of unit tests in the `my-spaces.component.spec.ts` that weren't performing as expected, so I'll be continuing to update the unit tests and add more tests for the new functionality as well.

Here's a gif of what the toggling looks like in action.
demo

These changes required a new empty-state page for the case where the user owns spaces, but not collaborate on any [[3]](https://github.com/openshiftio/openshift.io/issues/4326#issuecomment-430471985).
empty-state

[0] https://openshift.io/openshiftio/Openshift_io/plan/detail/844
[1] https://openshift.io/openshiftio/Openshift_io/plan/detail/786
[2] https://github.com/openshiftio/openshift.io/issues/4326
[3] https://github.com/openshiftio/openshift.io/issues/4326#issuecomment-430471985

PR #3382, opened 2018-10-22
fix(spaces): re-introduce missing 'pageSize' attribute when fetching user's spaces

aptmac:
This PR addresses a side-effect of my recent PR [[0]](https://github.com/fabric8-ui/fabric8-ui/pull/3381) to add shared spaces to the 'Spaces' page.

The service function used to retrieve spaces by username has an optional parameter `pageSize`, that defaults to 20 if not specified. The former 'Space' page had a variable called `pageSize` hold a value of '2000' [[1]](https://github.com/fabric8-ui/fabric8-ui/pull/3381/files#diff-751600f87e785e8a26f9e457250b05b1L45), and I had forgotten to replace the `pageSize` value when cleaning up the component. Instead of just setting an arbitrary large number such as '2000' I thought it may be a good idea to use the number returned by `user-spaces.service.getInvolvedSpacesCount()` because that total will encompass the number of spaces the user collaborates on and owns, which will be a closer representation of how many spaces to ask for when calling `getSpacesByUser()`.

[0] https://github.com/fabric8-ui/fabric8-ui/pull/3381
[1] https://github.com/fabric8-ui/fabric8-ui/pull/3381/files#diff-751600f87e785e8a26f9e457250b05b1L45


aptmac:
> Looks fine. I wonder about page performance though - there are a lot of potentially wasted requests going on with various subscriptions through the UserSpacesService at this point. Do you think that service can implement some form of caching as a followup?

Hm, yeah I think there could be a way to keep some information around so it doesn't need to be fetched every time. Perhaps listen for broadcasted events pertaining to space creation/editing/deletion and only then update some cached `sharedSpaces` list that gets initialized with the service? Also, I recall hearing about some sort of different way with talking to the backend, perhaps related to this [comment [0]](https://github.com/openshiftio/openshift.io/issues/4326#issuecomment-430379021)? If there was a better way to get this information (like shared spaces) from the database then we wouldn't need multiple functions chained together to get the information.

[0] https://github.com/openshiftio/openshift.io/issues/4326#issuecomment-430379021

andrewazores:
Sure, something like that might be workable. Even just temporally caching the UserSpacesResponse would do some good, since those values aren't likely to be rapid changing. getInvolvedSpacesCount could also be refactored to not actually fire off a new request but use a cached response from the last time getInvolvedSpaces was called (if any - if not then do fire off a request), for example.

joshuawilson:
Please talk to @christianvogt since he has been planning how to manage state.

PR #3385, opened 2018-10-23
fix(format): change all type casts to use `as`

christianvogt:
Switched all usage of `<>` type casting to use `as` since `<>` is not supported alongside jsx in tsx files.

This is needed to support compiling react jsx going forward.


joshuawilson:
I confirm the docs point to the same thing.

PR #3394, opened 2018-10-26
fix(workspaces): removing retry calls from open / create workspace methods (#4416)

ibuziuk:
openshift.io issue - https://github.com/openshiftio/openshift.io/issues/4416

`retry` calls should be removed completely from open / create workspace methods since those have been added back to the time when Che server used to be single-tenant and it was required to unidle it, which might caused timeouts. Currently, Che server is multi-tenant and is running on osd so those retries not needed anymore.


alien-ike:
### Ike Plugins (test-keeper)

Thank you @ibuziuk for this contribution!

PR #3397, opened 2018-10-26
feat(e2e): add scripts to simplify running e2e tests

christianvogt:
https://openshift.io/openshiftio/Openshift_io/plan/detail/651

Created a gulp file with tasks related to e2e setup and running.

* `npm run e2e` (`npm run test:e2e`): starts the e2e tests using the gulp tasks
* `npm run e2e:clean`: cleans up the `fabric8-test` project which is checked out when running e2e tests
* `npm run e2e:reconfig`: clears the test config file and re-prompts the user for input

Running the script will checkout the `fabric8-test` project as a sibling to the `fabric8-ui` project. It will then run `npm install` to setup the test dependencies and then prompt the user to fill out the missing configuration data (account user names and passwords). Finally it will ask the user which e2e test to run.

This is a step forward to helping devs run e2e smoke tests on a regular basis.

Unfortunately simple e2e tests are failing at this time.


joshuawilson:
marked it as work in progress since the tests are failing

christianvogt:
Whether or not the tests pass won't change the setup for running them. They do run but fail at some point due to assertion errors.

There is no harm in having the setup checked in.

I will be following up with someone from QE 
next week regarding the failures.

gulpfile.js
```diff
@@ -0,0 +1,122 @@
...
+gulp.task('e2e:clean', () => {
+  if (fs.existsSync(E2E_PROJECT)) {
+    console.log(`Deleting project ${E2E_PROJECT}`);
+    return gulp.src(E2E_PROJECT, {read: false}).pipe(clean({force: true}));
+  }
+  return Promise.resolve();
+});
+
+// fetch and update e2e tests
+gulp.task('e2e:fetch', async () => {
```
joshuawilson: `gulp.task` is the old way. In v4 they moved to `function`. 

https://gulpjs.com/docs/en/getting-started/creating-tasks

gulpfile.js
```diff
@@ -0,0 +1,122 @@
+const fs = require('fs');
+const path = require('path');
+const gulp = require('gulp');
```
joshuawilson: You can also make the code easier to read by importing the API's directly. 
`const { src, dest } = require('gulp');`
Then you can call `return src(E2E_PROJECT, {` without the `gulp.src`.

gulpfile.js
christianvogt: Updated usage to align with v4 syntax.

gulpfile.js
christianvogt: Updated imports.

gulpfile.js
```diff
@@ -0,0 +1,122 @@
...
+  const testUrls = {
+    'enter custom': null,
+    dev: 'http://localhost:3000',
+    'local prod': 'http://localhost:8080',
+    production: 'http://openshift.io',
+    'prod-preview': 'http://prod-preview.openshift.io',
+  };
+
+  // read the existing config and
+  return gulp
```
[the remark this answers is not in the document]
christianvogt: Broke out code into multiple smaller functions as part of v4 refactoring.

christianvogt:
e2e tests run locally. The issue was I entered my email address instead of the simpler username in the config of the e2e tests.

I'm still having an issue with the smoke tests running on mac. There's an issue where a select field isn't interacting with the test runner as expected.

gulpfile.js
```diff
@@ -0,0 +1,164 @@
...
+function e2eClean() {
+  if (fs.existsSync(E2E_PROJECT)) {
+    console.log(`Deleting project ${E2E_PROJECT}`);
+    return src(E2E_PROJECT, {read: false}).pipe(clean({force: true}));
+  }
+  return Promise.resolve();
+}
+
+// fetch and update e2e tests
+async function e2eFetch() {
```
ppitonak: Failed for me because I already had `fabric8-test` checked out and was on a different branch. It would be safer to run `git checkout master` before pulling changes.

gulpfile.js
```diff
@@ -0,0 +1,164 @@
...
+
+async function e2eStart() {
+  try {
+    execSync('npm start', {stdio: 'inherit', cwd: TEST_DIR});
+  } catch (err) {
+    // suppress as the test would have already logged to console
+  }
+}
+
+const e2eConfig = series(e2eEnsureConfigFile, e2eEditConfig);
```
ppitonak: Would it make sense to put `e2eConfig` and `e2eReconfig` directly to `fabric8-test` repository? WDYT @ljelinkova?

gulpfile.js
```diff
@@ -0,0 +1,164 @@
...
+}
+
+
+function promptTestSuite() {
+  return testSuites[
+    readlineSync.keyInSelect(testSuites, 'Select a test suite: ', {cancel: false})
+  ];
+}
+
+function promptOsioUrl(currentValue) {
```
ppitonak: How do I exit from this prompt? Neither `Ctrl+C` nor `Ctrl+D` work.

ppitonak: There is something wrong here... When I select `0` to use current URL, it shows the same prompt again

package.json
```diff
@@ -78,6 +81,7 @@
     "jest:debug": "node --inspect-brk ./node_modules/.bin/jest --runInBand --env=jsdom",
     "test": "npm run jest",
     "test:debug": "npm run jest:debug",
+    "test:e2e": "npm run e2e",
```
ppitonak: It might make sense to also introduce a task starting e2e tests inside docker container if you only need to verify that your change in UI doesn't break the test and you don't need to see what is happening.

christianvogt:
@ppitonak There is a change in `package-lock.json` because of the version ranges being used in the `package.json`. Unless you remove all your ranges, it's expected.

package.json
christianvogt: I could add this.
We have several people using mac, I being one. I get the following error when running `npm run image:start`:
```
The path /etc/localtime
is not shared from OS X and is not known to Docker
```

ppitonak: Please report this kind of issues in https://github.com/fabric8io/fabric8-test/issues/

gulpfile.js
christianvogt: Thanks. I'll add a fix for this.

gulpfile.js
christianvogt: argh! a silly mistake. Thanks for catching.

package.json
christianvogt: Reported: https://github.com/fabric8io/fabric8-test/issues/1123

PR #3405, opened 2018-11-02
update monorepo tsconfig and jest

christianvogt:
While attempting to run babel typescript instead of ts-loader, I made the following changes.

Need to separate out HTML and angular component transforms for jest so that it is easier to switch out the typescript transform portion only later.

Also update tsconfigs as noted by suggestions while attempting to use create-react-app typescript support.

Added suppprt to jest to read a `.ts` test setup file.


alien-ike:
### Ike Plugins (test-keeper)

PR #3408, opened 2018-11-06
Fix(recent-utils): fix signature of recent-utils#compareElements

christianvogt:
Fixes the signature of `recent-utils#compareElements` which causes test failures in the monorepo. The abstract function should not be typed as the generic comes from the class and not should not come from the function.

This function is implemented in `spaces.service` and `context.service` as follows:
```
  compareElements(s1: Space, s2: Space): boolean {
    return s1.id === s2.id;
  }
```
```
  compareElements(c1: Context, c2: Context): boolean {
    return c1.name === c2.name;
  }
```

### DIFF
--- a/src/app/shared/context.service.ts
+++ b/src/app/shared/context.service.ts
@@ -147,7 +147,7 @@ export class ContextService extends RecentUtils<Context> implements Contexts {
         .pipe(
           withLatestFrom(this._recent),
           map(([deletedSpace, recentContexts]: [Space, Context[]]): RecentData<Context> =>
-            this.onBroadcastSpaceDeleted(deletedSpace, recentContexts, this.compareContextToSpace)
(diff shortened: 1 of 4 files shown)


alien-ike:
### Ike Plugins (test-keeper)

joshuawilson:
Why does the generic type need to come from the class? I can't find anything in the docs saying that.

joshuawilson:
I don't think method is called anyplace but from within this class. The methods that are called don't use the compare feature. You might be able to remove it.

christianvogt:
The type applied at the function level `compareElements<T>` infers the type from the assignment which is wrong because the return value is a boolean. Whereas we should be using `T` from the class itself.

joshuawilson:
There is no unit test for this. 
I know it is an abstract class, but you should be able to mock it.

christianvogt:
@joshuawilson well that was fun writing tests :)

PR #3410, opened 2018-11-07
chore(less): remove relative paths to less files located in node_modules

christianvogt:
It's bad practice to use relative paths to reach into node_modules. This actually breaks the monorepo setup because the packages may be located in the monorepo root.

Added a new path resolution location to the less loader config for webpack.

Ideally we could move to using webpack's resolver using the `~` prefix however patternfly using absolute imports without the `~` and therefore we must continue to use the `less-loader` paths.


alien-ike:
### Ike Plugins (test-keeper)

PR #3413, opened 2018-11-08
chore(monorepo): update prettier version to support html

christianvogt:
Introduce the latest version of prettier capable of formatting angular html templates.
https://prettier.io/blog/2018/11/07/1.15.0.html


alien-ike:
### Ike Plugins (test-keeper)

talamer/packages/eslint-plugin-osio/package.json
```diff
@@ -12,7 +12,7 @@
   "license": "Apache-2.0",
   "peerDependencies": {
     "eslint": "^5.7.0",
-    "prettier": "^1.14.3",
+    "prettier": "^1.15.0",
```
joshuawilson: why is this not `1.15.1`, typo?

talamer/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "arrowParens": "always",
+  "singleQuote": true,
```
joshuawilson: What does this do? We have single quotes for code and double for html. That allows us to write html in code and not have to escape the quotes.

talamer/packages/eslint-plugin-osio/package.json
christianvogt: A peer typically is meant to be broad. In actuality it should probably be `*` since the package supports any prettier version.

talamer/.prettierrc
christianvogt: This actually refers to quotes in js. I copied our settings from the lint plugin. The reason for file is for the printwidth settings to apply to html.

I'd like to run prettier through our linter like js for html but it doesn't work.

PR #3417, opened 2018-11-13
fix(version): upgrade planner version to 0.58.1

sahil143:


joshuawilson:
This was updated in the overall upgrade so it is now out of date. Closing.

PR #3428, opened 2018-11-30
fix(pipelines): Add pointer cursor to pipeline dropdown

jarifibrahim:
This PR adds pointer cursor to `start pipeline` and `delete pipeline` option in the dropdown of the pipeline page.


jarifibrahim:
[test]

christianvogt:
This is so hacky. I wonder why the dropdown menus don't come with a pointer cursor always. They do if it's within a toolbar...
We really need to clean up the reusable components so we can avoid these types of fixes.

jarifibrahim:
@christianvogt I tried to figure out the why an `<a>` tag doesn't have a pointer cursor but I couldn't find it. An `<a>` tag has a cursor pointer by default but in our case the cursor was set to `auto`. Do you know where this might be coming from?

christianvogt:
@jarifibrahim `<a>` tags without an `href` attribute shows a `default` cursor and not a `pointer`. This is why we need to explicitly assign a `pointer` cursor. That being said, we should have reusable components that assign the correct styles. Part of the problem is that we do not use the hierarchy structure correctly in some cases and therefore we're missing the cascading styles, and another part is that we don't always have an encapsulated reusable component that hide the intricacies of such html widgets.

jarifibrahim:
@christianvogt Agreed. So what can we do to fix it? If it's an easy fix, I might be able to help :)

src/a-runtime-console/kubernetes/ui/pipeline/list/list.pipeline.component.html
```diff
@@ -23,20 +23,20 @@ <h3>
               </button>
               <ul class='dropdown-menu-right dropdown-menu' aria-labelledby='dropdownKebabRight9' role="menu" *dropdownMenu>
                 <li>
-                  <a (click)="startBuild(pipeline)" title="Start this Pipeline">Start Pipeline</a>
+                  <a class="pointer" (click)="startBuild(pipeline)" title="Start this Pipeline">Start Pipeline</a>
```
joshuawilson: In the Codebases page this `<a>` tag has `href="javascript:void(0)" class="dropdown-item secondary-action"` in it. It might be worth doing the same thing here instead. Or we should update that one too.

christianvogt: Semantically a link with href should be used only for navigating the app.

A button should be used for executing actions.

Buttons also do not show a pointer cursor by default. They also differ from links in that they can be clicked with the space button.

Overloading the href to execute js or assigning href value  to # is something I'd avoud. I suggest keeping the CSS to assign a pointer cursor.

joshuawilson: That is where my thoughts started to head too, so we should update the other one too. The one in Codebases. 

We also need to look at the dropdown component, like you said.

jarifibrahim:
@joshuawilson @christianvogt Shall we merge this PR?

PR #3434, opened 2018-12-07
feat(roles): add feature flags on settings icon and collaborators page

rohitkrai03:
Planner Story - https://openshift.io/openshiftio/Openshift_io/plan/detail/1211

This PR adds feature flag to the logic to hiding settings icon in the header and on the collaborators page inside settings.

- The header component is rendered by using `MenusService` to determine if the menu item needs to be rendered or not.
- `MenusService` checks for feature enabled menu items and send it to component level where the logic to hide something based on user is owner or admin is decided.
- This is why its tricky to introduce feature flag at the header level based on the user permission as well as the feature enablement.
- For achieving this we had to introduce on more property called `subFeature` in the `MenuItem` type interface. Based on the `feature` property menu item is deleted from the list using menu.service. After that the component uses `subFeature` to further decide which logic to use based on faeture flag along with user permission.
- Add collaborators page under feature flag.
- Feature flag for both the pages are set to `beta` as planner is also set at same level.

### DIFF
--- a/src/app/layout/header/header.component.ts
+++ b/src/app/layout/header/header.component.ts
@@ -44,7 +44,7 @@ export class HeaderComponent implements OnInit, OnDestroy {
       'settings', function(headerComponent, menuItem) {
         const subFeature = menuItem['subFeature'];
         const allFeatures = headerComponent.context.user['features'];
-        if (headerComponent.menusService.isFeatureUserEnabled(subFeature, allFeatures)) {
+        if (subFeature && allFeatures && headerComponent.menusService.isFeatureUserEnabled(subFeature, allFeatures)) {
           return headerComponent.loggedInUserNotSpaceAdmin();
         }
         return headerComponent.checkContextUserEqualsLoggedInUser();
--- a/src/app/layout/header/header.component.ts
+++ b/src/app/layout/header/header.component.ts
@@ -9,9 +9,10 @@ import { MenuItem } from '../../models/menu-item';
 import { Navigation } from '../../models/navigation';
 import { LoginService } from '../../shared/login.service';
 import { MenuedContextType } from './menued-context-type';
+import { MenusService } from './menus.service';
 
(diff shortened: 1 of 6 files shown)


src/app/layout/header/header.component.ts
```diff
@@ -30,18 +31,23 @@ export class HeaderComponent implements OnInit, OnDestroy {
...
         return headerComponent.checkContextUserEqualsLoggedInUser();
       }
     ],
     [
-      'settings', function(headerComponent) {
-        return headerComponent.loggedInUserNotSpaceAdmin();
+      'settings', function(headerComponent, menuItem) {
+        const subFeature = menuItem['subFeature'];
+        const allFeatures = headerComponent.context.user['features'];
+        if (headerComponent.menusService.isFeatureUserEnabled(subFeature, allFeatures)) {
```
divyanshiGupta: @rohitkrai03 please add a check for `subFeature` and `allFeatures` as these can be undefined.

src/app/space/settings/collaborators/collaborators.component.html
```diff
@@ -1,99 +1,185 @@
...
+                  <span class="fa fa-ellipsis-v"></span>
+                </button>
+                <ul class="dropdown-menu-right dropdown-menu" aria-labelledby="dropdownKebab" *dropdownMenu>
+                  <li role="menuitem">
+                    <a [routerLink]="['/', item.attributes.username]" class="secondary-action">View user</a>
                   </li>
-                  <ng-template #cannotRemoveAdmin>
-                    <li class="disabled" title="Cannot remove owner of space as admin">
-                      <a class="secondary-action menu-item">Remove as admin</a>
+                  <li *ngIf="!adminCollaborators.includes(item.id) else removeAdmin" role="menuitem">
```
joshuawilson: you need a semi-colon after `!adminCollaborators.includes(item.id)`

PR #3437, opened 2018-12-10
fix(add-app-overlay): enable next button if form is valid

vikram-raj:
issue: https://github.com/openshiftio/openshift.io/issues/4616

### DIFF
--- a/src/app/space/add-app-overlay/add-app-overlay.component.html
+++ b/src/app/space/add-app-overlay/add-app-overlay.component.html
@@ -3,7 +3,7 @@
     <div class="modal-content">
       <div class="modal-header">
         Create an Application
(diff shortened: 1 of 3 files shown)


christianvogt:
@vikram-raj while this may fix the issue, can we unmount the component to reset all state when the modal isn't opened?

Btw i noticed that the ID `cancelImportsButton` is assigned to the next button.

vikram-raj:
> Btw i noticed that the ID cancelImportsButton is assigned to the next button.

Yes, it was in the black modal. and I didn't change it because of e2e test is using this id.

vikram-raj:
@vikram-raj while this may fix the issue, can we unmount the component to reset all state when the modal isn't opened?

> ok, I will do this.

vikram-raj:
[e2e-test-beta]

PR #3438, opened 2018-12-10
fix(spaces): fix save pin bug

rohitkrai03:
Fixes https://github.com/openshiftio/openshift.io/issues/4581


christianvogt:
@rohitkrai03 please remove `console.log` statements.
Why'd you remove a test case?

rohitkrai03:
@christianvogt That was a redundant test case.

PR #3442, opened 2018-12-13
fix(app-wizard-overlay): auto focus the input on modal open

vikram-raj:
Issue: https://github.com/openshiftio/openshift.io/issues/4617


joshuawilson:
How did auto focus get to be so hard?

joshuawilson:
you just include the html attribute `autofocus`

vikram-raj:
@joshuawilson ```autofocus``` attribute works only if the component gets loaded and visible. But here the case is different this modal component seats in the app component and it's been loaded as the app component gets loaded but not visible. That is why ```autofocus``` attribute doesn't in this case.

PR #3446, opened 2018-12-15
fix(build): exclude planner and ngx-widget packages until they are fixed

christianvogt:
Since packages `planner` and `ngx-widgets` haven't been updated to work in the monorepo, the build is breaking.

This change simply whitelists the packages that are valid in the monorepo until all packages are valid, at which point we can return to specifying `packages/*`.


alien-ike:
### Ike Plugins (test-keeper)

PR #3450, opened 2018-12-17
fix(table-config): Set arrows as disabled when no column is selected

abhinandan13jan:
Fix for table-cofig arrow issue
Ref: https://github.com/openshiftio/openshift.io/issues/3856


packages/planner/src/app/components_ngrx/table-config/table-config.component.html
```diff
@@ -1,55 +1,57 @@
...
+              type='checkbox'
+              [id]="col.name"
+              (change)='toggleCheckbox($event, col)'/>
+            <label [attr.for]="col.name">{{ col.name }}</label>
+          </li>
+        </ul>
+      </aside>
+      <aside>
+        <span class="db fa fa-angle-right"
+          tooltip placement="top" tooltip="{{isNoAttributeSelected?'Please select an attribute':'Move to Displayed Attributes'}}"
```
christianvogt: Duplicate `tooltip` attributes.

packages/planner/src/app/components_ngrx/table-config/table-config.component.html
```diff
@@ -1,55 +1,57 @@
...
+            <input
+              type='checkbox'
+              [id]="col.name"
+              (change)='toggleCheckbox($event, col)'/>
+            <label [attr.for]="col.name">{{ col.name }}</label>
+          </li>
+        </ul>
+      </aside>
+      <aside>
+        <span class="db fa fa-angle-right"
```
christianvogt: We should use a button instead of a span to be semantically correct. Then you also wouldn't need a custom disabled class but could use the disabled attribute instead.

packages/planner/src/app/components_ngrx/table-config/table-config.component.html
```diff
@@ -1,55 +1,57 @@
...
+          tooltip placement="top" tooltip="{{isNoAttributeSelected?'Please select an attribute':'Move to Displayed Attributes'}}"
+          [ngClass]="{'arrow-disabled':isNoAttributeSelected}"
```
christianvogt: This arrow is enabled even if there is no selected attributes to move to the other side.

packages/planner/src/app/components_ngrx/table-config/table-config.component.html
```diff
@@ -1,55 +1,57 @@
...
+          (click)="moveToDisplay()">
+        </span>
+        <span class="db fa fa-angle-left"
+          tooltip placement="top" tooltip="{{isNoAttributeSelected?'Please select an attribute':'Move to Available Attributes'}}"
+          [ngClass]="{'arrow-disabled':isNoAttributeSelected}"
```
christianvogt: This arrow is enabled even if there is no selected attributes to move to the other side.

abhinandan13jan:
@christianvogt I've updated and low applied the selection logic for each set of attributes. Also updated the arrows to buttons.

christianvogt:
@abhinandan13jan please apply bootstrap styles `btn btn-default` to the `button` tags. Unfortunately we don't have a common component yet for a button that applies the correct classes and therefore you must apply the bootstrap styles to all html components you create.
Also increase the margin between them to maybe `10px`.

I quickly tried this in web inspector and it looks like this.

Your changes:
image

With new button styles:
image

abhinandan13jan:
@christianvogt Added classes btn btn-default and increased margin to 10px

joshuawilson:
[test-e2e-beta]

PR #3455, opened 2018-12-18
[WIP] fix(quick-preview): auto-save field value when user leaves the field

divyanshiGupta:
Fixes https://github.com/openshiftio/openshift.io/issues/3951

This PR introduces the fix wherein when the user moves out of a field and clicks somewhere the field value gets automatically saved.


christianvogt:
Added concerns regarding the UX and the ok / checkmark button to the parent issue.

christianvogt:
@divyanshiGupta please look at issue #3951 and issue #4643. The real problem is data loss due to a bug in refreshing the work item details view.

divyanshiGupta:
@christianvogt taking a look at it.

joshuawilson:
Please rebase.

christianvogt:
@joshuawilson do not merge this (marking WIP). Not until we have clarity on the correct approach for fixing the issue which I believe to be something else and auto save should be a separate discussion.

PR #3459, opened 2018-12-18
chore(lint): clean up list of disabled eslint rules

christianvogt:
Enabled a bunch of eslint rules in ngx-widgets there are now just a handful left disabled that should be reviewed further.

Ran eslint fix to auto fix as many issues as possible then proceeded to manually fix other issues.


christianvogt:
@joshuawilson I disabled the `typescript/no-inferrable-types` rule in the common eslint package. I will look into adding tslint support to add rules that eslint does not implement in another story.

PR #3460, opened 2018-12-20
fix(planner): Moving tooltip on + of "iterations" in planner #3068

Bumbilo:
- Fix position of tooltip

screen shot 2018-12-20 at 12 33 32
screen shot 2018-12-20 at 12 33 44


joshuawilson:
[test]

PR #3463, opened 2018-12-26
fix(style): width of the elements

vikram-raj:
Fix: https://github.com/openshiftio/openshift.io/issues/4645

**Before:**
Long title is broken in planner
screenshot from 2018-12-26 10-49-56

**After:**
Add truncate for long 
screenshot from 2018-12-26 10-50-58

**Before:**
Input field width is not 100%
screenshot from 2018-12-26 10-52-06

**After:**
Input field width is 100%
screenshot from 2018-12-26 10-52-57


joshuawilson:
@christianvogt this may have been broken by the linting or prettier, please take a look.

vikram-raj:
@joshuawilson No, it is not broken by lint. In planner it was written as ```width: ~'calc(e("100% - 90px"))';```. 
But it is not working after moving to mono repo. So, we need this changes.

joshuawilson:
ok

PR #3472, opened 2019-01-07
fix(login_process): remove the _gettingstarted page

karthikjeeyar:
* _gettingstarted page is removed
* account linking logic is moved to home component
* added error message in  home component, if account linking is unsuccessful

Depends on : https://github.com/openshiftio/openshift.io/pull/4684

Fixes: https://github.com/openshiftio/openshift.io/issues/2747


alien-ike:
### Ike Plugins (test-keeper)

Thank you @karthikjeeyar for this contribution!



joshuawilson:
Why was the package-lock file updated?

rohitkrai03:
I think the package-lock.json changes are from the initial `npm install`.

karthikjeeyar:
@joshuawilson package-lock.json was updated when i merged with master, i replaced the file with the contents in the master file

karthikjeeyar:
@joshuawilson removed package-lock.json.

christianvogt:
The home component is doing too much now. It would be better to have wrapped the home component in another component which would either load the home component or show the errors.

In general we want small concise components which perform only specific tasks to make them more portable.

ljelinkova:
Note that this might break e2e tests on prod-preview even if PR checks pass because PR check use 'local' mode which has different login page. 

@ppitonak FYI

karthikjeeyar:
@christianvogt i have refactored the code and added the wrapper component to hold old_home component and error component.

rohitkrai03:
@karthikjeeyar You should remove WIP label if you think the PR is ready for review.

rohitkrai03:
@karthikjeeyar It seems like tests are failing. The build failed because of that.

rohitkrai03:
```
 FAIL  src/app/shared/login.service.spec.ts
  ● LoginService › login › with "error" URL param › should handle new login

    expect(jest.fn()).toHaveBeenCalledWith(expected)

    Expected mock function to have been called with:
      "/"
    as argument 1, but it was called with
      "/_home".

      73 |         expect(authService.logIn).toHaveBeenCalledWith('some_token');
      74 |         expect(loginService.openShiftToken).toEqual('some_token');
    > 75 |         expect(router.navigateByUrl).toHaveBeenCalledWith(originalUrl);
         |                                      ^
      76 |       });
      77 |
      78 |       it('should handle existing login', () => {

      at src/app/shared/login.service.spec.ts:75:38
```

karthikjeeyar:
@rohitkrai03  fixed it.

rohitkrai03:
@karthikjeeyar Also, add a gif for the whole workflow including when the account is disconnected and linking goes wrong. I think @fabric8-ui/uxd should review this as well.

packages/fabric8-ui/src/app/home/home.component.ts
```diff
@@ -1,33 +1,66 @@
...
+        if (!isConnected && !wait) {
+          // first time through and user isn't connected - automatically connect accounts
+          this.providerService.linkOpenShift(
+            this.loggedInUser.attributes.cluster,
+            window.location.origin + '/_home?wait=true',
+          );
+        } else if (!isConnected && wait) {
+          // second time through - do not try again if wait is on URL and the user is still not connected
+          // something happened, show error
+          this.errorConnecting = true;
```
sbose78: @karthikjeeyar @rohitkrai03 
Are we redirecting to an error page? Or are showing a notification? 
What is the message we are showing in case of 'user still not connected'

rohitkrai03: @sbose78 If the user is not connected to openshift, home page will try to connect to openshift in the background. If still some error happens then this screen will be shown to the user - 

screenshot 2019-01-09 at 7 52 29 pm

sbose78: Perfect, thanks.

packages/fabric8-ui/src/app/home/link-error/link-error.component.html
```diff
@@ -0,0 +1,14 @@
+<div class="blank-slate-pf col-sm-8 col-sm-offset-2" id="getting-started-error">
```
joshuawilson: The ID `getting-started-error` is the same as on `getting-started.component.html`. We only use IDs as a reference from the tests. Having duplicate IDs could cause an issue for the tests. I don't see that you reference the ID. If you don't need it please remove it. If this page is meant to duplicate `getting-started.component.html` then why not point to it. On that note, why do you have a duplicate of `getting-started.component.html`?

karthikjeeyar: @joshuawilson `getting-started`  component will be removed completely.

PR #3476, opened 2019-01-11
fix(app-launcher): add service for build tool detection

vikram-raj:
Releted to - https://github.com/fabric8-launcher/ngx-launcher/pull/561
story:  https://openshift.io/openshiftio/Openshift_io/plan/detail/404

screenshot from 2019-01-12 00-07-31

screenshot from 2019-01-12 00-07-50


packages/fabric8-ui/src/app/space/app-launcher/services/app-launcher-gitprovider.service.ts
```diff
@@ -142,6 +144,21 @@ export class AppLauncherGitproviderService implements GitProviderService {
     );
   }
 
+  getDetectedBuildRuntime(repoUrl: string): Observable<BuildTool> {
+    const url = this.detectorApiUrl + 'detect/build/' + repoUrl;
+    if (this.headers.has('X-app')) {
+      this.headers = this.headers.delete('X-app');
```
christianvogt: Why are you deleting these headers?
Does this return a copy of the headers object or does it mutate the headers object? If it mutates, this will negatively affect the other service calls using these headers.

packages/fabric8-ui/src/app/space/app-launcher/services/app-launcher-gitprovider.service.ts
```diff
@@ -142,6 +144,21 @@ export class AppLauncherGitproviderService implements GitProviderService {
     );
   }
 
+  getDetectedBuildRuntime(repoUrl: string): Observable<BuildTool> {
+    const url = this.detectorApiUrl + 'detect/build/' + repoUrl;
```
christianvogt: Do you need to `encodeURIComponent(repoUrl)`?

vikram-raj: Yes, It is required. API is expecting only encoded URL.

sbose78:
@vikram-raj , How are you populating the drop-down ?

@tinakurian could we ensure that the selected value after the choice is made in the drop-down, is something that the backend API would recognize and return the right Jenkinsfiles for that?

packages/fabric8-ui/src/app/space/app-launcher/services/app-launcher-gitprovider.service.ts
christianvogt: @vikram-raj confirmed to me that the `repoUrl` is already encoded by the caller.

vikram-raj:
@sbose78 Currently It is hardcoded in the UI.

joshuawilson:
please rebase

vikram-raj:
@joshuawilson Done

vikram-raj:
[e2e-smoke-test]

PR #3477, opened 2019-01-13
preparation for new nav

christianvogt:
The following changes were done while adding support for the new nav. Created a separate PR as these are the only files affected by the larger change that adds a new package for the react app.

* Support dynamic imports by setting `module` to `esnext` in tsconfig
* Moved style definitions in `fabric8-ui` from loader to `app.modules`
  * needed because the loader is not used with the react app, but the module is and we need the styles to load lazily with the module
* Removed tooltips which opened up, covering the nav bar. These tooltips had redundant text which added zero value.
* Fixed a cut off tab due to a bad margin style

Removed margin fix:
image
image

Removed these useless tooltips:
image
image
image
image

### DIFF
--- a/packages/fabric8-ui/src/main.browser.ts
+++ b/packages/fabric8-ui/src/main.browser.ts
@@ -18,12 +18,6 @@ import { decorateModuleRef } from './app/environment';
 import { AppModule } from './app/index';
 
 /**
- * Import application wide styles
- */
-import './assets/stylesheets/fabric8-ui-global-overrides.less';
-import './assets/stylesheets/shared/osio.less';
-
-/**
  * Bootstrap our Angular app with a top level NgModule
  */
 export function main(): Promise<any> {
--- a/packages/scripts/config/tsconfig/tsconfig.json
+++ b/packages/scripts/config/tsconfig/tsconfig.json
@@ -11,7 +11,7 @@
     "isolatedModules": false,
     "jsx": "react",
     "lib": ["dom", "es2015", "es2016", "es2017"],
-    "module": "es2015",
+    "module": "esnext",
     "moduleResolution": "node",
     "noImplicitAny": true,
     "noImplicitReturns": true,
--- a/packages/scripts/lib/rewire-scripts.js
+++ b/packages/scripts/lib/rewire-scripts.js
@@ -8,6 +8,22 @@ const IgnoreNotFoundExportPlugin = require('./IgnoreNotFoundExportPlugin');
 
 process.env.SKIP_PREFLIGHT_CHECK = true;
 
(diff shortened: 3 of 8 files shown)


alien-ike:
### Ike Plugins (test-keeper)

packages/fabric8-ui/src/main.browser.ts
```diff
@@ -17,12 +17,6 @@ import 'reflect-metadata';
 import { decorateModuleRef } from './app/environment';
 import { AppModule } from './app/index';
 
-/**
- * Import application wide styles
- */
-import './assets/stylesheets/fabric8-ui-global-overrides.less';
```
joshuawilson: Why did you move this to app.module?

packages/scripts/config/tsconfig/tsconfig.json
```diff
@@ -11,7 +11,7 @@
     "isolatedModules": false,
     "jsx": "react",
     "lib": ["dom", "es2015", "es2016", "es2017"],
-    "module": "es2015",
+    "module": "esnext",
```
joshuawilson: what impact does this have?

packages/scripts/lib/rewire-scripts.js
```diff
@@ -138,7 +154,10 @@ function rewireWebpack(prod) {
         limit: 10000,
         name: `${staticDir}/media/[name].[hash:8].[ext]`,
         // TODO why is this path needed?
-        includePaths: [path.resolve(__dirname, '../../../node_modules/patternfly/dist/fonts/')],
+        includePaths: [
+          path.resolve(paths.appPath, 'node_modules/patternfly/dist/fonts/'),
+          path.resolve(__dirname, '../../../node_modules/patternfly/dist/fonts/'),
```
joshuawilson: Why have 2 here? Don't they hopefully resolve to the same thing?

packages/fabric8-ui/src/main.browser.ts
joshuawilson: I see why now in your notes above.

packages/scripts/lib/rewire-scripts.js
christianvogt: These are temp measures until we can remove the pattern fly dependency on CSS directly.
We have two levels of node_modules in our monorepo: those at the root and those at the app. The new addition looks in the app whereas before it didn't.

packages/scripts/config/tsconfig/tsconfig.json
christianvogt: Gives us new language features but code output is the same. From what I read it is only the dynamic import support.

PR #3482, opened 2019-01-15
update builds to point to packages/toolchain

christianvogt:
Update the build files to reference the `packages/toolchain` package instead of `packages/fabric8-ui`.

* Moved `semantic-release` to the `packages/toolchain` package.
* Renamed `package.json/name` of two packages to ensure semantic release continues to function like the new package is the same as the old and the release trail remains.
  * `fabric8-ui` -> `ngx-fabric8-ui`
  * `@osio/toolchain` -> `fabric8-ui`
* Also increased the specificity of override css rules to ensure the overrides take precedence.

~~Outstanding questions:~~
* Why do we need this for `ngx-launcher`: https://github.com/christianvogt/fabric8-ui/blob/d4e337147cefd0989781293831ebbe1cc59cf2e6/packages/fabric8-ui/src/app/environment.ts#L26
  * Can launcher function without this static injector code being executed?
  * What is the StaticInjector used for in ngx-launcher? It says to get the correct broadcaster instance but why is this is a "special" case? Why doesn't normal dependency injection work when asking for the broadcaster?
  * Is it possible that this is no longer needed and was a work around because multiple instances of the ngx-base library was previously installed in the application?

@jyasveer @edewit


alien-ike:
### Ike Plugins (test-keeper)

edewit:
@christianvogt I was a work around to get the broadcaster injected, I
remember that I tried to run it without that as I felt that it wasn't
necessary, but I couldn't get it to work without it. So it's possible that
now with fixes you mention it could be removed again and normal injection
will work.

christianvogt:
Looked at the code closer. Unless we update `ngx-launcher` we need to still use this `StaticInjector` otherwise `ngx-launcher` won't be sending any analytics events.

christianvogt:
I have moved the initialization of `ngx-launcher`'s `StaticInjector` to the `app.module` so this code is still executed with the new navigation entry point.

christianvogt:
[e2e-test-released]

PR #3488, opened 2019-01-16
fix(deployment): Use Build Tool Detector URL from config map

jarifibrahim:
Use `BUILD_TOOL_DETECTOR` env value from config map instead of template params.


joshuawilson:
How does this change read the value?

PR #3493, opened 2019-01-17
fix(loading_state): Add test cases for home component

karthikjeeyar:
add test case for home component

Fixes: https://github.com/openshiftio/openshift.io/issues/4709


karthikjeeyar:
@christianvogt  fixed the test cases

PR #3494, opened 2019-01-17
fix(navigation): fix space navigation to return to root for 'All Spaces' in overview

christianvogt:
fixes: https://github.com/openshiftio/openshift.io/issues/4715
There's a bug in the new nav where visiting the overview while in the context of a space does not return you to the root when switching back to `All Spaces`. Instead it brings you to the user profile page.

This changes fixes the routing context to ensure that `All Spaces` for the overview page is the root (or /_home for that matter).

### DIFF
--- a/packages/planner/src/app/components_ngrx/iterations-modal/iterations-modal.component.ts
+++ b/packages/planner/src/app/components_ngrx/iterations-modal/iterations-modal.component.ts
@@ -277,7 +277,7 @@ export class FabPlannerIterationModalComponent implements OnInit, OnDestroy, OnC
     // Format 2016-11-29T23:18:14Z
     if (event.jsdate !== null) {
       this.startDate = { date: event.date };
-      this.iteration.startAt = moment(event.jsdate).format('YYYY-MM-DD') + 'T12:00:00Z';
+      this.iteration.startAt = moment(event.jsdate).format('YYYY-MM-DD') + 'T00:00:00Z';
     } else {
       this.startDate = '';
       this.iteration.startAt = '';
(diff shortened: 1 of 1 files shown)


packages/toolchain/src/app/components/SpaceDropdownPageSection.tsx
```diff
@@ -44,7 +44,8 @@ export class SpaceDropdownPageSection extends React.Component<Props> {
             {username != null && (
               <DropdownItemRoute
                 key="all"
-                href={`/${username}/${NO_SPACE_PATH}${subPath ? `/${subPath}` : ''}`}
+                // if there is no subPath we must go to the root
+                href={subPath ? `/${username}/${NO_SPACE_PATH}/${subPath}` : ''}
```
vikram-raj: Instead of ```username```, it should be space owner username. Otherwise, it will give 404 error If the user visits the other space. where the user is the collaborator of that space.

christianvogt: This dropdown only deals with the current user's spaces and not spaces from other users.

vikram-raj: Isn't this is the same dropdrown that is visible on secondary header of the dropdown.

christianvogt: It is. But that header is only showing the logged in users's spaces. Perhaps it should show all spaces the user collaborates with. But that is a feature change. I'll talk to UX as the intent from the invision wasn't clear.

christianvogt: My mistake. You're right, the api does return all spaces the user collaborates with. Will address.

vikram-raj: Yes, it shows all the space the users collaborates with.

christianvogt:
@vikram-raj improved the nav experience
I don't think UX really thought it out too much about what it means to navigate between users and spaces and what the links should be in those cases. Either way this change is now an improvement and can be iterated on.

christianvogt:
@vikram-raj @joshuawilson Since this feature is still under an internal flag I'd like to get these changes in so initial testing can work with it. I have another PR that'll come with many unit tests for the react app.

PR #3502, opened 2019-01-24
Remove.wit.deployments.api.calls

corinnekrych:
Related to fixes https://openshift.io/openshiftio/Openshift_io/plan/detail/1315
- This PR makes sure all call to WIT deployments API are removed in the UI (before we could merge PR: https://github.com/fabric8-services/fabric8-wit/pull/2395).
- It does not change any call to a-runtime-console


corinnekrych:
@christianvogt done.
If we don't want to ckeck-in packages/<project>/package-lock.json we should add them to .gitignore, wdyt?

christianvogt:
Ya we should add exclusion to the packages for now. When I install & bootstrap I do not get the package-lock generated in the sub dir.

packages/fabric8-ui/src/app/space/app-launcher/services/app-launcher-dependency-check.service.ts
```diff
@@ -39,9 +35,7 @@ export class AppLauncherDependencyCheckService implements DependencyCheckService
    *
    * @returns Observable
    */
-  getApplicationsInASpace(): Observable<Application[]> {
-    return this.deploymentApiService.getApplications(
-      this.context.space ? this.context.space.id : '',
-    );
+  getApplicationsInASpace(): Observable<any[]> {
+    return of([])
```
christianvogt: @vikram-raj how is this used in launcher?

vikram-raj: @christianvogt It is use to validate the application name while creating the application. So, that user can not create multiple application with the same name.

corinnekrych:
@christianvogt @vikram-raj using `PipelinesService` with BuildConfig information gives us the name of the application available on OS, so the validation is added back on app creation flow:
unspecified

packages/fabric8-ui/src/app/space/add-app-overlay/add-app-overlay.component.spec.ts
```diff
@@ -216,15 +212,6 @@ describe('AddAppOverlayComponent', () => {
       expect(component.applications).toEqual([]);
     });
 
-    it('should retieve applications if the current space is defined', () => {
```
christianvogt: Can we keep these test cases now that you've found a work around to the list of applications.

packages/fabric8-ui/src/app/space/app-launcher/services/app-launcher-dependency-check.service.ts
christianvogt: @corinnekrych can we use the same workaround here to get the application names?

vikram-raj: @corinnekrych ```getApplicationsInSpace``` is getting called in ```project-name.validator``` directive in ngx-launcher. https://github.com/fabric8-launcher/ngx-launcher/blob/master/projects/ngx-launcher/src/lib/shared/project-name.validator.ts#L34

corinnekrych:
@christianvogt I put back the tests, unit tests are ok now but CI is failing with this:
```
FATAL: could not log in into vault
com.bettercloud.vault.VaultException: Vault responded with HTTP status code: 502
	at com.bettercloud.vault.api.Auth.loginByAppRole(Auth.java:386)
	at com.datapipe.jenkins.vault.credentials.VaultAppRoleCredential.authorizeWithVault(VaultAppRoleCredential.java:42)
Caused: com.datapipe.jenkins.vault.exception.VaultPluginException: could not log in into vault
	at com.datapipe.jenkins.vault.credentials.VaultAppRoleCredential.authorizeWithVault(VaultAppRoleCredential.java:44)
	at com.datapipe.jenkins.vault.VaultAccessor.auth(VaultAccessor.java:29)
```

🤔 #puzzled

corinnekrych:
[test]

packages/fabric8-ui/src/app/space/app-launcher/services/app-launcher-dependency-check.service.ts
corinnekrych: @edewit do you know why we used project-name.validator given that we've already checked the name is not already in used in https://github.com/fabric8-ui/fabric8-ui/blob/8ebda0a75136aaaa525d54da5c018e026ce01409/packages/fabric8-ui/src/app/space/add-app-overlay/add-app-overlay.component.ts#L165 ?

christianvogt: @corinnekrych Looks like the app launcher has two more areas after the initial screen where the user can change the application name:

image
image

christianvogt:
Something is wrong now with local dev and login. I'm getting a `401` error on load calling endpoint `/api/user/services` which is then deleting the `auth_token` from localstorage before the app loads which then puts me back to the login button screen.

I cannot login to the dev app.

christianvogt:
Looked into the login issue a bit further. This is caused by the `add-app-overlay.component` now requiring the pipeline service which is run immediately even though the `add-app-overlay.component` isn't even shown but it is initialized.

@vikram-raj @rohitkrai03 could one of you help @corinnekrych in addressing this issue. Maybe somehow move the init code to only when the modal is actually needed rather than on app load.

corinnekrych:
@christianvogt @rohitkrai03 indeed testing with a new incognito browser I can see the issue :( 
the injection of PipelinesService seems to be the root cause. 
I'm 👀  too.

corinnekrych:
Interesting i've just discovered https://github.com/fabric8-ui/fabric8-ui/blob/master/packages/fabric8-ui/src/app/space/create/pipelines/services/pipelines.service.ts#L131

packages/fabric8-ui/src/app/space/app-launcher/services/app-launcher-dependency-check.service.ts
corinnekrych: gotcha indeed!

corinnekrych:
not to worry @edewit enjoy your ⛷ holidays, we got it sorted with, as Christian identified where it was used:
unspecified-3

@christianvogt ready for another review

edewit:
Sorry but I have no idea

PR #3505, opened 2019-01-28
fix(env): assign wit url to feature toggles and add fallback to /api/

christianvogt:
The `api-locator.service.ts` in the angular app does not expose any URL which uses the `FABRIC8_FEATURE_TOGGLES_API_URL` env variable. The `toggles.api.provider.ts` provides the wit api url for feature toggles.

Using `FABRIC8_FEATURE_TOGGLES_API_URL` breaks the react app in prod preview today because the URL is incorrect. Furthermore, in production, the `fabric8-ui.env.js` file doesn't even include a value for the feature toggles. Most URLs are empty. Which means we need to add a default fallback to `/api/` just like the `api-locator.service.ts` does.

The result of not having this fix is that on prod-preview, the react app queries for features at `https://api.prod-preview.openshift.io/api/features?strategy=enableByLevel` while the angular app queries `https://prod-preview.openshift.io/api/features?strategy=enableByLevel`. The value the react app gets is invalid and does not include the correct user enablement settings for whatever reason.

We should fix the URL in prod-proview env, but for now this change aligns the react and angular apps in how they provide api URLs.

### DIFF
--- a/packages/toolchain/src/app/api/internal/__tests__/api.config.spec.ts
+++ b/packages/toolchain/src/app/api/internal/__tests__/api.config.spec.ts
@@ -3,16 +3,28 @@ describe('api.config', () => {
     jest.resetModules();
   });
 
+  it('should fallback to api URLs default', () => {
+    const bak = process.env;
+    process.env = {};
+
+    // eslint-disable-next-line
+    const config = require('../api.config');
+    expect(config.AUTH_API_URL).toBe('/api/');
+    expect(config.FEATURE_TOGGLES_API_URL).toBe('/api/');
+    expect(config.WIT_API_URL).toBe('/api/');
+
+    process.env = bak;
+  });
+
   it('should fallback to api URLs from process.env', () => {
     const bak = process.env;
     process.env.FABRIC8_AUTH_API_URL = 'test-auth-url';
-    process.env.FABRIC8_FEATURE_TOGGLES_API_URL = 'test-ft-url';
     process.env.FABRIC8_WIT_API_URL = 'test-wit-url';
 
     // eslint-disable-next-line
     const config = require('../api.config');
     expect(config.AUTH_API_URL).toBe('test-auth-url');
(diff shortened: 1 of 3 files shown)


rohitkrai03:
The code looks good to me. But i wonder why `api.prod-preview.openshift.io/api/` is returning wrong value and `prod-preview.openshift.io` correct because `prod-preview.openshift.io` is actually being resolved to `api.prod-preview.openshift.io/api/` in the backend. So, the actual value of `WIT_API_URL` should be `api.prod-preview.openshift.io/api/`.

rohitkrai03:
Even this approach doesn't work in local dev. We need to prune the config object to remove properties with `"undefined"` values.

christianvogt:
A default fallback will not work in local dev. Which is why we need to source an environment locally before starting our dev server.
Is this what you're referring to?

rohitkrai03:
So, this line `export const WIT_API_URL = config.witApiUrl || FABRIC8_WIT_API_URL || DEFAULT_URL;` basically says that if `config.witApiUrl` is falsy then use `FABRIC8_WIT_API_URL` which is coming from `process.env` and will always be there in dev environment because we export it in `env.js`. But `config.witApiUrl` is always true because its value is `"undefined"`.

rohitkrai03:
So, if we add something like 
```
// Cleanup config properties with falsy or 'undefined' values
Object.keys(config).forEach(
  (key) => !config[key] || (config[key] === 'undefined' && delete config[key]),
);
```

It will work in local environment even without exporting anything explicitly.

christianvogt:
Oh the value was a string. I see now.

I'll update the pr with your suggestion.

christianvogt:
Fixed the local env webpack config which was causing strings in the api urls to have a value of `'undefined'`.

rohitkrai03:
Great. Code looks good. Already approved. ;)

PR #3510, opened 2019-01-30
fix(preview): Markdown component layout breaks on copy-paste

Bumbilo:
fixes: https://github.com/openshiftio/openshift.io/issues/3972
- Made the textarea instead of a contenteditable element
screen shot 2019-01-30 at 11 30 19

### DIFF
--- a/packages/ngx-widgets/src/app/area-size/area-size.directive.spec.ts
+++ b/packages/ngx-widgets/src/app/area-size/area-size.directive.spec.ts
@@ -1,0 +1,47 @@
+import { Component } from '@angular/core';
+import { TestBed, ComponentFixture } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { AreaSizeDirective } from './area-size.directive';
+
+@Component({
+  selector: 'my-test-component',
+  template: '<textarea textAreaResize class="editor-box editor-markdown">{{content}}</textarea>',
+})
+class TestComponent {
+  content: string = '';
+}
+
+describe('area-size directive: ', () => {
+  let component: TestComponent;
+  let fixture: ComponentFixture<TestComponent>;
+  let textarea: HTMLTextAreaElement;
+  let autosize: any;
+
+  beforeEach(async () => {
+    TestBed.configureTestingModule({
+      declarations: [TestComponent, AreaSizeDirective],
+    });
+  });
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(TestComponent);
+    fixture.detectChanges();
+    component = fixture.componentInstance;
+    textarea = fixture.nativeElement.querySelector('textarea');
+    autosize = fixture.debugElement.query(By.directive(AreaSizeDirective));
+  });
+
+  it('textarea should have equal content height', () => {
+    expect(textarea.clientHeight).toEqual(textarea.scrollHeight);
+
+    fixture.componentInstance.content = `
+     Et harum quidem rerum facilis est et expedita distinctio. Nam libero tempore, cum soluta nobis est eligendi optio cumque nihil impedit quo minus id quod maxime placeat facere possimus, omnis voluptas assumenda est, omnis dolor repellendus. Temporibus autem quibusdam et aut officiis debitis aut rerum necessitatibus saepe eveniet ut et voluptates repudiandae sint et molestiae non recusandae. Itaque earum rerum hic tenetur a sapiente delectus, ut aut reiciendis voluptatibus maiores alias consequatur aut perferendis doloribus asperiores repellat`;
+    autosize.triggerEventHandler('input', null);
+    fixture.detectChanges();
+    expect(textarea.clientHeight).toEqual(textarea.scrollHeight);
+  });
+
+  it('the component should be identified by the directive', () => {
+    expect(autosize).toBeDefined();
+  });
+});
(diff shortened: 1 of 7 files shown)


packages/ngx-widgets/src/app/area-size/area-size.directive.ts
```diff
@@ -0,0 +1,17 @@
+import { Directive, ElementRef, HostListener, Input, Renderer2 } from '@angular/core';
+
+@Directive({
+  selector: '[textAriaResize]',
```
christianvogt: Rename to `textAreaResize`

packages/ngx-widgets/src/app/area-size/area-size.directive.ts
```diff
@@ -0,0 +1,17 @@
...
+})
+export class AreaSizeDirective {
+  private element: HTMLElement = this.elementRef.nativeElement;
+
+  constructor(private elementRef: ElementRef, private renderer: Renderer2) {}
+
+  @HostListener('input', ['$event'])
+  resizeOnInput() {
+    const input = event.target as HTMLInputElement;
+    this.renderer.setStyle(this.element, 'height', '1px');
```
christianvogt: Why do you set to `1px` here only to set it to the `scrollHeight` immediately afterwards?

packages/ngx-widgets/src/app/markdown/markdown.component.html
```diff
@@ -17,15 +17,17 @@
     <div class="editor-icon-container">
       <i *ngIf="editAllow" (click)="activeEditor()" class="pficon-edit edit-icon"></i>
     </div>
-    <p
+    <textarea
       almEditable
+      textAriaResize
       #editorInput
+      [style.height.px]="editorInput.scrollHeight"
```
christianvogt: What is the purpose of this if you already have the text area resize directive?

packages/ngx-widgets/src/app/area-size/area-size.directive.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      declarations: [TestComponent, AreaSizeDirective],
+    });
+    fixture = TestBed.createComponent(TestComponent);
+    component = fixture.componentInstance;
+    textAreaElement = fixture.debugElement.query(By.directive(AreaSizeDirective));
+  });
+
+  it('textarea should have equel content height', () => {
```
christianvogt: `equel` => `equal`

packages/ngx-widgets/src/app/area-size/area-size.directive.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    expect(textAreaElement.nativeElement.offsetHeight).toEqual(
+      textAreaElement.nativeElement.scrollHeight,
+    );
+  });
+
+  it('the component should be identified by the directived', () => {
```
christianvogt: `directived` => `directive`

packages/ngx-widgets/src/app/area-size/area-size.directive.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    textAreaElement.triggerEventHandler('input', null);
+    spyOn(textAreaElement.nativeElement, 'focus');
+    expect(textAreaElement).not.toBeNull();
+    fixture.detectChanges();
+    setTimeout(() => {
```
christianvogt: Is this timeout necessary?
Async tests need to use the `done()` protocol.

packages/ngx-widgets/src/app/area-size/area-size.directive.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  it('textarea should have equel content height', () => {
+    expect(textAreaElement.nativeElement.offsetHeight).toEqual(
+      textAreaElement.nativeElement.scrollHeight,
+    );
+  });
+
+  it('the component should be identified by the directived', () => {
+    textAreaElement.triggerEventHandler('input', null);
+    spyOn(textAreaElement.nativeElement, 'focus');
```
christianvogt: Why add a focus spy if there's no assertions being made?

packages/ngx-widgets/src/app/area-size/area-size.directive.ts
Bumbilo: When I remove a line in the text area. scrollHeight has the same height and necessary to redefine it.

christianvogt: Ok I understand why you're doing it now.
I think you need to ensure as well that `overflow-y` is set to `hidden` to account for the scrollbar appearing for OS / browser settings where a scrollbar consumes space. The scrollbar will flash as well as push the text resulting possibly in an incorrect calculation. Test this on several browsers.

I don't know if this is the best solution as it's doing a re-layout of the component on each input change. I can think of other ways to do this but they are certainly more involved with separate height calculations in another element. So if this works and doesn't give any negative experiences to the user (flashing, etc...), i'm ok with it.

christianvogt:
@Bumbilo please test in firefox as I observed editing an existing description loses all new line characters in the text.

christianvogt:
@Bumbilo
Tested in FireFox version:
image

Here's what happens:
markdown_ff

Bumbilo:
@christianvogt  
Fixed
firefox

christianvogt:
Your screen cap shows issues with lists showing up. I tested locally in chrome and FF and experienced the same issue. I cannot create markdown lists anymore.

PR #3511, opened 2019-01-30
[WIP] fix(func-test): fix functional tests

sahil143:


rgarg1:
@sanketpathak Please help/review @sahil143 to get this PR going.

packages/planner/tests/src/specs/collaboratorTest.spec.ts
```diff
@@ -2,7 +2,7 @@ import { browser } from 'protractor';
 import { PlannerPage } from '../page_objects/planner';
 import * as support from '../support';
 
-describe('Planner Collaborator Tests:', () => {
+xdescribe('Planner Collaborator Tests:', () => {
```
sanketpathak: Are we  will excluding these specs from execution

sahil143: I didn't notice that. I will enable these tests.

vikram-raj (commented):
Two tests are failing for me 
```1) Edit Comment and Save```
```2) Iteration test updating iteration should update workitem associated to iteration```

File: packages/planner/tests/package.json [hunk 2]
```diff
@@ -18,12 +18,8 @@
   },
   "dependencies": {
     "jasmine-spec-reporter": "4.2.1",
-    "protractor": "5.2.2",
+    "protractor": "5.2.0",
     "typescript": "2.9.2"
   },
-  "peerDependencies": {},
-  "devDependencies": {
```
sanketpathak: Why have you removed the devDependencies it is giving error in protractor.conf.ts that `protractor.conf.ts can not read jasmine`

sanketpathak:
Fifteen tests are failing when locally tested.

packages/planner/tests/package.json
sahil143: I didn't get this error after removing @types/jasmine. @vikram-raj can you confirm this?

I removed these because even if these dependencies are installed, compiler is always looking for these packages in the parent node_modules and these should be in parent node_modules because we also use jasmine for our unit tests.

sahil143:
> Fifteen tests are failing in total when locally tested.

@sanketpathak can you try in headless mode also.

packages/planner/tests/package.json
vikram-raj: @sahil143  @sanketpathak Tests are running without any error for me.

sanketpathak: It's failing not just for me but I've tested with other machines and its failing on those as well.

